PR #56, opened 2026-05-02
refactor(storage): retire BYTEA, S3-compatible only + flip CI to MinIO

Rasbandit:
## Summary

Plaintext attachment bytes no longer touch Postgres. The `Engram.Storage` behaviour now requires an external S3-compatible backend (Tigris in prod, MinIO in dev/CI, ETS-backed `InMemory` adapter in unit tests). `Engram.Storage.Database` is deleted along with the `attachments.content` BYTEA column.

**Stacked on top of #55 (Phase A encrypt-at-rest).** GitHub diff shows both until #55 merges; this PR's own additions are the last 3 commits (`ec462c5`, `cbc3977`, `3f94106`).

### Why

- Postgres BYTEA caps practical attachment size at ~5MB (TOAST overhead, replication bloat, buffer-cache pollution). Engram attachments include images / PDFs / audio — the cap was artificially imposed by the storage choice, not the product.
- Production already runs Tigris. The BYTEA path was the *least* tested under prod-like conditions.
- Future S3-native features (presigned URLs, multipart, lifecycle policies, server-side preview gen) can ship without per-adapter parity work.
- Self-hosted users get a MinIO sidecar — `docker-compose.elixir.yml` already bundles it, so the marginal deployment cost is one extra container.

### Changes

**Schema + code retirement (commit `ec462c5`)**

- Migration: `attachments.content` column dropped.
- `lib/engram/storage/database.ex` deleted.
- `lib/engram/attachments.ex`: dual-path `get_attachment` collapsed; `maybe_include_content` + `Storage.Database` delete clause removed; `prepare_upload` no longer takes ciphertext.
- `lib/engram/workers/encrypt_attachments.ex`: BYTEA branch in `encrypt_one` removed; only S3 round-trip remains.
- `Attachment` schema: `:content` becomes a virtual field (transient, populated by `get_attachment` after fetch + decrypt).
- `config/config.exs` default → `Engram.Storage.S3`. `config/runtime.exs`: `STORAGE_BACKEND` env removed (unconditional S3). `config/test.exs` → new `Engram.Storage.InMemory` ETS-backed adapter under `test/support/`.
- Tests updated: BYTEA-specific scenarios removed; cleanup-vault test switches to MockStorage-raise to exercise the failure-tolerant cleanup branch.

**CI flip + cleanup hygiene (commit `cbc3977`)**

- `docker-compose.ci.yml`: MinIO + minio-init services added; engram service depends on `minio-init` completion + ships with `STORAGE_HOST=minio` etc. compose `--wait` now blocks on MinIO healthcheck.
- `docker-compose.parity.yml`: collapsed back to embed-only overrides; storage env removed (inherited from base).
- `e2e/helpers/cleanup.py`: new `cleanup_minio_bucket()` — `mc rm --recursive --force` against `engram-attachments`. Wired into `full_cleanup`. No-op when MinIO container is absent.
- `e2e/helpers/crypto_probe.py`: `assert_attachment_ciphertext_at_rest` now `mc stat`s the MinIO object and asserts byte size == `size_bytes + 16` (AES-GCM tag suffix). Falls back to row-only assertions if MinIO is unreachable.

### Test plan

- [x] `mix test` — 842/842 (was 860 on PR #55; -1 BYTEA-specific worker test, -database adapter tests).
- [ ] PR CI E2E: confirm MinIO healthcheck passes during compose `--wait`, `test_attachment_ciphertext_at_rest` exercises real S3 path, `cleanup_minio_bucket` runs in teardown.
- [ ] Manual: dev stack via `docker-compose.elixir.yml`, upload via plugin, verify attachment lands in MinIO bucket and round-trips through GET.
- [ ] Self-hosted smoke: bring up `docker-compose.elixir.yml`, confirm MinIO sidecar boots cleanly with default `minioadmin` creds.

### Out of scope

- Production data migration: not needed — early dev, no BYTEA rows in prod.
- Streaming + S3 multipart for >5MB attachments: separate PR (raise the cap, then ship).
- BYOK-bucket support for self-hosted: future Phase F-adjacent work.


Rasbandit:
Folded into #55 — same branch contains all commits now.

PR #123, opened 2026-05-13
perf(ci): tighten compose healthchecks for faster stack-up

Rasbandit:
## Summary
Engram CI stack \`Start CI stack\` step burned ~10s on coarse healthcheck intervals while BEAM/postgres/minio were already ready mid-interval. Drop interval to 1s on all three services.

| service | before | after |
|---------|--------|-------|
| engram  | start_period 3s, interval 3s, retries 15 (48s budget) | start_period 10s, interval 1s, retries 60 (70s budget) |
| postgres | interval 3s, retries 10 (30s) | interval 1s, retries 30 (30s) |
| minio | interval 5s, retries 5 (25s) | interval 1s, retries 30 (30s) |

Same or slightly larger total retry budget, but \`docker compose --wait\` detects "healthy" within 1s of the service actually being ready (instead of up to 3-5s late).

## Test plan
- [x] mix.exs bumped to 0.5.84
- [ ] CI green
- [ ] Compare \`Start CI stack\` step duration vs prior runs

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #124, opened 2026-05-13
perf(ci): skip test jobs on main when tree fingerprint already proved green

Rasbandit:
## Summary
Squash merges produce a commit whose **tree** is identical to the PR's head. The PR's CI already validated that tree — re-running 5+ min of tests on main is wasted runtime + extra deploy latency. This adds a content-fingerprint short-circuit.

### Mechanism
- **\`fingerprint\` job** (new): hashes every source path that affects test outcomes via \`git ls-tree -r HEAD -- <paths>\`. Paths covered: \`lib priv config test frontend Dockerfile entrypoint.sh mix.exs mix.lock docker-compose.ci*.yml .github/workflows/ci.yml\`.
- **Cache lookup**: \`actions/cache/restore@v5\` with \`lookup-only: true\` and key \`ci-pass-<hash>\`. Sets \`cache-hit\` output.
- **Test jobs** (unit-tests, lint, e2e-clerk, e2e-local, e2e-browser): gated on \`needs.fingerprint.outputs.cache-hit != 'true'\`. Skipped entirely on cache hit.
- **\`ci\` aggregator**: also gated on cache-hit, so it doesn't fail when all upstream jobs are skipped.
- **\`record-pass\` job** (new): on green CI, writes a marker and saves to \`ci-pass-<hash>\`. Next run with same tree → cache hit.
- **\`deploy-fastraid\`**: now accepts EITHER \`needs.ci.result == 'success'\` OR \`needs.fingerprint.outputs.cache-hit == 'true'\`.

### Why this is safe
- Branch protection enforces required checks on the **PR head commit**, not the merge commit. PR's CI runs in full. The merge to main is the only run that skips.
- The fingerprint includes \`ci.yml\` itself, so any workflow change forces a full re-run.
- actions/cache TTL is 7 days untouched — fingerprints older than that fall out and pay one full run.

### Expected impact
- PR push: full CI as today (~3-4 min). Now also saves cache.
- Merge to main: **fingerprint job (~3s) + deploy-fastraid (~3 min)**. Saves 5+ min off every merge.

## Test plan
- [ ] First push to this branch: fingerprint should cache-miss (workflow itself changed), full CI runs
- [ ] Verify \`record-pass\` job completes
- [ ] After merge to main: fingerprint should cache-hit, all test jobs skip, deploy runs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,60 @@ env:
   ENCRYPTION_MASTER_KEY: ${{ secrets.CI_ENCRYPTION_MASTER_KEY }}
 
 jobs:
+  # Fingerprint: content-hash of every source path that affects test
+  # results. If the same hash already produced a green CI run, downstream
+  # test jobs skip entirely and deploy-fastraid (on main) runs against
+  # the already-validated tree.
+  #
+  # The merge-to-main commit produced by a squash merge has the same
+  # `git ls-tree` output as the PR's head commit — same content, same
+  # fingerprint, same cache key. PR CI proves the tree; main CI doesn't
+  # need to re-run 5+ minutes of e2e for the same bits.
+  fingerprint:
+    # Runs on every trigger (push, repository_dispatch, workflow_dispatch).
+    # Cross-repo plugin dispatch has its own checkout shape but `git ls-tree`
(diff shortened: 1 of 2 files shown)


Rasbandit:
Collapsed into #126 alongside #123. See #126 for the unified pipeline tuning.

PR #134, opened 2026-05-14
ci: split fingerprint into backend + e2e (plugin-aware)

Rasbandit:
## Summary
- Fingerprint job now resolves the plugin SHA and emits two hashes: `backend-hash` (gates pure-backend jobs) and `e2e-hash` = sha256(backend-hash + plugin-sha) (gates `e2e-clerk`).
- `e2e-clerk` checks out plugin at the exact `PLUGIN_SHA` from the fingerprint job, eliminating drift between cache key and tested code.
- `record-pass` saves both markers; `deploy-fastraid` requires either `ci` success or BOTH cache hits.
- New `report-cached-pass-to-plugin` job posts cached green to plugin commits when `e2e-clerk` is skipped on dispatch.

## Why
Plugin PR re-dispatches against an unchanged backend tree previously hit the backend-only cache and skipped `e2e-clerk` → plugin PR got a false green. Symmetrically, backend pushes with the plugin main branch moved would also skip e2e against the new plugin code. The split-hash fingerprint makes the cache key mean what it should: "this exact (backend, plugin) pair has been validated."

## Test plan
- [ ] First push: backend + e2e caches both miss → full suite runs and seeds both markers
- [ ] Empty-content re-push on same branch: both cache hits → `ci` skips, `deploy-fastraid` gates correctly on main
- [ ] Dispatch from a plugin PR with new plugin SHA: e2e cache miss → `e2e-clerk` runs, posts status to plugin
- [ ] Dispatch from a plugin PR with unchanged plugin SHA after a green pass: e2e cache hit → `e2e-clerk` skips, `report-cached-pass-to-plugin` posts success to plugin
- [ ] Backend push with plugin main moved since last pass: e2e miss but backend hit → only `e2e-clerk` re-runs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,42 +65,106 @@ jobs:
   # fingerprint, same cache key.
   fingerprint:
     # Runs on every trigger (push, repository_dispatch, workflow_dispatch).
+    # Emits TWO hashes:
+    #   - backend-hash: pure backend tree. Gates unit-tests/lint/e2e-local/e2e-browser/deploy.
+    #   - e2e-hash:     backend-hash + plugin SHA. Gates e2e-clerk only (the plugin↔backend contract).
+    # This lets a plugin-only change (same backend, different plugin SHA) miss the
+    # e2e cache and re-run integration, while still hitting the backend cache so
+    # pure backend jobs skip.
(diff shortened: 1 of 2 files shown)


Rasbandit:
Reopening from chore/ prefix — ci/ prefix not in push-trigger list.

PR #154, opened 2026-05-18
test(e2e): SyncCenter activity log + Ignored restore (#57)

Rasbandit:
## Summary

- Adds `e2e/tests/test_57_sync_center_log.py` with two tests:
  1. **Activity log**: push a note → assert `action == "push"` entry appears → Clear → assert empty
  2. **Ignored restore**: CDP-pivot ignore a clean file → delete from server → Restore via Sync Center → assert re-pushed via `trigger_full_sync` + `api_sync.get_note`
- Bumps `mix.exs` to `0.5.117`
- Stacks on #148–#153 (feat/e2e-06-sync-center-issues)

## Key implementation decisions

**Activity action value:** `"push"` (lowercase) — `renderActivityRow` in `sync-center-render.ts` writes `entry.action` directly as text content of `.engram-sync-center-activity-action`.

**Ignore-clean-file resolution — CDP manual-mutation pivot:**
The `Ignore` button only exists on Issues panel rows (`renderIssueRow`). Clean files have no Ignore button in the UI. The test uses `_cdp_ignore_file()`, which calls `plugin.syncEngine.ignoredFiles.add(path)` + `persistEngineState()` + `syncCenter?.refresh?.()` via CDP evaluate.

**ApiClient methods added:** None — `get_note` and `delete_note` already existed in `e2e/helpers/api.py`.

## Test plan

- [ ] `has_sync_center()` guard skips cleanly on old plugin builds
- [ ] Activity log test verifies `action == "push"` and `status == "ok"`
- [ ] Clear wipes the actual `syncLog` ring buffer (not just DOM)
- [ ] Restore test confirms note absent from server after delete, present after Restore + fullSync
- [ ] Cleanup in `finally` blocks removes ignored entries, server notes, and local files

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #155, opened 2026-05-18
test(e2e): command palette entries fire handlers (#58)

Rasbandit:
## Summary

- Adds `e2e/tests/test_58_commands_palette.py` with five tests covering `sync-now`, `push-all`, `pull-all`, `check-sync`, and `show-sync-log` command palette entries
- Each test uses `has_command()` skip-gate, installs a spy/wrapper in `try/finally`, and restores state unconditionally
- `Notice` interception patches `require('obsidian').Notice` (not `window.Notice`) so compiled plugin callsites route through the wrapper; original is captured before patching and restored in `finally`
- `pushAll`/`pullAll` spies capture the original via `.bind()` and restore the exact reference on cleanup
- Bumps `mix.exs` to `0.5.118`

## Command id verification (source: `src/main.ts`)

| Command id | Status |
|---|---|
| `sync-now` | confirmed |
| `push-all` | confirmed |
| `pull-all` | confirmed |
| `check-sync` | confirmed |
| `show-sync-log` | confirmed |

## SyncEngine method names (source: `src/sync.ts`)

- `pushAll` — line 1999: `async pushAll(opts: { deleteRemoteExtras?: boolean } = {})`
- `pullAll` — line 942: `async pullAll(opts: { deleteLocalExtras?: boolean } = {})`

## Fixes vs plan

- Plan cleanup for Notice used `window.Notice.__wrapped__` (only works if that attribute was explicitly set); replaced with `window.__e2e_notice_orig` captured before wrapping
- Plan cleanup for push-all spy deleted `p.syncEngine._origPushAll` (never set); replaced with `window.__e2e_pushAll_orig` pattern matching pull-all
- Added polling loops for async commands instead of fixed `asyncio.sleep`

Stacks on #148–#154.

## Test plan

- [ ] Static syntax check passes (`python3 -m py_compile`)
- [ ] All 5 command ids confirmed in plugin source
- [ ] `pushAll`/`pullAll` method names confirmed in `src/sync.ts`
- [ ] Notice cleanup restores `require('obsidian').Notice` to original
- [ ] Each test cleans up spy/wrapper in `finally` block

### DIFF
--- a/docs/superpowers/plans/2026-05-17-obsidian-plugin-e2e-coverage-gaps.md
+++ b/docs/superpowers/plans/2026-05-17-obsidian-plugin-e2e-coverage-gaps.md
@@ -8,6 +8,33 @@
 
 **Tech Stack:** Python 3.11 + pytest-asyncio, Chrome DevTools Protocol via `helpers/cdp.py` (`CdpClient`), Obsidian AppImage + Xvfb (`helpers/obsidian.py`), `helpers/api.py` `ApiClient` for server-side assertions, `helpers/vault.py` for filesystem reads/writes.
 
+## Selector corrections (applied during Task 1 — apply everywhere)
+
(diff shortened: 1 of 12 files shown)


Rasbandit:
Consolidating into umbrella PR #148.

PR #284, opened 2026-05-24
fix(embed): snooze EmbedNote on Voyage 429 (layer 1 of 3)

Rasbandit:
## Summary

- Pattern-match `{:error, {429, _}}` from `Indexing.index_note/2` in `EmbedNote.run_embed/2` and return `{:snooze, N}` instead of `{:error, _}`. Oban's `:snooze` backs the job off WITHOUT incrementing `attempt`, so rate-limit pressure stops burning through `max_attempts: 5`.
- Snooze duration env-driven via `EMBED_429_SNOOZE_SECONDS` (default `60`) — tune as Voyage RPM allotment grows.
- Emits `[:engram, :embed, :rate_limited]` telemetry so layer 3 can hang an alert off it.

## Why

A recent incident produced 1167 discarded `EmbedNote` jobs against a free-tier 3-RPM Voyage bucket. The `ReconcileEmbeddings` 15-min cron eventually re-queued them once paid tier kicked in, but the failure mode is intrinsic to the worker — any future spike (large import, Voyage outage, noisy minute) reproduces it. Background in [`/tmp/handoff-2026-05-24-embed-rate-limit-defenses.md`](#).

## Layer 1 of 3 (each its own PR)

1. **This PR** — root-cause fix: snooze on 429, stop burning Oban attempts.
2. Preventative rate limiter in front of Voyage adapter (RPM via env, `hammer` already a dep).
3. Telemetry alert on Oban discard rate (after verifying PromEx state).

## Test plan

- [x] New unit test: `MockEmbedder` returns `{:error, {429, _}}` → assert `{:snooze, 60}` from `perform_job/2`.
- [x] New unit test: non-429 errors still return `{:error, _}` (preserves existing retry behavior).
- [x] Full suite: 1496 tests, 0 failures, 7 skipped.
- [x] `mix credo --strict` clean for `embed_note.ex`.
- [ ] Manual smoke after merge: rely on prod telemetry — there's no good way to synthesize a Voyage 429 in staging without breaking real users.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -81,6 +81,13 @@ if config_env() != :test do
     config :engram, :query_embed_model, query_model
   end
 
+  # Voyage 429 → snooze duration (seconds). Tune as Voyage RPM grows: lower
+  # values churn jobs faster once budget is restored; higher values are gentler
+  # on a small bucket. Default 60s suits both free-tier (3 RPM) and paid.
+  if secs = System.get_env("EMBED_429_SNOOZE_SECONDS") do
+    config :engram, :embed_429_snooze_seconds, String.to_integer(secs)
+  end
+
   if System.get_env("QDRANT_URL") do
     config :engram, :qdrant_url, System.get_env("QDRANT_URL")
   end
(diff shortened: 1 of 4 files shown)


Rasbandit:
Superseded by #289 — collapsed all 3 layers into a single PR after review-pass surfaced fixes that crossed layer boundaries (notably: L2's shared bucket needed split by purpose, and L3's reason-leak fix needed coordination with how layers reference each other in moduledocs).

PR #286, opened 2026-05-24
feat(embed): client-side Voyage rate limiter (layer 2 of 3)

Rasbandit:
## Summary

- Wraps `Engram.Embedders.Voyage.embed_texts/2` with a token-bucket check against the already-supervised `EngramWeb.RateLimiter` (Hammer/ETS) before the HTTP call.
- Enabled when `VOYAGE_RPM` env var is set. Unset = no throttle — self-host default; nothing changes for users without the env var.
- When the bucket is empty, returns `{:error, {429, %{\"detail\" => \"client_rate_limited\", \"retry_after_ms\" => N}}}` — **the exact same shape** Voyage returns on a real 429, so Layer 1's `EmbedNote` snooze pattern fires identically for either source. No new branches in the worker.
- Bucket key configurable via `:voyage_throttle_key` so async tests use unique keys and don't collide on the shared ETS table.
- Emits `[:engram, :embed, :client_rate_limited]` telemetry for Layer 3.

## Why

Layer 1 (#284) handles the response side — if Voyage 429s, we snooze. Layer 2 prevents most of those 429s from ever happening: fail fast locally against a known bucket size before burning a real API call. Self-tuning is deferred — `VOYAGE_RPM` is env-driven so it can be bumped as the Voyage paid-tier allotment grows.

## Layer 2 of 3

1. ~~Layer 1: snooze on real Voyage 429~~ — #284
2. **This PR** — preventative client-side throttle
3. Layer 3: telemetry alert on Oban discard rate (next)

## Test plan

- [x] New test: `VOYAGE_RPM` unset → no throttle, real HTTP hit.
- [x] New test: bucket exhausted → synthetic 429 returned WITHOUT hitting HTTP.
- [x] New test: under-bucket calls all succeed.
- [x] Existing voyage_test.exs assertions unchanged.
- [x] Full suite: 1497 tests, 0 failures, 7 skipped.
- [x] `mix credo --strict` clean for `voyage.ex`.
- [ ] Manual smoke after merge: set `VOYAGE_RPM=1` in a staging container, watch `[:engram, :embed, :client_rate_limited]` telemetry fire and EmbedNote jobs snooze cleanly.

## Config

```sh
VOYAGE_RPM=2000   # paid-tier RPM; unset for no throttle
```

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -81,6 +81,14 @@ if config_env() != :test do
     config :engram, :query_embed_model, query_model
   end
 
+  # Client-side Voyage rate limit. Unset = no throttle (self-host default).
+  # Set to your Voyage paid-tier RPM (e.g. 2000) to fail fast with a synthetic
+  # 429 before burning real API calls. EmbedNote snoozes on the synthetic 429
+  # just like a real one. Bump this as the Voyage allotment grows.
+  if rpm = System.get_env("VOYAGE_RPM") do
+    config :engram, :voyage_rpm, String.to_integer(rpm)
+  end
+
   if System.get_env("QDRANT_URL") do
     config :engram, :qdrant_url, System.get_env("QDRANT_URL")
   end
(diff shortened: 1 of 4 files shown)


Rasbandit:
Superseded by #289 — collapsed all 3 layers into a single PR after review-pass surfaced fixes that crossed layer boundaries (notably: L2's shared bucket needed split by purpose, and L3's reason-leak fix needed coordination with how layers reference each other in moduledocs).

PR #302, opened 2026-05-25
perf: cut per-request DB round trips (subscription/onboarding/activity/plan caches + notes_cap counter)

Rasbandit:
## Summary

Reduces the DB round trips on every authenticated request. Steady-state authenticated read path drops from ~5 sequential round trips toward ~1-2; the write path loses its per-insert full `COUNT(*)`. All caches are per-node (ETS / `:persistent_term`), matching the Redis-free architecture.

Five independently-scoped commits:

- **P2 — subscription memoization** (`1fe6364`): `has_one :subscription`; `Billing.get_subscription/1` reuses a preloaded assoc; `Auth` plug preloads it once (guarded on `billing_enabled`, so self-host skips). Collapses the 2-4 per-request subscription queries (RequireOnboarding + the two limit plugs) to a single load.
- **P4 — BumpActivity ETS fast-path** (`494137d`): per-node `ActivityCache` holds the last stamp; a request inside the 1h debounce window skips the `usage_meters` read entirely. Cache miss falls back to the authoritative DB read.
- **P5 — plan-limit cache** (`75b4a49`): plan rows are seeded and static at runtime, so cache each plan's `limits` map in `:persistent_term`. `plan_lookup` no longer queries `plans` for API-key traffic. `invalidate/1` exposed for the rare runtime plan edit. Resolution semantics preserved (JSON `false` vs missing key).
- **P3 — TOS-acceptance cache** (`7c4d655`): acceptance is append-only and monotonic per version, so a positive-only `{user_id, version}` ETS cache removes the `user_agreements` read for onboarded users — **no PubSub invalidation needed** (a version bump changes the key).
- **P1 — maintained `notes_count`** (`a2ff68b`): replaces the per-insert `COUNT(*)` with a counter on `usage_meters`. `+1` on insert, `-1` on soft-delete (both atomic in the tenant transaction), bulk `-N` on vault hard-delete. User hard-delete needs no decrement (meter row cascades). Migration backfills live counts for existing users; `recount_notes/1` reconciles drift.

## Migration

`20260525001318_add_notes_count_to_usage_meters` adds `notes_count bigint NOT NULL DEFAULT 0` and backfills (updates existing meter rows + inserts missing ones). Additive and idempotent.

## Test plan

- [x] `mix test` — full suite green (1538 tests, 0 failures)
- [x] `mix compile --warnings-as-errors` clean
- [x] `mix credo --strict` clean
- [x] New coverage: subscription/plan-cache query-count assertions (telemetry), BumpActivity warm-hit no-query, TOS positive/negative caching, notes_count inc/dec/idempotency/cap-enforcement, vault-cleanup decrement, backfill SQL pinned
- [ ] Post-merge: run the migration on FastRaid and smoke-check notes_cap enforcement + onboarding gate

### DIFF
--- a/lib/engram/accounts/user.ex
+++ b/lib/engram/accounts/user.ex
@@ -26,6 +26,7 @@ defmodule Engram.Accounts.User do
     field :og_grandfather_redeemed_at, :utc_datetime_usec
 
     belongs_to :plan, Engram.Billing.Plan
+    has_one :subscription, Engram.Billing.Subscription
     has_many :notes, Engram.Notes.Note
     has_many :api_keys, Engram.Accounts.ApiKey
     has_many :vaults, Engram.Vaults.Vault
--- a/lib/engram/application.ex
+++ b/lib/engram/application.ex
@@ -23,6 +23,7 @@ defmodule Engram.Application do
         EngramWeb.Presence,
         Engram.Crypto.DekCache,
         Engram.UsageMeters.ActivityCache,
+        Engram.Onboarding.TermsCache,
         {EngramWeb.RateLimiter, [clean_period: :timer.minutes(2)]},
         {Oban, Application.fetch_env!(:engram, Oban)},
         clerk_strategy_child(),
--- a/lib/engram/application.ex
+++ b/lib/engram/application.ex
@@ -22,6 +22,7 @@ defmodule Engram.Application do
         {Phoenix.PubSub, name: Engram.PubSub},
         EngramWeb.Presence,
         Engram.Crypto.DekCache,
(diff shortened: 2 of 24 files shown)


Rasbandit:
Follow-ups from the failure-mode audit are tracked (not blocking this PR — happy path is exact and the ETS-absence availability fix landed here):
- #304 — notes_count drift observability/reconciliation, counter concurrency tests, TermsCache revocation invalidate hook
- #305 — pre-existing `add_embed_tokens(_, 0)` reads user 0 (not introduced here)

PR #310, opened 2026-05-26
chore(dev): make dev-ui-staging — iterate UI against staging

Rasbandit:
## Summary
Adds a dev loop for **pure frontend/UI work** that points Vite at the live **staging** backend — HMR on your branch, real Clerk + data, and **no local Phoenix node joining the shared Oban queue**.

- `make dev-ui-staging` — runs Vite under **node** with `--host` and `VITE_API_TARGET=https://staging.engram.page`.
- `vite.config.ts` — proxy `changeOrigin`/`secure` + an IPv4-forcing preamble.
- `docs/context/dev-iteration-loop.md` — full setup, gotchas, and dead ends.

## Why the IPv4 hack
This VM has no IPv6 route, but staging's DNS returns AAAA records, so Vite's proxy happy-eyeballs raced a dead IPv6 connect → `502 Bad Gateway`. Forcing IPv4 (`dns.setDefaultResultOrder('ipv4first')` + `net.setDefaultAutoSelectFamily(false)`) fixes it. **Must run under node** — bun ignores those node DNS APIs, and node needs `VITE_API_TARGET` inline (it doesn't auto-load `.env.local` into `process.env`).

## Setup (one-time, gitignored)
`backend/frontend/.env.local` needs `VITE_AUTH_PROVIDER=clerk` + staging's `VITE_CLERK_PUBLISHABLE_KEY` (instance `longhorn-79`). See the context doc.

## Test plan
- [x] `make dev-ui-staging` serves :5173; `curl :5173/api/billing/config` → 401 (proxy reaches staging)
- [ ] Sign in via Clerk from a browser and confirm authed requests succeed


Rasbandit:
Closing for now. UI-only-against-staging is too niche to maintain a parallel dev path. Root fix for the Oban hazard is an env flag to disable queues in dev — cheaper than a separate target. IPv4/bun-vs-node finding worth salvaging into existing dev-iteration doc if we hit it again.

PR #313, opened 2026-05-26
feat(legal): binding-grade ToS/Privacy acceptance (hash-verified, min_required gate)

Rasbandit:
## Summary
Phase B of binding-grade legal agreement (spec: engram-workspace#32; depends on marketing engram-marketing#95 for canonical hashes).
- `user_agreements.content_hash` column + schema field.
- `Onboarding.accept_terms/6` atomically records ToS + Privacy rows, each with `content_hash` (wrapped in `Repo.transaction` — a ToS row without its paired Privacy row would be an incomplete audit record).
- Gate moved from `current_tos_version` to **`min_required_tos_version`**: a minor (display-only) version bump no longer re-prompts users; only a material bump (raising the floor) does.
- `POST /api/onboarding/accept-terms` now requires `tos_version`, `tos_hash`, `privacy_version`, `privacy_hash` and **verifies each against canonical config hashes — 409 `stale` on any mismatch**, so app/backend drift fails loud instead of recording unverifiable consent.
- `status` exposes `current_privacy_version`.
- Config keys (`runtime.exs`, guarded out of `:test`): `current_tos_version`, `min_required_tos_version`, `current_tos_hash`, `current_privacy_version`, `current_privacy_hash`. Defaults match marketing's manifest hashes.

## Notes
- Frontend (Phase C) will send the new request shape + the `?raw`-hashed `.md` bytes; until it ships, the old single-`version` POST is replaced (coordinated deploy: marketing → backend → app).
- v1 text is provisional pending counsel.

## Test plan
- [x] `mix test` — 1568 tests, 0 failures
- [x] Context TDD: min_required gate (accept between min_required and current still valid; below min_required re-accepts), content_hash + privacy row recorded
- [x] Controller TDD: 409 on ToS hash mismatch, 409 on privacy hash mismatch, 201 + both rows on full match
- [x] Root-caused a config-precedence regression (legal block ran in :test, overriding test.exs) — fixed via the `config_env() != :test` guard

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -314,20 +314,24 @@ end
 #
 # content hashes pin the exact accepted document text; min_required_tos_version
 # is the floor below which an acceptance is no longer considered binding.
-config :engram,
-  current_tos_version: System.get_env("CURRENT_TOS_VERSION", "2026-05-19"),
-  min_required_tos_version: System.get_env("MIN_REQUIRED_TOS_VERSION", "2026-05-19"),
-  current_tos_hash:
-    System.get_env(
-      "CURRENT_TOS_HASH",
-      "6785e725e9900d5f87a90eca362c388d3254dac0e5b7105ba5da29d316551e5c"
-    ),
-  current_privacy_version: System.get_env("CURRENT_PRIVACY_VERSION", "2026-05-19"),
-  current_privacy_hash:
-    System.get_env(
-      "CURRENT_PRIVACY_HASH",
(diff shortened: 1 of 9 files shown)


Rasbandit:
Folded into #309 — the backend (config, onboarding context, controller, migration) was merged into `fix/onboarding-flow` so the binding-grade legal work ships as a single Engram PR (backend + frontend together, as accept-time hash verification requires). Closing in favor of #309.

PR #424, opened 2026-06-03
ci: fold build-and-push-ecr into build-and-publish-image

Rasbandit:
## Summary

Combine the two post-merge-to-main build/push jobs (`build-and-publish-image` → GHCR + `build-and-push-ecr` → AWS ECR) into a single buildx invocation. One image build, layers pushed to both registries, Sentry source maps baked once.

## Why

Pre-fold flow on every `main` push: two side-by-side `runs-on: [self-hosted, isolated]` jobs running the same `docker buildx build` against the same `Dockerfile` and same context. The Phoenix asset compile (bun install + Vite build + esbuild + Tailwind) dominates that cost and was being paid twice per merge.

Buildx supports multi-registry push in a single command — log into both registries, pass `--tag` for each desired image:tag in either registry, push runs the build once and uploads layers per-registry.

## What changed

**`build-and-publish-image` becomes the unified job:**

- Permissions: `packages: write` + `id-token: write` + `contents: read` (was `packages: write` + `contents: read`)
- Concurrency: inherited from pre-fold `build-ecr.yml` — `cancel-in-progress: false` so an in-flight push to either registry isn't dropped on a fast follow-up merge
- Checkout with `fetch-depth: 0` so `git describe --tags --exact-match` can pick up the optional `v*` annotated tag for ECR
- Single `Compute tags and decide build vs skip` step emits the full multi-registry tag list via `tags<<EOF` multi-line output
- Logins: GHCR (token) + AWS OIDC + ECR
- One `docker/build-push-action@v7` step with all tags + Sentry `build-args` + `secrets: sentry_auth_token=…`
- Sentry release steps (frontend + backend) run after the build, same as before

**`build-and-push-ecr` deleted entirely** (was ~180 lines).

**`bump-infra-tfvars` unchanged** — still `needs: build-and-publish-image`.

## Tags pushed (per merge)

GHCR:
- `ghcr.io/engram-app/engram:latest`
- `ghcr.io/engram-app/engram:<semver>` (from `mix.exs`)
- `ghcr.io/engram-app/engram:<7-char sha>`

ECR (`vars.AWS_ECR_REPOSITORY`):
- `…:sha-<7-char sha>`
- `…:v<X.Y.Z>` (only when HEAD has an annotated `v*` tag — same logic as before)

## Behavior preserved

- GHCR semver-exists guard still forces a `mix.exs` bump on first-attempt + cache-miss merges. Cache-hit + rerun paths skip the build for BOTH registries (re-running on the same commit produces a byte-identical image; ECR's existing `sha-` tag already serves the prod GitOps chain).
- Sentry token presence detected via the same env-output indirection (the existing ECR job's workaround for GH Actions' no-`secrets`-in-`if:` rule).
- ECR concurrency group preserved (now `build-and-publish-image-\${{ github.ref }}` at job level).

## Behavior changed (intentional)

- **Sentry source maps + release tagging now also cover the GHCR image.** Pre-fold, only the ECR build called the Vite plugin with `SENTRY_AUTH_TOKEN`; staging-fastraid (which serves the GHCR image) shipped un-symbolicated. After fold, both registries serve the same image with the same release tag → staging errors symbolicate under the same Sentry release as prod. If this is unwanted for any reason, the build-args + secret can be gated on a flag, but folding them in matches what most teams want.
- **ECR push gains the local `/tmp/buildx-cache`.** Pre-fold the ECR job ran cold every time. Post-fold it shares the GHCR job's warm layer cache.

## Risk

- Partial-push: if the GHCR push lands but the ECR push fails (or vice versa), one registry has the new image and the other doesn't. Mitigation: re-running the job is idempotent — the same commit produces a byte-identical image and both registries reconcile to the same content. The `image_exists` guard against the GHCR semver also short-circuits cleanly on rerun.

## Test plan

- [ ] Merge → confirm a single `build-and-publish-image` job runs (not two)
- [ ] Verify all 4-5 tags land (`ghcr.io/.../engram:{latest,<semver>,<sha>}` + `<ECR>:sha-<sha>` + optional `<ECR>:v<X.Y.Z>`)
- [ ] Confirm `bump-infra-tfvars` opens the staging-fastraid PR with the bumped `engram_saas_image_tag`
- [ ] Sentry: a new release tagged with `\${{ github.sha }}` appears for both frontend + backend projects
- [ ] Run-attempt 2 on an already-built version → SKIP path takes (no build, no push, both registries left as-is)
- [ ] Wall-clock check: post-merge build phase ~halved vs pre-fold

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -1974,13 +1974,25 @@ jobs:
             -f description="E2E tests passed (cached fingerprint match)" \
             -f target_url="${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
 
-  # Build + push the release image to GHCR. Container swap on FastRaid
-  # happens downstream: `bump-infra-tfvars` opens a PR in
-  # engram-app/engram-infra bumping `engram_saas_image_tag`; merge of
-  # that PR triggers `tf-apply.yml` which POSTs the daemon's `/tf-apply`
-  # endpoint → `terraform apply` recreates the container on the new tag.
-  # `tf-apply` is now the SOLE image-mover; the legacy `/deploy` endpoint
-  # on engram-deployer is kept live as break-glass only (not called from CI).
+  # Build + push the release image to GHCR (staging-fastraid) AND ECR
+  # (prod ECS Fargate) in a single buildx invocation. Layers build once,
+  # push to both registries; multi-registry login is set up upfront.
+  #
+  # GHCR side feeds the FastRaid staging chain: `bump-infra-tfvars` opens
+  # a PR in engram-app/engram-infra bumping `engram_saas_image_tag`;
+  # merge of that PR triggers `tf-apply.yml` which POSTs the daemon's
+  # `/tf-apply` endpoint → `terraform apply` recreates the container on
+  # the new tag. `tf-apply` is the SOLE image-mover for FastRaid; the
+  # legacy `/deploy` endpoint on engram-deployer is kept live as
+  # break-glass only (not called from CI).
+  #
+  # ECR side feeds the prod GitOps chain: image lives in ECR tagged
+  # `sha-<7chars>` (deterministic deploy handle) plus a `v<X.Y.Z>` tag
+  # when HEAD has an annotated `v*` git tag (marketing/audit milestone —
+  # NOT `release-v*`, that's the deploy trigger handled by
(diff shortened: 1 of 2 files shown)


Rasbandit:
Re-opening under `chore/` prefix — branch name `ci/...` was excluded from verify.yml's push trigger allowlist so no CI ran. See replacement PR.

PR #436, opened 2026-06-04
Tier 2.C: mix engram.preflight self-host CLI

Rasbandit:
## Summary

Adds `mix engram.preflight` — a self-host CLI that previews what `Engram.Release.migrate()` is about to do on the next container start. Operators run it before `docker compose down && docker compose up` to plan downtime windows.

**Output:**
- Pending migrations with their phase tag (`expand`, `migrate-data`, `contract`, `single-shot`, `unknown`)
- Irreversibility flag (detected via `# rollback-irreversible` magic comment)
- Lock-impact estimate (`:low` / `:medium` / `:high`)
- Copy-paste rollback command — only emitted when every pending migration is reversible

**Inside a running container:**

    docker compose exec engram bin/engram eval 'Mix.Tasks.Engram.Preflight.run([])'

## Plan

`docs/superpowers/plans/2026-06-04-migration-safety-tier-2.md` § Tier 2.C. Tier 1 (#432) and Tier 2.D (#434) are independent prerequisites for the wider migration-safety story but this PR does not depend on either.

## Test plan

- [x] 9 ExUnit tests pass: 6 from the spec + 3 added during code review (drop_table / rename_table / modify-with-options all flagged `:high`)
- [x] `report/2` is pure (filesystem + MapSet only) — fully testable without a database
- [x] `run/1` calls `app.start` + `Ecto.Migrator.migrated_versions(Engram.Repo)` — exercised in CI when the project compiles

PR #434, opened 2026-06-04
Tier 2.D: SCHEMA-IMPACT release notes

Rasbandit:
## Summary

Adds a `create-release-notes` job to `deploy-prod.yml` that auto-creates a GitHub Release for every `release-v*` tag whose range contains a `phase/*`-labeled PR. Release notes include the downtime upgrade procedure, list of impacted PRs, and a conditional rollback hint (omitted when any migration in the range is marked `# rollback-irreversible`).

Helper `scripts/generate_schema_impact_notes.sh` is unit-tested via `test/scripts/generate_schema_impact_notes_test.sh` (5 scenarios covering empty, phase-labeled, non-phase, irreversible, multi-label cases).

Self-hosters see the SCHEMA-IMPACT block on the release page before pulling the image — single source of truth for downtime windows and rollback procedures, no need to read individual PR descriptions.

## Plan reference

`docs/superpowers/plans/2026-06-04-migration-safety-tier-2.md` § Tier 2.D — Tier 2 of the migration-safety roadmap. Tier 1 shipped as #432.

## Test plan

- [x] Helper unit tests pass locally (`bash test/scripts/generate_schema_impact_notes_test.sh` — 5/5)
- [x] Dry-run against most-recent release range produces empty output (no phase-labeled merged PRs between `release-v0.5.320` and `release-v0.5.321` — expected; phase labels were only just created on PR #432)
- [ ] Live validation: on the next `release-v*` tag that contains a `phase/*` PR, the workflow should auto-create a Release with the SCHEMA-IMPACT block

PR #437, opened 2026-06-04
Tier 2.B: N-1 backwards-compat CI smoke

Rasbandit:
## Summary

Adds the `n1-compat` CI job: on `phase/expand` PRs, applies this branch's new migrations on top of the previous `release-v*` tag's schema and verifies the migrations apply cleanly. Catches the dominant expand-phase failure mode: `ADD COLUMN ... NOT NULL` without default, incompatible type changes, etc. — issues that a Squawk per-statement scan can miss in the catalog-level view.

**Simpler than the original plan:** the plan called for running the previous release's full `mix test` against the migrated DB. This implementation drops that — cross-version test runs are brittle, the marginal value over "migrations cleanly apply" is small for expand-phase invariants. If we hit a real-world bug that "migrate succeeds + tests would have caught it", we extend at that point.

**Now wired into `ci` aggregator** — a failing n1-compat blocks merge via branch protection (covers a gap found in code review).

## Plan

`docs/superpowers/plans/2026-06-04-migration-safety-tier-2.md` § Tier 2.B. Tier 1 (#432) added the gates; Tier 2.D (#434) and Tier 2.C (#436) are sibling Tier 2 PRs.

## Known overlap with #432

This branch creates `priv/repo/list_new_migrations.sh` locally because the file isn't on `origin/main` yet — it was added in #432, not merged. When #432 merges, this branch will rebase and the helper will deduplicate naturally. No action needed pre-merge of #432.

## Test plan

- [ ] Live validation: this PR has no migrations of its own, so `n1-compat` will short-circuit with "no new migrations" notice. CI should be green with `n1-compat` either skipping (no `phase/expand` label) or short-circuiting (label but no migration files).
- [ ] Future real `phase/expand` PR — n1-compat fires against the prev release's schema and gates the merge.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -1492,6 +1492,179 @@ jobs:
           fi
           exit $failed
 
+  n1-compat:
+    # On phase/expand PRs only: prove the new migrations cleanly apply on
+    # top of the previous release-v* tag's schema. This catches malformed
+    # expand-phase migrations (NOT NULL without default, incompatible
+    # type changes, etc.) that Squawk's per-statement scan can miss in
+    # the catalog-level view.
+    #
+    # Simpler than the original plan: applies migrations, doesn't run
+    # the prev release's full mix test (brittle across versions). The
+    # "migrate applies cleanly" check is the load-bearing invariant.
+    if: |
+      github.event_name == 'push' &&
+      github.ref != 'refs/heads/main' &&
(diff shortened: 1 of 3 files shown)


Rasbandit:
Rolled into #432 — Tier 1 + 2 are being merged as one combined PR. The n1-compat job + ci-aggregator wire are now commit fa086e2 on the #432 branch (the local duplicate of priv/repo/list_new_migrations.sh was dropped — the tier-1 version is canonical).

PR #532, opened 2026-06-12
fix(frontend): don't show free-tier caps/upgrade framing on self-host

Rasbandit:
## Problem

Self-host runs with billing disabled (`billingEnabled=false`) and **unlimited** connections. But the backend's `/billing/status` still reports `tier: "free"` for any subscription-less user (`Billing.tier/1` only returns `:starter`/`:pro` for an active subscription; everything else is `:free`). Several frontend sites compared `tier === 'free'` directly — so self-host wrongly rendered free-tier gating:

- **Onboarding tools step** (`/onboard/tools`) — forced single-select with "Free tier — pick 1 to start. Upgrade…" (the Upgrade link points at `/onboard/billing`, a step that doesn't even exist on self-host)
- **App sidebar** (editor pane) — "Free tier — 1 connection. Upgrade" footer
- **Onboarding checklist widget** — "You're on Free — 1 connection. Upgrade" footer
- **Connections page** — fell back to a 1-connection cap
- **Note viewer** — gated image / PDF / other non-text attachment embeds behind a fallback lock

Backend was already correct: `limits_enforced=false` on self-host → `Billing.effective_limit/2` returns `:unlimited` → `/billing/status` caps are `null`. This was purely a frontend display/gate bug keyed off the raw tier.

## Fix

New single source of truth `useIsFreeTier()` that ANDs the tier check with `billingEnabled`:

```ts
export function useIsFreeTier(): boolean {
  const { billingEnabled } = useConfig()
  const { data } = useBillingStatus()
  if (!billingEnabled) return false
  return data?.tier === 'free' || data?.tier === 'none'
}
```

All five sites route through it instead of comparing `tier` directly. Notably, the note-viewer change **unblocks image/PDF embeds on self-host**.

## Tests

- New `useIsFreeTier` unit test: SaaS free/none → `true`, paid → `false`, **self-host → `false`**, loading → `false`
- Self-host cases added to the tools step, sidebar footer, and note-view attachment rendering tests
- Existing SaaS-tier tests updated with a `billingEnabled: true` config mock and still pass (57 tests green across the 6 files)

Verified live in the running self-host web app: the onboarding tools step now allows multi-select with no upgrade banner.

### DIFF
--- a/frontend/src/billing/use-is-free-tier.test.ts
+++ b/frontend/src/billing/use-is-free-tier.test.ts
@@ -1,0 +1,54 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+import { renderHook } from '@testing-library/react'
+
+let billingEnabled = true
+let tier: string | undefined = 'free'
+
+vi.mock('../config-context', async () => {
+  const actual = await vi.importActual<typeof import('../config-context')>('../config-context')
+  return { ...actual, useConfig: () => ({ billingEnabled }) as ReturnType<typeof actual.useConfig> }
+})
+
+vi.mock('../api/queries', async () => {
+  const actual = await vi.importActual<typeof import('../api/queries')>('../api/queries')
+  return { ...actual, useBillingStatus: () => ({ data: tier ? { tier } : undefined }) }
+})
+
+import { useIsFreeTier } from './use-is-free-tier'
+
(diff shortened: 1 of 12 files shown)


Rasbandit:
Superseded by #536 (both fixes bundled, rebased onto current main).

PR #535, opened 2026-06-12
fix(frontend): show file tree when a vault has root notes but no folders

Rasbandit:
## Problem

Creating a new note at the vault **root** (no folders) left the file pane showing **"No notes yet."** — the just-created note was invisible in the tree.

`viewer/folder-tree.tsx` gated its empty state on folders alone:

```jsx
if (!folders || folders.length === 0) {
  return <p>No notes yet.</p>
}
```

Root-level notes come from a separate source (`useFolderNotes('')` → `rootNotes`) that the loader stitches under ROOT. With zero folders, the guard short-circuited before that rendered — so any vault containing only root notes looked empty.

## Fix

Render the empty state only when there are **neither** folders **nor** root notes:

```jsx
if (!folders || (folders.length === 0 && rootNotes.length === 0)) {
```

## Tests

- zero folders + one root note → the note renders as a `treeitem`, no empty state
- zero folders + zero root notes → "No notes yet." still shown
- existing folder/root-note render tests still pass (8 total)

Verified live in the self-host web app: a doc created at root now appears in the file pane immediately.

(Includes the mandatory `mix.exs` version bump 0.5.395 → 0.5.396.)

### DIFF
--- a/frontend/src/viewer/folder-tree.test.tsx
+++ b/frontend/src/viewer/folder-tree.test.tsx
@@ -14,16 +14,35 @@ vi.mock('sonner', () => ({
   toast: { error: vi.fn(), success: vi.fn(), info: vi.fn() },
 }))
 
+const DEFAULT_FOLDERS = [
+  { id: '1', parent_id: null, name: 'Projects', count: 1 },
+  { id: '2', parent_id: null, name: 'archive', count: 0 },
+]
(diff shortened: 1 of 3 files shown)


Rasbandit:
Superseded by #536 (both fixes bundled, rebased onto current main).

PR #548, opened 2026-06-12
perf: embed batch splitting + global reconcile query + Redix cache pool

Rasbandit:
## Summary

Wave-2g of the 2026-06-12 performance audit — background pipeline scaling.

**Embed batch splitting** (`Indexing`)
`prepare_index` embedded ALL of a note's chunks in one Voyage call: a 10MB-class note (~5k chunks) is a guaranteed 4xx (1,000-input API cap) that retries can never fix — the job then churns through ReconcileEmbeddings forever, burning RPM budget. The matching single Qdrant PUT carried tens of MB of vectors. Now ≤128 texts per embed call, ≤256 points per upsert.

**ReconcileEmbeddings single query**
Loaded EVERY vault then ran one stale-notes query per vault every 15 minutes — O(total vaults) queries at scale. Now one query (inner join on live vaults, global 500 cap, oldest-first). Per-vault fairness isn't needed: EmbedNote is uniq-deduped and backlogs drain across ticks.

**Redix cache pool**
`Engram.Cache.Redix` owned ONE connection — a per-node serialization point where load flipped the 250ms fail-open exactly when traffic was heaviest. Now 8 connections, callers pid-hashed (per-caller ordering preserved). Pool names precomputed at compile time (no runtime atoms). Hammer's rate-limiter connection is library-managed and stays single-conn — upstream pooling tracked as follow-up.

## Test plan
- [x] RED-first: batch-size assertions via embedder/Bypass capture (≤128 / ≤256, totals reconcile), query-count test (1 notes query, 0 vaults queries), global-cap test (505 notes across 2 vaults → 500 jobs), pool child_spec shape + TLS propagation + pid-routing distribution
- [x] credo --strict / format / dialyzer clean; full suite running (will report before merge)


Rasbandit:
Full suite (PG18): 2567 tests, 0 failures, 7 skipped — completes the test-plan checkbox.

Issue #558, opened 2026-06-13
Clerk Orphan Reaper cron silently stopped (last run 2026-06-02) — quota exhaustion returns


Rasbandit:
The hourly `clerk-orphans.yml` schedule hasn't fired since 2026-06-02 (`gh run list --workflow clerk-orphans.yml` shows nothing newer). Ten days of leaked `e2e-*` users re-exhausted the Clerk dev instance's 100-user quota on 2026-06-12, killing e2e-clerk runs with `403 api.clerk.dev/v1/users` (the 2026-04-13 incident pattern from `docs/context/e2e-debugging.md`).

Immediate recovery done manually: ran `e2e/scripts/cleanup_clerk_users.py --older-than 1h` with the SOPS `ci.enc.yaml` key — deleted 81 orphans.

To fix:
1. Find out why the schedule stopped (GitHub disables schedules in some cases; runner-pool changes around 2026-06-02?).
2. Add `workflow_dispatch:` to `clerk-orphans.yml` so manual recovery doesn't require local SOPS access.
3. Consider an alert when the reaper hasn't run in >3h (the silent-stop failure mode is exactly what monitoring is for).

Rasbandit:
Correction after digging in: the reaper was **not** stopped. It's a job in the consolidated `cron.yml` (the standalone `clerk-orphans.yml` was folded in ~2026-06-02 — that's the "last run 2026-06-02" I mis-read). The `cron` workflow fires daily and the reaper has been deleting users successfully (the 06-12 run reaped dozens).

Real root cause = **cadence**: consolidation changed it from hourly → once-daily, and a heavy-CI burst (release day + e2e-rerun storm) mints 100+ `e2e-*` users within hours, blowing the 100-user Clerk dev cap before the daily sweep. Fix in #562 restores hourly. The `workflow_dispatch` recovery path already exists too: `gh workflow run cron.yml -f task=clerk-orphans`.

Rasbandit:
Recurred 2026-06-19. e2e-clerk is failing **112/112 with `403 Forbidden from api.clerk.dev/v1/users`** in fixture setup (no tests run) — the Clerk dev instance is at its 100-user limit again. Seen on diag run 27817955407 (and contributing to today's cross-branch e2e-clerk reds).

The orphan reaper is evidently not reaping (this issue's own title predicted "quota exhaustion returns"). A heavy run-volume day (many e2e-clerk reruns while root-causing the WS-broadcast flake) accumulated users past the cap.

[...]

PR #585, opened 2026-06-15
fix(e2e): Sync Center/modal UI test updates + Clerk leak fixes

Rasbandit:
Two related e2e fixes (folded together per request). Version 0.5.437 → 0.5.438.

## 1. Plugin UI test updates (was the original scope)
The plugin PR (engram-app/Engram-obsidian#101) renames the sync-modal "Merge" → "Sync" and redesigns the Sync Center issues panel into per-reason cards. #101's `backend/e2e` check runs against backend **main**, so these must land here first.
- `test_51`: dispatch param `"Merge"` → `"Sync"`.
- `cdp.get_issue_groups()`: read `.engram-sync-center-card` / `-card-title` (cards) instead of the removed `.engram-sync-center-group`.
- choice spy: neutralize `runSyncWithProgress` so the new live progress modal doesn't linger across parametrized runs.

## 2. Clerk test-user leak + 429 resilience (root-caused this session)
Recurring Clerk `403` (quota) / `429` (rate-limit) e2e failures. #558 fixed only the dead reaper cron; these are the upstream leaks:
- **A — no backoff on Clerk Backend API.** A burst of runs gets 429'd on `POST /users`, fails the suite mid-run, which skips per-test cleanup → leaks. Added a urllib3 `Retry` adapter (429 + transient 5xx, exponential backoff, honors `Retry-After`).
- **B — orphan-on-partial-failure.** `create_user` created the user, then awaited `_wait_until_session_ready`; if that threw, the id never reached the caller's `try/finally` → orphan. Now best-effort deletes it before re-raising.
- **C — brittle prefix lists.** Both the in-run sweep and the reaper hard-coded email prefixes, and new suites kept being missed (`e2e-conn-` was absent from **both** → test_71's ~10 users/run never swept). Replaced with one shared predicate `helpers/clerk_constants.is_e2e_clerk_email` (`e2e-` local part + `+clerk_test` tag) — robust to new suites, can't match a real user.

Verified via mock (retry config, orphan-delete, predicate) + the reaper script still imports/runs. Does **not** retroactively delete existing orphans — clear those via `cleanup_clerk_users.py` or the hourly reaper.

### DIFF
--- a/e2e/helpers/cdp.py
+++ b/e2e/helpers/cdp.py
@@ -525,6 +525,12 @@ class CdpClient:
                 const p = {PLUGIN_PATH};
                 if (p.__origRunSyncFromChoice) return 'already-installed';
                 p.__origRunSyncFromChoice = p.runSyncFromChoice.bind(p);
+                // runSyncWithProgress wraps runSyncFromChoice with a live
+                // progress modal. Bypass it under the spy so the dispatch test
+                // doesn't leave a (never-completing, swallowed) progress modal
+                // mounted across parametrized runs.
+                p.__origRunSyncWithProgress = p.runSyncWithProgress.bind(p);
+                p.runSyncWithProgress = async (choice) => p.runSyncFromChoice(choice);
                 p.__lastSyncChoice = null;
                 const swallow = {swallow_js};
                 p.runSyncFromChoice = async (choice) => {{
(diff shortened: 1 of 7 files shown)


Rasbandit:
Superseded by the same-named branch PR so backend↔plugin e2e auto-pairs with Engram-obsidian#101 (legitimately green, no override). Same commits.

Issue #605, opened 2026-06-15
Backfill worker for notes_fts keyword index (+ Tiger Cloud migration tool)


Rasbandit:
Deferred from #595 (keyword search). The hybrid keyword leg indexes notes via `EmbedNote` on create/edit; existing un-indexed notes are **not** backfilled. Deferred deliberately pre-launch because "no data is sacred" — wiping + re-syncing a vault repopulates `notes_fts` through the normal embed path.

## What to build

A parameterized Oban backfill worker (cursor + batch, mirroring `Engram.Workers.BackfillContentHashHmac`):

- Cursor over `notes` (optionally scoped per user/vault), decrypt `content`/`title`, `KeywordIndex.upsert/1`. **No Voyage call** (this is not a re-embed).
- Per #595's perf note: for a large corpus, drop/disable the GIN index → bulk-load `notes_fts` rows → rebuild the GIN once at the end, rather than N incremental `fastupdate` merges.
- **Target-parameterized**: the same job is the future migration tool for the managed-BM25 store (Tiger Cloud `pg_textsearch`) — see the decision log on #595. It should accept a target keyword-index adapter so it can backfill `notes_fts` today and a remote BM25 store later.

## Why it's not launch-critical

New/edited notes self-index. A wipe+re-sync repopulates everything. This is purely for backfilling an existing corpus without a wipe, and for the eventual Tiger Cloud cutover.

Refs #595

Rasbandit:
Superseded: notes_fts was never built — keyword search shipped on Qdrant sparse (PR #610), not Postgres tsvector. A re-index/backfill for the sparse path is tracked under #613 (clear embed_hash to force reindex). Closing this notes_fts-scoped issue.

Issue #595, opened 2026-06-15
Hybrid full-text/keyword search (Postgres tsvector) alongside vector search


Rasbandit:
## Problem

No keyword / full-text / substring search exists today. Content is stored as ciphertext, so Postgres cannot `LIKE`/`tsvector` on it directly; retrieval is vector-only (Qdrant) plus HMAC exact-match on tags/folder/path. Users expect literal keyword search ("find the note that says `PADDLE_API_KEY`").

Dense embeddings cannot fill this gap: they encode meaning, not tokens, and face-plant on exact strings, identifiers, code, typos, and boolean queries — exactly where keyword search wins. The two indexes fail in opposite directions, so we run both (hybrid).

## Approach

Build a `tsvector` from the **decrypted plaintext** at index time (same `EmbedNote` pass that already decrypts to embed), GIN-indexed, stored encrypted-at-rest at the Postgres storage layer. Merge keyword ranking + vector similarity for **hybrid retrieval**.

## Why Postgres

Already on PG; avoids standing up another search store; tsvector + GIN is fast at per-user vault scale. The keyword index is also cheaper than the vector store we already run (a 1024-dim float32 vector is ~4 KB/chunk; the tsvector + GIN adds only ~0.5–1.3× the plaintext text size).

## Implementation specifics (capture before anyone ships the naive version)

**Storage shape — side table, not a column on `notes`:**
- Put the `tsvector` in a `notes_fts` side table (or at minimum a dedicated column), NOT inline on the hot `notes` row — keeps the frequently-read note row narrow.
- It **cannot** be a Postgres `GENERATED ... STORED` column: a generated column can't read `content_ciphertext`. Compute the tsvector **in the app after decrypt** (in the EmbedNote pass) and write it explicitly.
- `strip()` the tsvector only if we forgo phrase search; keep positions if we want phrase (`<->`) queries.

**Backfill — build the GIN index once, after the bulk load (this is the one real perf hazard):**
- Re-indexing the whole corpus via per-row GIN inserts hammers the index. Instead: drop/disable the GIN index → bulk-load the `tsvector` rows in batches → build the GIN index in one shot at the end (far cheaper than N incremental updates).
- For steady-state sync (notes trickling in) GIN `fastupdate` keeps individual writes cheap; raise `gin_pending_list_limit` if merge spikes show up.
- Safe to run post-launch (server can decrypt the whole corpus at any time).

[...]

PR #606, opened 2026-06-15
feat(search): hybrid keyword + vector search (tsvector + RRF) (#595)

Rasbandit:
Implements the keyword/full-text leg of search and fuses it with the existing vector (Qdrant) leg. Closes the core of #595; full decision + rationale live in the [#595 decision comment](https://github.com/engram-app/Engram/issues/595#issuecomment-4706713327).

## Approach (the short version)

Native Postgres `tsvector` + `ts_rank_cd` on the existing RDS, behind a swappable `Engram.KeywordIndex` behaviour. True BM25 is **not installable on managed RDS** (needs `shared_preload_libraries`) and the cheap "ParadeDB replica" path is broken by our ciphertext-at-rest, so BM25 is deferred to managed Tiger Cloud `pg_textsearch` at scale — the `KeywordIndex` seam means that's a one-adapter swap, not a rewrite. Vector store stays on Qdrant.

## What's in here

- **`notes_fts` side table** (`phase/expand` migration): note-level `tsvector`, GIN index, RLS `tenant_isolation_notes_fts`. Computed in-app after decrypt (a `GENERATED` column can't read `content_ciphertext`); plaintext-derived, encrypted at rest at the storage layer.
- **`Engram.KeywordIndex` behaviour + `.Postgres` adapter**: title weighted `A` / body `B`, `websearch_to_tsquery('english', …)`. Writes via the trusted worker role (like chunk inserts); **searches inside `Repo.with_tenant` so Postgres RLS enforces tenant isolation** row-level.
- **`EmbedNote` hook**: indexes the keyword leg in the same already-decrypted pass (best-effort — a keyword failure logs + emits telemetry but doesn't fail the embed / trigger a wasteful re-embed).
- **`Engram.Search.Rrf`**: reciprocal rank fusion (k=60, tunable weights), fuses on rank so `ts_rank_cd` vs cosine scales never reconcile.
- **Hybrid `Search`**: `?mode=keyword|vector|hybrid`. Web endpoint defaults to **hybrid**; legs run concurrently (`Task.async`); keyword-only hits get a `ts_headline`-highlighted snippet. Internal `Search.search` default stays `:vector` so the **MCP path and other callers are unchanged**.

## ⚠️ Client-facing note (plugin / web)

On the **hybrid** path the result `score` is now the **RRF fused value** (rank-meaningful, e.g. ~0.016), not raw cosine (~0.95). Ordering is preserved and correct; clients that sort by score are fine, clients that threshold on an absolute cosine value are not. The MCP and `?mode=vector` paths still return cosine.

## Deferred

- **Backfill worker → #605.** Existing un-indexed notes aren't backfilled; new/edited notes self-index, and a wipe+re-sync repopulates (pre-launch, "no data is sacred"). #605 is also the future Tiger Cloud migration tool.
- **Cross-vault keyword** leg (single-vault for v1) → cross-vault hybrid runs vector-only.
- Honest labeling: it's `ts_rank_cd` (TF/proximity), **not** BM25.

## Tests

TDD throughout. Full suite green locally: **2771 tests, 0 failures**. New: `keyword_index` (4), `rrf` (4), keyword/hybrid `search` modes (3), `embed_note` keyword hook (1); existing `search_controller` updated for the hybrid score contract.

Refs #595

### DIFF
--- a/lib/engram/keyword_index.ex
+++ b/lib/engram/keyword_index.ex
@@ -1,0 +1,39 @@
+defmodule Engram.KeywordIndex do
+  @moduledoc """
+  Behaviour for the keyword (full-text) leg of hybrid search (#595).
+
+  This is the swap seam. Today the only impl is `Engram.KeywordIndex.Postgres`
+  (native `tsvector` + `ts_rank_cd` on the primary RDS). When keyword relevance
+  becomes a measured bottleneck at scale, a `KeywordIndex.TigerCloud`
+  (`pg_textsearch` BM25) adapter drops in behind the same callbacks — call
+  sites in `Engram.Search` and `Engram.Workers.EmbedNote` never change. See the
+  decision log on issue #595.
+
+  Resolve the configured impl with `module/0`.
+  """
+
+  @typedoc "Plaintext-populated note (persisted row + decrypted `content`/`title`)."
+  @type note :: Engram.Notes.Note.t()
+
+  @typedoc "Tenant + vault scope for a search. `vault_id: nil` searches all of the user's vaults."
+  @type scope :: %{required(:user_id) => String.t(), optional(:vault_id) => String.t() | nil}
+
+  @typedoc "A ranked hit: the note id and its keyword relevance score (higher = better)."
+  @type hit :: {note_id :: String.t(), rank :: float()}
+
+  @doc "Index (insert-or-replace) a note's plaintext into the keyword store."
+  @callback upsert(note()) :: :ok | {:error, term()}
+
+  @doc "Remove a note from the keyword store."
+  @callback delete(note_id :: String.t()) :: :ok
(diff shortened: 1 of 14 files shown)


Rasbandit:
Superseded by #610, which implements the keyword leg as a per-chunk Qdrant sparse vector keyed with HMAC(user_DEK, token) instead of a Postgres `tsvector`. The tsvector approach was a plaintext index at rest (a DB dump reconstructs every note's bag-of-words), breaking the encryption-at-rest posture; the Qdrant-sparse approach stores only non-reversible keyed fingerprints, delivers real BM25 (Qdrant IDF + client TF/length-norm), is multilingual (incl. CJK), and eliminates the deferred Tiger Cloud migration. See #610.

PR #622, opened 2026-06-16
feat(sync): change-log backbone PR A — per-vault seq stamping

### LINKED ISSUE #614: rename_folder/delete_folder use two separate transactions — partial-state risk (renames commit without tombstones)
Surfaced during sync-backbone PR A code review (branch `feat/sync-change-log-backbone`).

`Notes.do_rename_folder/5` runs the row updates in one `Repo.with_tenant` transaction and the old-path tombstone `insert_all` in a **separate** `with_tenant` transaction. If the second transaction fails, the renamed rows commit but the old-path tombstones never land — poll/offline clients then miss the delete signals for the vacated paths (the exact resurrection class the sync backbone is closing). `do_delete_folders` / batch folder ops likely share the shape.

Pre-existing (not introduced by seq stamping; the seq reuse across the two transactions is itself correct). Low impact pre-launch, but it undercuts the change-log convergence guarantees once the backbone lands.

**Fix:** fold the folder rename/delete updates + tombstone insert into a single transaction so the whole logical op commits atomically. Adjacent to the existing non-transactional-broadcast follow-up note in `notes.ex` (~line 2254).

Relates to the sync change-log backbone (spec `docs/superpowers/specs/2026-06-16-sync-change-log-backbone-design.md`) and the attachment/rename-tombstone reframe in #608.

Rasbandit:
## Summary

First PR of the **sync change-log backbone** (design: `docs/superpowers/specs/2026-06-16-sync-change-log-backbone-design.md`). Adds a per-vault monotonic `seq` to notes/attachments, allocated from a `vaults.change_seq` counter **in the same transaction** as each row write, and stamps it on every write path. This is the ordering substrate the later PRs build on; **the timestamp `/changes` feed stays primary** — nothing reads `seq` yet, so this is additive and safe to ship alone.

Closes #614.

## What's in this PR

- **Schema (phase/expand):** `vaults.change_seq bigint NOT NULL DEFAULT 0`; nullable `notes.seq` / `attachments.seq`; indexes `(vault_id, seq, id)` (the `id` tiebreaker keeps PR B's keyset cursor pull an index-ordered scan despite bulk ops sharing a seq); `vaults` `fillfactor=80` so the per-write counter bump stays a HOT update.
- **Allocator:** `Engram.Vaults.next_seq!/1` — atomic `UPDATE vaults SET change_seq = change_seq + 1 RETURNING`, run inside the caller's `with_tenant` transaction (RLS-scoped; row lock serializes per vault). Excluded from the changeset cast so no client attr can clobber it.
- **Stamping:** every note/folder/attachment write — single insert/update, soft-delete, rename, folder-marker create/delete, folder rename (cascade + tombstones), folder delete cascade, batch insert, attachment upsert/delete. One seq per bulk op (shared); one per single op.
- **#614 fix:** folder rename now commits cascade row-updates **and** old-path tombstones in a **single transaction** at one seq — previously two transactions, where a cursor pull between commits could miss the tombstones (resurrection).
- **Backfill (phase/migrate-data):** assigns `seq` to any pre-existing rows per vault and advances each `change_seq` to its max; idempotent (`seq IS NULL`-gated).

## Not in this PR (rollout plan)

PR B = `vault_device_cursors` + `?cursor=` pull + device id · PR C = snapshot bootstrap + `HISTORY_EXPIRED` · PR D = Oban low-water-mark compaction · PR E = deprecate the timestamp feed. **#614's atomic-folder-rename is now landed here**, ahead of PR B which would have exposed it.

## Test plan

- [x] `mix test` — 2783 tests, 0 failures
- [x] Per-write-path seq tests (notes/folders/attachments/batch) + backfill test, incl. a single-`seq > cursor`-read assertion that renamed rows + tombstones are visible together
- [x] `mix format`, `credo`, `sobelow`, migration Squawk lint — clean
- [x] Migration `down` paths verified; `structure.sql` correctly untouched (baseline-snapshot repo)
- [ ] CI: phase-label gate (expand + migrate-data), migration immutability, N-1 compat


Rasbandit:
Split the backfill migration out to its own phase/migrate-data PR (tracked above) so this PR carries exactly one phase label (`phase/expand`) per AGENTS.md. Also bound the `insert_all` return in `do_rename_folder` to fix a Dialyzer `unmatched_return` introduced by the #614 atomicity merge. This PR is now expand-only: schema + seq stamping; no data backfill.

PR #624, opened 2026-06-17
refactor(test): ServiceConfig seam — flip 8 Voyage/Qdrant files to async

Rasbandit:
## What

Adds `Engram.ServiceConfig` — a per-process read-seam for the Voyage embedder + Qdrant client config (base URLs, API keys, peer knobs) — and routes both clients' config reads through it. Then flips **8 test files** from `async: false` → `async: true`.

Stacks on #620 (base = `fix/bootstrap-lock-async`).

## Why

Tests that point these clients at a per-test `Bypass` via `Application.put_env/3` were forced `async: false`: two async tests would clobber the same global `:voyage_url` / `:qdrant_url` key mid-flight. A **per-process** override lets them run async.

Mechanism mirrors `Mox` / `Ecto.Sandbox`: overrides live in a shared ETS table keyed by owner pid, resolved via `[self() | $callers]` (covers direct calls, inline `perform_job/2`, and `Task` children). **Compile-time gated** (`@is_test_build`) — in prod, `get/2` is a plain `Application.get_env`, byte-identical, zero hot-path cost; `put_override/2` doesn't exist.

## Converted (all green @ seeds 0/1/42)

- `voyage_test`, `qdrant_test`
- Tier A (pure Bypass): `qdrant_collection`, `qdrant_named_vector`, `qdrant_hybrid`, `prom_ex/qdrant`
- Tier B (DataCase, URL-only): `search_hybrid`, `indexing_keyword`

Plus one latent flake fixed: `DeviceFingerprintTest` correctly marked `async: false` (it makes global telemetry assertions incl. a `refute_received` no-op — only sound in isolation; the added async load exposed it deterministically).

## Honest speedup: marginal

The files safe to convert here are the **fast** ones. The slow serial files (EmbedNote, ConversationMeter, crypto/rotation, ReconcileEmbeddings) all couple on *additional* global state and are **deferred** — they need their own prod-seam / lock-gate work (follow-up issue). The real deliverable is the **seam**, which makes those future conversions possible.

## Deferred (follow-up issue)

| File | Needs |
|------|-------|
| `search_test` | route `:reranker` / `:jina_url` / `:query_embed_model` |
| `indexing_test` | route `:doc_embed_model` |
| `embed_note_test` | route `:embedder` / `:require_phone_for_embed` / `:limits_enforced` |
| `aad_tamper_test` | `DekCache.invalidate_all()` is global |
| `search_integration_test` | integration-excluded + `invalidate_all` + `on_exit` reads service config |
| `user_dek_rotation_test`, `rotate_user_dek_test`, `mix rotate_user_dek_test` | gate `RotationLock` `pg_advisory_xact_lock` (#620-style) |
| `cleanup_vault_test`, `orphan_sweep_test` | route `:storage` + attachments advisory lock |

## Verification

- Full suite green at seeds 0/1/42: **2821 tests, 0 failures**
- `mix dialyzer` clean, `mix credo` clean, `--warnings-as-errors` clean


Rasbandit:
Deferred-conversion follow-up tracked in #625.

PR #631, opened 2026-06-19
feat: attachment actions (delete/rename/move) + uuid-keyed storage + tombstone convergence

Rasbandit:
## Summary

Turns on **delete / rename / move** for attachment rows in the web file tree (single + batch, mixed note+attachment selections), backed by **UUID-keyed S3 storage** and a **durable old-path tombstone** so a move converges on every device — no duplicate, no resurrection.

Spec: `docs/superpowers/specs/2026-06-16-attachment-actions-and-rename-convergence-design.md`
Plan: `docs/superpowers/plans/2026-06-17-attachment-actions-rename-convergence.md`

## What's in

**Backend**
- `Storage.object_key/3` — S3 key by row UUID (`<user>/<vault>/objects/<uuid>`), decoupled from the mutable vault path. New uploads key by UUID; legacy path-keyed rows keep working via the `storage_key` column. Fixes a latent `delete_external` bug (was recomputing the path key instead of deleting the stored column).
- `Attachments.move_attachment/4` — pure-metadata move: repoint the live row (path re-encrypted under the **unchanged** id-AAD, `storage_key`/blob untouched) + insert a soft-deleted **tombstone** at the old path, both stamped with **one `seq` in one transaction** (the #614 discipline). 409 on conflict, idempotent no-op, `:not_found` on miss.
- `batch_move/4` + `batch_delete/3` — all-or-nothing batch ops (`batch_delete` counts only real deletions).
- Real-time: reuses the existing **`note_changed`** socket event with `kind:"attachment"` (the plugin already dispatches by kind) — **no new socket event**. A move broadcasts delete(old)+upsert(new); delete broadcasts only when a live row actually transitioned.
- Endpoints: `POST /attachments/rename` (billing-gated), `/attachments/batch-move` + `/attachments/batch-delete` (under the `IdempotencyKey` pipeline). `batch_delete` intentionally **not** gated (cleanup must never trap a downgraded user).

**Frontend**
- `useRenameAttachment` / `useBatchMoveAttachments` / `useBatchDeleteAttachments` (idempotency headers, real cache-key invalidation, failure toasts).
- Un-suppressed context-menu / long-press on attachment rows; threaded the `attachment` kind through `folder-tree` (path-keyed, not id) + `ATTACHMENT_ACTIONS`.

## Re-scope vs spec

The sync-cursor-pull detour (#622/#628) landed under the spec and already shipped the convergence substrate (per-vault `seq` change-log + tombstone-carrying feed) and the folder-rename tombstone. So this PR:
- **Reuses `note_changed`+`kind`** instead of the spec's separate `attachment_changed` event.
- Leaves the spec's "PR2" (single-note `rename_note` tombstone) as a **separate follow-up PR** — folder rename already has it; single-note rename is the only gap.

## Paired plugin PR

A 2-line plugin fix (clear `syncState`/`baseStore` on socket-driven delete, matching the poll-delete convention) ships on the same-named branch `feat/attachment-actions-and-rename-convergence` — linked below once opened. The plugin otherwise converges attachment moves with zero source change.

## Tests

- Backend: **2860 tests, 0 failures** (full suite). New: storage keying, move/tombstone invariants, batch all-or-nothing rollback, broadcast gating, controller auth/idempotency/conflict.
- Frontend: hook + tree-row vitest green, biome clean.

## Follow-ups (not in this PR)

- **E2E convergence test** (web move → Obsidian, no dup over socket + offline poll) — convergence is unit-covered at every layer; the integration e2e is a tracked follow-up.
- Tombstone GC; legacy attachment blob re-key backfill (optional — reads already route via `storage_key`).

### DIFF
--- a/docs/superpowers/plans/2026-06-17-attachment-actions-rename-convergence.md
+++ b/docs/superpowers/plans/2026-06-17-attachment-actions-rename-convergence.md
@@ -1,0 +1,1109 @@
+# Attachment Actions + Rename Convergence Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Turn on delete / rename / move for attachment rows in the web file tree (single + batch), backed by UUID-keyed S3 storage and a durable old-path tombstone so a move converges on every device with no duplicate or resurrection.
+
+**Architecture:** Decouple the S3 object key from the mutable vault path (key by row UUID). A move/rename becomes pure metadata: repoint the live row's encrypted path + insert a soft-deleted tombstone at the old path, both stamped with one per-vault `seq` in a single transaction. The tombstone surfaces `{old, deleted:true}` through the *existing* `/attachments/changes` poll AND the new seq-cursor pull; the plugin already trashes on `deleted` and writes on upsert, so convergence needs near-zero plugin code. Real-time parity reuses the existing `note_changed` socket event with `kind:"attachment"` (the plugin already dispatches by kind).
+
+**Tech Stack:** Elixir/Phoenix (backend), React + TanStack Query + TypeScript (frontend), TypeScript Obsidian plugin, ExUnit / bun test / Jest, headless-Obsidian e2e.
+
+---
+
+## Re-scope note (READ FIRST — the spec is partly OBE)
+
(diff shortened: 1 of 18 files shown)


Rasbandit:
Paired plugin PR: engram-app/Engram-obsidian#110 (socket-delete syncState cleanup).

Rasbandit:
Follow-up: #632 applies the same tombstone primitive to single-note rename (the spec's PR2).

Rasbandit:
**E2E added** (`4f866055`): `e2e/tests/test_79_attachment_move_convergence.py` — proves a web-originated attachment move converges in Obsidian with no duplicate at the old path, over **both** the live socket and an offline disconnect→reconnect catch-up (the durable-tombstone path). Adds `api.rename_attachment` helper.

Locally: `py_compile` clean, pytest collects both cases. Not run against the live CI stack here — the e2e job should run it paired with plugin branch `feat/attachment-actions-and-rename-convergence` (#110). The convergence itself doesn't depend on the plugin fix (that only guards a stale-syncState echo on path *reuse*), so it should also pass against plugin `main`.

PR #681, opened 2026-06-21
Add consolidated /api/bootstrap + ETS-cached entitlement matrix

Rasbandit:
Collapse the 3-4 serial first-load round-trips (onboarding status, billing
status, vaults) into one GET /api/bootstrap call, and back the user's
resolved capability matrix with a long-lived, explicitly-invalidated ETS
cache so repeated resolution is a single ETS read.

Backend:
- Engram.Billing.EntitlementCache: node-local ETS cache of {tier, limits}
  keyed by user id, 24h TTL. Mirrors OverrideCache (CacheSync cross-node
  evictions + LISTEN on user_limit_overrides_changed). The TTL is a backstop;
  freshness comes from explicit invalidation, not expiry.
- Engram.Billing.capabilities/1: resolves every LimitKeys key via
  effective_limit/2 into a JSON-ready %{tier, limits} map, served through
  EntitlementCache.
- Invalidation wired into the two entitlement-changing sites: the
  subscription-mutation chokepoint (broadcast_subscription_activated/2) and
  the override-expiry sweep, plus the shared PG NOTIFY for out-of-band
  override writes.
- EngramWeb.BootstrapController + GET /api/bootstrap on the onboarding
  pipeline (Auth but not RequireOnboarding) so the SPA can fetch it before
  onboarding completes. Reuses extracted *_payload/1 builders from the
  onboarding/billing/vaults controllers so each slice is byte-identical to
  its standalone endpoint.

Capabilities are advisory UX state; server-side check_limit/3 and
check_feature/2 remain the authoritative gates. Onboarding progress and
usage counts (vaults, connections) stay computed-fresh — only the stable
tier+limits slice is cached.

Frontend:
- useAppBootstrap(): one /bootstrap fetch that seeds the onboarding, billing,
  vaults, and capabilities query caches. Mounted in OnboardingGate (the first
  authed gate) so the seed lands before any vault-scoped view mounts.
- Short staleTime on useVaults/useBillingStatus so the seeded payload isn't
  immediately refetched on mount; mutations still invalidate explicitly.

Tests: EntitlementCache mechanics, capabilities resolution + cache/evict
behavior, the bootstrap endpoint (billing on/off), and the updated gate test.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01MohCsduC8R2YmAq1cGv5Jm

### DIFF
--- a/frontend/src/api/queries.ts
+++ b/frontend/src/api/queries.ts
@@ -619,6 +619,10 @@ export function useBillingStatus() {
   return useQuery({
     queryKey: ['billing', 'status'],
     queryFn: () => api.get<BillingStatus>('/billing/status'),
+    // Seeded fresh by useAppBootstrap on first load; mutations that change
+    // billing invalidate this key explicitly, so a short staleTime just
+    // suppresses a redundant refetch-on-mount of the seeded payload.
+    staleTime: 60_000,
   })
 }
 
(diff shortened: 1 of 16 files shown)


Rasbandit:
Superseded by #690 — same work moved off the `claude/*` branch (CI's `on: push` allowlist only matches conventional prefixes like `feat/`, so this PR never ran checks). #690 is rebased, has the failing test root-caused + fixed, and is running full CI.

Issue #689, opened 2026-06-21
Daily-cap (token bucket) telemetry


Rasbandit:
`Engram.Usage.DailyCap.spend/4` (#680) emits no telemetry (allow/deny/fail-open). Add counters so the daily search cap is observable. Pairs with the rate-limiter telemetry follow-up.

Rasbandit:
Closing — done. `Engram.PromEx.Usage` subscribes `[:engram, :usage, :daily_cap]` and exports `..._daily_cap_total` (prom_ex/usage.ex:31-32). The plugin is wired in `Engram.PromEx.plugins/0`, so the counter reaches the scraped `/metrics` endpoint.

Issue #688, opened 2026-06-21
usage_buckets: clean up rows on account deletion


Rasbandit:
`usage_buckets` (token-bucket cap, #680) has no FK to users, so rows orphan on account deletion (harmless — tiny, self-refilling, never read for a deleted user). Wire a `delete_all` into `Lifecycle.do_hard_delete/2` or an Oban prune.

Rasbandit:
Already implemented in `Lifecycle.do_hard_delete/2` (line 180-183): raw `DELETE FROM usage_buckets WHERE user_id = $1::uuid` inside the commit-point transaction.

Issue #604, opened 2026-06-15
Setup checklist: completed items should check off + strike through, not disappear


Rasbandit:
## Problem

Setup checklist widget removes items from the list the moment they complete. Users lose the sense of progress/accomplishment, and the list silently shrinks — confusing if they're partway through and watching the next step appear without context.

## Desired behavior

Completed items stay visible, rendered with:
- Filled checkbox (☑ or shadcn `Check` icon) instead of empty (☐)
- Strike-through on label (`line-through`)
- Muted color (`text-muted-foreground`)
- Action buttons (Start / Setup guide ↗ / dismiss ×) hidden

## Repro

1. Open app, expand setup checklist (bottom-right pill).
2. Complete any item (e.g. create first vault, install plugin).
3. **Expected:** item stays in list, checked off with strike-through.
4. **Actual:** item vanishes; only the count (`X of Y done`) reflects the change.

## Where

`frontend/src/onboarding/checklist-widget.tsx`:
- **Line 146:** `const visible = items.filter((i) => !i.done)` — drop the filter, render all items.
- **Lines 203-237:** item render — branch on `i.done` for the checkbox glyph + apply `line-through text-muted-foreground` to label + suppress action buttons when done.

Sort order suggestion: keep done items in their original position so the list doesn't reorder under the user's eye. Optionally group done at bottom — pick one and stay consistent.

## Tests

Update `checklist-widget.test.tsx`:
- Completed item still rendered (currently asserts removal)
- Has `line-through` class
- Has checked checkbox glyph
- Action buttons absent
- Progress count unchanged

Filed via dump session 2026-06-15.

Rasbandit:
Already implemented — closing as done.

Verified on `main` (97edc9d6), `frontend/src/onboarding/checklist-widget.tsx`:

- **Line 168** — `const visible = items.filter((i) => !i.dismissed)` filters on `dismissed`, **not** `done`. Completed items stay in the list.
- **Lines 229-233** — `i.done` branches the label class to `text-muted-foreground line-through`.
- **Line 235** — `{i.done ? "☑" : "☐"}` for the checkbox glyph.
- **Line 239** — `{!i.done && (...)}` suppresses the Start / Setup guide / dismiss actions on completed rows.
- Done items keep their original position (no reorder), matching the issue's stated preference.

Test coverage is in place too — `checklist-widget.test.tsx` asserts `line-through` and `☑` at lines 186-187, 227-228, 239-240, 310, 437-439.

[...]

PR #695, opened 2026-06-21
Knock out three contained issues: #604, #689, #688

Rasbandit:
Batches three small, well-scoped issues. Each change ships with the test that pins its behavior. Three commits, one per issue, for clean review/cherry-pick.

## #604 — Setup checklist: completed items check off + strike through, not disappear
`fix(onboarding)` · `frontend/src/onboarding/checklist-widget.tsx` (+ test)

The checklist removed each item the moment it completed, so the list silently shrank and users lost any sense of progress. Completed rows now stay in place — checked off (☑), struck through, muted, with action buttons suppressed — and the widget retires only once nothing is left to act on.

**Note on the issue's "just drop the filter" instruction:** doing that literally would have broken the × button, because `done` conflated genuine *completion* with user *dismissal*. I split them: dismissal still removes a row outright (that's what dismiss means), while a vault created or an MCP/Obsidian connection resolving now leaves the row visible and checked. Tests updated accordingly (the two old "auto-clears / hides on connection" tests now assert the struck-through render), plus new coverage for the checked glyph, `line-through`/`text-muted-foreground` classes, suppressed actions, and an unchanged progress count.

## #689 — DailyCap: allow/deny/fail-open telemetry
`feat(usage)` · `lib/engram/usage/daily_cap.ex`, `lib/engram/prom_ex/usage.ex`, `lib/engram/prom_ex.ex` (+ test)

`DailyCap.spend/4` enforced the daily token-bucket cap but emitted no telemetry. Now emits `[:engram, :usage, :daily_cap]` on every check, tagged by bucket `kind` + `decision`, registered as a counter in a new `Engram.PromEx.Usage` plugin so it lands on the scraped `/metrics` endpoint. `fail_open` is the key signal: a non-trivial rate there means the cap is allowing traffic through during a DB outage and isn't actually enforcing. Cardinality contract honored — `kind` is a fixed label, `decision` is one of three atoms, never user_id.

## #688 — usage_buckets: clean up rows on account deletion
`fix(accounts)` · `lib/engram/accounts/lifecycle.ex` (+ test)

`usage_buckets` is a system table with no FK to users, so the hard-delete Postgres cascade never reached it and a deleted user's rate-limit rows orphaned. They're now deleted explicitly inside the hard-delete transaction (raw `DELETE` — the table has no Ecto schema and is written outside tenant scope), committing atomically with the user-row removal.

## Testing
No Elixir toolchain in the authoring environment, so backend tests run in CI. Tests were written alongside each change (TDD): `daily_cap_test.exs` telemetry cases, `lifecycle_test.exs` usage_buckets purge case, and the `checklist-widget.test.tsx` updates above.

https://claude.ai/code/session_01QRtZakuEpzeG6JoPeXeSFS

---
_Generated by [Claude Code](https://claude.ai/code/session_01QRtZakuEpzeG6JoPeXeSFS)_

### DIFF
--- a/frontend/src/onboarding/checklist-widget.test.tsx
+++ b/frontend/src/onboarding/checklist-widget.test.tsx
@@ -127,7 +127,7 @@ describe('ChecklistWidget — per-tool rows', () => {
     expect(screen.getByText(/connect cursor/i)).toBeInTheDocument()
   })
 
-  it('auto-clears a tool row when a matching MCP connection exists', () => {
+  it('keeps a tool row visible (struck through) when a matching MCP connection exists', () => {
     onboardingStatusValue.data!.profile = { uses_obsidian: false, tools: ['claude', 'cursor'] }
     connectionsValue = {
       data: [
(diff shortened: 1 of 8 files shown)


Rasbandit:
Superseded by #701 — branch re-homed from `claude/*` to `feat/*` so CI runs (the rename auto-closed this PR). Same signed commits.

PR #725, opened 2026-06-24
fix: cascade folder rename/delete/move to attachments + MCP move_attachment

Rasbandit:
## Problem

Folder operations only touched `notes` rows, **silently orphaning attachments** (separate table, encrypted `path`, no folder column). A folder rename/delete/move left attachment rows stranded at their old paths — in the DB and in every synced client. No error; the op reported success.

Three real surfaces were affected: `POST /api/folders/rename`, `POST /api/folders/batch-delete`, `POST /api/folders/batch-move`, plus MCP `rename_folder`. The MCP layer also had no way to move a single attachment.

## Change

- **`Engram.Attachments.rename_folder/4` + `delete_folder/3`** — leaf cascades, decrypt + prefix-match paths, reuse the existing `move_attachment/4` (per-item seq+tombstone, the #614 discipline) and `batch_delete/3`.
- **`Engram.Notes.batch_delete_folders` / `batch_move_folders`** — now additively report the folder paths / `{old,new}` pairs they touched. `Notes` stays note-only.
- **`Engram.Folders` coordinator (new)** — the single home that calls the Notes leg then fans touched folders out to the Attachments leg. REST + MCP repoint to it, so no surface can forget attachments. Exists because `Notes` must not depend on `Attachments` (cycle).
- **MCP `move_attachment` tool** — wraps the existing context fn.

`DELETE /api/folders/:path` (`delete_folder_marker`) is intentionally left alone — it's marker-only and doesn't touch notes either.

## Design notes

- **Per-table consistency, not one cross-table transaction.** #614 is a within-table guarantee (renamed row + tombstone share one seq/txn) — preserved by reusing `move_attachment/4`. Cross-table, a client may briefly see notes move before attachments; sync converges next pull.
- **Accepted residual risk:** on rename, if the note leg commits then the attachment leg conflicts, notes moved but attachments didn't — rare, user-recoverable. Documented in the spec.
- Attachment blob/`storage_key` never moves on rename (path is metadata; blob is content-addressed).

## Tests

TDD per function. 98 touched-suite tests, 0 failures; `mix format` / `credo --strict` / `sobelow` clean. Includes a surface-level drift guard (folder rename moves an attachment end-to-end through the controller) and an empty-`ids` regression test (batch ops return 200, not 500).

Design + plan: `docs/superpowers/specs/2026-06-23-folder-attachment-cascade-design.md`, `docs/superpowers/plans/2026-06-23-folder-attachment-cascade.md`.

### DIFF
--- a/lib/engram/attachments.ex
+++ b/lib/engram/attachments.ex
@@ -317,7 +317,7 @@ defmodule Engram.Attachments do
         # trash are sent.
         _ =
           if deleted? do
-            EngramWeb.Endpoint.broadcast("sync:#{user.id}:#{vault.id}", "note_changed", %{
+            Engram.Sync.Broadcast.emit("sync:#{user.id}:#{vault.id}", "note_changed", %{
               "event_type" => "delete",
               "kind" => "attachment",
               "path" => path,
@@ -665,7 +665,7 @@ defmodule Engram.Attachments do
       "mtime" => att.mtime
     }
 
-    _ = EngramWeb.Endpoint.broadcast("sync:#{user_id}:#{vault_id}", "note_changed", payload)
+    _ = Engram.Sync.Broadcast.emit("sync:#{user_id}:#{vault_id}", "note_changed", payload)
     :ok
   end
 
--- a/lib/engram/attachments.ex
+++ b/lib/engram/attachments.ex
@@ -509,24 +509,50 @@ defmodule Engram.Attachments do
 
   def batch_move(user, vault, paths, target_folder)
       when is_list(paths) and is_binary(target_folder) do
-    Repo.transaction(fn ->
-      Enum.reduce_while(paths, %{moved: 0}, fn old_path, acc ->
+    pairs =
+      Enum.map(paths, fn old_path ->
         base = Path.basename(old_path)
         new_path = if target_folder == "", do: base, else: Path.join(target_folder, base)
+        {old_path, new_path}
+      end)
 
+    # This surface tags conflict/not_found with the offending path so the REST
+    # controller can name it in the 409/404 body (`{:conflict, path}`).
+    case move_pairs(user, vault, pairs, &tag_move_error/2) do
+      {:ok, count} -> {:ok, %{moved: count}}
+      {:error, _} = err -> err
+    end
+  end
+
+  # Shared move-loop for every "relocate these [{old, new}] pairs atomically"
+  # caller (`batch_move/4` + the folder-rename cascade). One `Repo.transaction`
+  # wraps `reduce_while` over `move_attachment/4` (which carries the #614 per-item
+  # repoint+tombstone-share-one-seq discipline); any item error halts and rolls
+  # the WHOLE batch back. `on_error.(reason, old_path)` shapes the rollback value
+  # so each surface keeps its own contract (bare `:conflict` for folder rename,
+  # `{:conflict, path}` for `batch_move`). Returns `{:ok, count}` | `{:error, _}`.
+  defp move_pairs(_user, _vault, [], _on_error), do: {:ok, 0}
+
+  defp move_pairs(user, vault, pairs, on_error) do
+    Repo.transaction(fn ->
+      Enum.reduce_while(pairs, 0, fn {old_path, new_path}, count ->
         case move_attachment(user, vault, old_path, new_path) do
-          {:ok, _} -> {:cont, Map.update!(acc, :moved, &(&1 + 1))}
-          {:error, :conflict} -> {:halt, {:rollback, {:conflict, old_path}}}
-          {:error, :not_found} -> {:halt, {:rollback, {:not_found, old_path}}}
-          {:error, reason} -> {:halt, {:rollback, reason}}
+          {:ok, _} -> {:cont, count + 1}
+          {:error, reason} -> {:halt, {:rollback, on_error.(reason, old_path)}}
         end
       end)
       |> case do
         {:rollback, reason} -> Repo.rollback(reason)
-        acc -> acc
+        count -> count
       end
     end)
   end
+
+  # `batch_move/4`'s error shape: tag the offending path onto conflict/not_found,
+  # pass any other reason through unchanged.
+  defp tag_move_error(:conflict, old_path), do: {:conflict, old_path}
+  defp tag_move_error(:not_found, old_path), do: {:not_found, old_path}
+  defp tag_move_error(reason, _old_path), do: reason
 
   @doc """
   Cascades a folder rename across attachments: every live attachment whose path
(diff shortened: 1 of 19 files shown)


lib/engram/attachments.ex
```diff
@@ -528,6 +528,79 @@ defmodule Engram.Attachments do
...
+          end)
+
+        do_rename_folder_pairs(user, vault, pairs)
+
+      {:error, reason} ->
+        {:error, reason}
+    end
+  end
+
+  defp do_rename_folder_pairs(_user, _vault, []), do: {:ok, 0}
```
Rasbandit: **[Critical] Conflict returns a tuple, not `:conflict` → 500 + split state**

`do_rename_folder_pairs` does `Repo.rollback({:conflict, new_path})`, so `rename_folder/4` returns `{:error, {:conflict, new_path}}` (a tuple). But the coordinator + callers match bare `{:error, :conflict}`. On a real attachment-destination collision during a folder rename, `FoldersController.rename`/MCP hit a `CaseClauseError` → HTTP 500, and the note rows are already renamed → permanent notes/attachments divergence.

Fix: return a shape callers handle — either bare `{:error, :conflict}` from the cascade, or thread the tuple through every caller `case`.

lib/engram/mcp/handlers.ex
```diff
@@ -421,6 +426,17 @@ defmodule Engram.MCP.Handlers do
     {:ok, "Note deleted: #{path}"}
   end
 
+  def handle("move_attachment", user, vault, args) do
```
Rasbandit, line 435: **[High] Missing error clause → 500 on DEK failure**

`move_attachment/4` can return `{:error, reason}` from its crypto `with` head (`ensure_user_dek`/`get_dek`/`dek_filter_key`), but this handler matches only `{:ok,_}` / `:not_found` / `:conflict`. A DEK-load failure → `CaseClauseError` → the MCP tool 500s with opaque text. Add a catch-all `{:error, reason}` clause returning a clean message.

lib/engram_web/controllers/folders_controller.ex
```diff
@@ -276,9 +283,9 @@ defmodule EngramWeb.FoldersController do
         conn |> put_status(400) |> json(%{error: "invalid_ids"})
 
       {:ok, ids} ->
-        case Notes.batch_delete_folders(user, vault, ids) do
-          {:ok, %{deleted: n}} ->
-            body = %{deleted: n}
+        case Folders.batch_delete(user, vault, ids) do
+          {:ok, %{notes: n, attachments: a}} ->
```
Rasbandit, line 290: **[High] batch_delete: attachment-leg failure after notes commit masks data loss**

The coordinator deletes notes (committed) then attachments separately. If the attachment leg errors, the controller skips `Idempotency.remember` + broadcast and 500s. Notes are already gone; peers never get the delete; a retry finds markers already deleted → `{:not_found}` → 404, masking the data loss as “folder not found”. Record idempotency/broadcast off the note result, or reconcile the legs.

Rasbandit:
## High-effort code review — 10 findings (all verifier-CONFIRMED)

Ranked most-severe first. Top inline comments posted on the diff; full list here.

### Must-fix (reachable 500s / data integrity)
1. **`attachments.ex:564` — conflict returns `{:error, {:conflict, new_path}}` (tuple), callers expect bare `{:error, :conflict}`.** On an attachment-destination collision during folder rename → `CaseClauseError` → HTTP 500, with notes already renamed (permanent split). *Root cause of #5.*
2. **`mcp/handlers.ex:429` — `move_attachment` handler lacks an `{:error, reason}` clause.** `move_attachment/4` can return a crypto-failure reason (DEK load) → `CaseClauseError` → 500.
5. **`folders_controller.ex:372` — `send_move_result` `{:error, {:conflict, id}}` matches the attachment leg’s `{:conflict, new_path}`** → 409 returns a file **path** in the `item_id` (UUID) field. Fixed together with #1.
6. **`folders_controller.ex:287` — batch_delete attachment-leg failure after notes commit** skips idempotency + broadcast, 500s; retry → 404, masking note data loss as “folder not found”.

### By-design but worth hardening
3. **`folders.ex:22` — non-atomic cross-table cascade.** Documented accepted-risk for v1; #1’s 500 makes the split-state worse. Fixing #1 (clean error, not 500) mitigates the user-facing part.
4. **`attachments.ex:556` — per-item broadcasts fire before the outer rollback** → phantom move events to clients on a late conflict. Same caveat the existing `batch_move/4` already documents; consistent, not newly introduced.

### Minor (doc / spec / perf / cleanup)
7. **`folders_controller.ex:192` — single `DELETE /api/folders/:path` bypasses the coordinator.** It calls marker-only `delete_folder_marker` (leaves notes too, so nothing is *orphaned*), but the `Engram.Folders` moduledoc overclaims “every folder-mutating surface routes through here.” Soften the moduledoc or route it through.
8. **`attachments.ex:595` — `delete_folder` reuses `batch_delete` (per-attachment seq), diverging from the spec’s “one seq.”** #614 isn’t violated for pure deletes; reconcile spec/moduledoc with the reuse decision (or rewrite as one `update_all`).
9. **`folders.ex:54` — N folders ⇒ N full-vault `list_attachments` scans** (O(N·attachments)). Single scan + partition by prefix.
10. **`attachments.ex:538` — `do_rename_folder_pairs` duplicates `batch_move/4`’s txn+reduce_while scaffolding.** Extract one shared `[{old,new}]` helper so the #614 discipline lives in one place.

*Generated by high-effort multi-agent review (43 agents, finder→independent-verifier). #1/#2 are genuine bugs the per-task + whole-branch internal reviews missed — the conflict path was untested.*

Rasbandit:
## ✅ All 10 review findings fixed (commits `cc940d25`..`4d7ade63`)

Two TDD fix rounds; an independent verification pass (Opus) confirms **ALL RESOLVED, NO REGRESSIONS** — with behavioral (not just structural) tests for the load-bearing ones.

**Correctness (Round 1):**
1. `Attachments.rename_folder` now returns **bare `{:error, :conflict}`** (was a tuple) → no more `CaseClauseError`/500 on attachment-path collision.
2. MCP `move_attachment` + `rename_folder` handlers gained an `{:error, reason}` **catch-all** → clean message on DEK/crypto failure, no 500.
3 + 6. **Cascade is now atomic**: the coordinator wraps both legs in one `Repo.transaction` (`atomic/1`); any leg error `Repo.rollback`s **both tables** (verified: a conflicting rename leaves the note at its old path). No split state, no half-delete.
5. Attachment conflict surfaces as bare `:conflict` → controller returns a clean **409 with no file path in `item_id`**.

**Doc / spec / perf / cleanup (Round 2):**
7. `Engram.Folders` moduledoc scoped to content-mutating surfaces; the marker-only single `DELETE` is explicitly carved out (it deletes no content, so nothing to cascade).
8. `delete_folder`'s per-item seq reconciled in spec + code comment (the #614 same-seq invariant applies only to rename's repoint+tombstone pair; delete atomicity now comes from the coordinator txn).
9. Batch ops now **scan attachments once** and partition by folder prefix (was O(N·attachments)); +2 multi-folder tests.
10. Extracted a shared **`move_pairs/4`** loop backing both `batch_move` and folder-rename, with a parameterized `on_error` so each keeps its distinct error contract + the #614 per-item discipline.

**Residual / follow-up (not blocking):**
- #4 per-item socket broadcasts still fire before an outer rollback (clients self-heal on next pull; DB state is atomic) — documented in the moduledoc, matches the existing `batch_move` caveat.
- *Pre-existing* (not introduced here): nested-folder batch-move `rename_target` first-match-wins ambiguity when a parent **and** its child are selected together — worth a follow-up if that's a supported UI path.

Touched suite: 80 tests, 0 failures; `mix format` / `credo --strict` / `sobelow` clean.

Rasbandit:
## ✅ Remaining two concerns now fixed too (`6db1b732`, `e5c06fed`)

Independently verified (Opus) — **SAFE TO PUSH, no regressions**.

- **#4 broadcast-before-rollback** — new `Engram.Sync.Broadcast`: a transaction-scoped deferral buffer that `atomic/1` brackets around its `Repo.transaction`. Cascade broadcasts now buffer during the txn and **flush in order only on commit, discard on rollback** — no more phantom move events to clients on a rolled-back cascade. Buffer is OFF by default (unconditional `after` cleanup, exception-safe), so every single-item sync path broadcasts immediately exactly as before; the `broadcast_from` SSE echo-suppression path is untouched. Tests: refute-receive on rollback, assert-receive on commit, standalone-still-immediate.
- **Nested-folder batch-move** — attachment→destination resolution is now **longest-prefix (most-specific) match**, so a file under `Docs/Sub/` follows the `Docs/Sub` mapping deterministically regardless of selection order (was first-match-wins).

That closes every finding from the review **plus** the two residuals. Full suite green (115 + 99 + new broadcast tests), format/credo/sobelow clean.

PR #149, opened 2026-05-18
test(e2e): SearchModal indexed-note round-trip

Rasbandit:
## Summary

- Adds `e2e/tests/test_52_search_modal.py` with two tests:
  - **indexed-note round-trip**: write note → `trigger_full_sync` → poll `POST /search` until backend indexes it (up to 30 s) → open SearchModal via CDP → type unique query token → assert result appears in `.engram-search-result-item` list
  - **empty-query state**: open modal with blank input → assert `.engram-search-empty` hint paragraph renders
- Adds `ApiClient.search(query, folder=None)` to `e2e/helpers/api.py` — used by the indexing-wait poll loop before modal interaction
- Bumps `mix.exs` to `0.5.112`

## Selector notes

All selectors use the corrected classes from Task 1 (commit `9735baa`):
- Result items: `.engram-search-result-item` (plan draft had `.engram-search-result`)
- Empty state: `.engram-search-empty` — confirmed in `src/search-modal.ts` `renderEmpty()` and `renderResults()`

## Concerns / known flakiness risks

- **30 s indexing deadline**: the Qdrant embedding pipeline can be slower under CI load. If this test flaps with `Backend never indexed seed note`, raise the deadline to 60 s and back off the poll interval.
- **1.5 s debounce wait**: SearchModal debounce is 300 ms + async fetch. 1.5 s should be sufficient on fast CI but may need tuning on slower environments.

## Test plan

- [ ] CI `build-and-test` passes
- [ ] `backend/e2e` passes (or skips cleanly on pre-SearchModal plugin builds)

> **Stacks on top of #148** — merge after that PR lands (this branch is based off `feat/e2e-01-cdp-helpers` which provides `has_search_modal()`, `open_search_modal()`, `wait_for_search_modal()`, `type_search_query()`, and `get_search_results()`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #776, opened 2026-06-27
fix(crdt): only announce CRDT for markdown — non-md modify regression

Rasbandit:
## The bug (live on staging since the Phase 2 flip)

Under CRDT-default, **modifications to non-markdown notes (`.canvas`) are silently dropped** — they never reach the server. Creates work; modifies are swallowed.

### Root cause

gap③'s `deliver_out` announced `crdt_doc_ready` for **every** note upsert, including non-markdown. For a `.canvas` note the client then:
1. enrolls it into a Yjs doc (it shouldn't — the plugin only routes `.md` to CRDT),
2. `flushFromCrdt`'s it to disk, which marks the path `recentlyFlushed`,
3. so `handleModify` drops the user's **next real edit** as an echo.

### Fix

CRDT manages **markdown only** — gate `deliver_out` on `.md`. This mirrors the plugin's `routeModify` (`.md`-only) and **Relay's `SyncType.Document = "markdown"`** (canvas is a separate, non-markdown-Yjs sync type with its own provider — it's never pulled into the markdown Yjs path). Non-markdown notes stay on the legacy push path.

Pinned via CDP instrumentation (the `recentlyFlushed` guard was swallowing the modify) and **verified on the harness** — the canvas modify now reaches the server (test_41 green).

A paired plugin change adds a matching `.md` guard to `enroll()` (defense-in-depth, mirroring Relay's layered `isDocument` gate).

> Independent of the Phase 3 PR (#775): this fixes a live gap③ regression directly on main.

### DIFF
--- a/lib/engram/notes/crdt_deliver.ex
+++ b/lib/engram/notes/crdt_deliver.ex
@@ -39,8 +39,18 @@ defmodule Engram.Notes.CrdtDeliver do
   @spec deliver_out(String.t(), String.t(), String.t(), String.t(), String.t()) :: :ok
   def deliver_out(user_id, vault_id, path, note_id, content)
       when is_binary(content) do
-    push_to_live_room(note_id, content)
-    announce(user_id, vault_id, path)
+    # CRDT manages MARKDOWN content only — the plugin's routeModify enrolls only
+    # `.md` into Yjs (mirroring Relay's SyncType.Document = "markdown"; canvas and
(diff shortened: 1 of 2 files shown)


Rasbandit:
Folded into #775 (the consolidated CRDT-default completion). The 1-line `.md` deliver gate is cherry-picked there as commit c59f7f11; the plugin half #149 is already merged. Closing to keep the delivery fixes in one green set.

Issue #777, opened 2026-06-27
Flaky unit-tests: CrdtChannelTest :global room terminate poisons the Ecto sandbox


Rasbandit:
## Summary

The `unit-tests` CI job intermittently fails with a `DBConnection`/`client exited` cascade originating in `EngramWeb.CrdtChannelTest`. One CRDT-room test poisons the shared Ecto sandbox connection, and a cluster of **unrelated** tests fail as collateral (e.g. `Engram.Cluster.CacheSyncTest`, `Engram.Usage.DailyCapTest`). It is seed/order-dependent, so a re-run usually clears it — but it has blocked the `unit-tests` check on multiple PRs (e.g. #771, #776) this session and is documented inline in `test/engram_web/channels/crdt_channel_test.exs`.

## Evidence (PR #776 run 28305234956)

```
 1) test crdt_msg second crdt_msg for the same doc reuses the cached room (EngramWeb.CrdtChannelTest)
    ** (stop) exited in: DBConnection.Holder.checkout(...)
       ** (EXIT) shutdown: "owner #PID<...> exited"
 2) test join/3 accepts join for own user_id and vault (EngramWeb.CrdtChannelTest)
 3) test join/3 rejects join for vault belonging to another user (EngramWeb.CrdtChannelTest)
 4) test crdt_msg malformed base64 is silently ignored — no crash (EngramWeb.CrdtChannelTest)
 5) test ... (Engram.Cluster.CacheSyncTest)
 6-10) test ... (Engram.Usage.DailyCapTest)
```
Earlier in the same run: `Postgrex.Protocol (...) disconnected: ** (DBConnection.ConnectionError) client #PID<...> exited`.

## Root cause

A `crdt_msg` in `CrdtChannelTest` lazily spins a real `:global` `CrdtDoc`/`SharedDoc` room (`CrdtRegistry.ensure_started`). The room has `auto_exit: true`, so when the test's channel process dies at end-of-test the room terminates and runs `CrdtPersistence.unbind/3`, which does **DB work** (final snapshot flush). But the room pid is **never granted sandbox access** — the test only does `Sandbox.allow(Repo, self(), channel_pid)`, not the room. The room's unbind therefore checks out a connection that is being torn down → `client exited` → the shared sandbox connection is poisoned → concurrent/subsequent tests fail.

[...]

PR #780, opened 2026-06-28
test: fix #777 :global CrdtDoc room sandbox poisoning + local-test tooling

Rasbandit:
## What

Fixes the order-dependent `unit-tests` flake (#777): `CrdtChannelTest` poisoning
the shared Ecto sandbox, cascading `DBConnection`/`owner exited` failures into
unrelated tests (`CacheSyncTest`, `DailyCapTest`).

## Root cause

Tests on the CRDT path spin up `:global` `CrdtDoc` rooms. A room is a
sandbox-using process **not linked to the test**, so it outlives the test; its
`terminate` -> `CrdtPersistence.unbind/3` (a DB write) runs *after*
`DataCase.setup_sandbox`'s `stop_owner/1` on_exit has torn down the connection's
owner. The room then checks out a dead-owner connection and poisons the shared
sandbox for every later test.

## Fix

`DataCase.setup_sandbox` now registers a second `on_exit` (after `stop_owner`, so
LIFO runs it **first**; gated on shared/non-async mode where the test runs alone)
that synchronously stops every live `CrdtDocSupervisor` room. The room's
unbind then executes while the shared sandbox connection is still checked in.

This self-heals the whole CRDT test class, including future CRDT tests.

## Validation

- Reproduced the poisoning trio (`crdt_channel_test` -> `cache_sync_test` ->
  `daily_cap_test`) failing pre-fix; green 3x post-fix.
- Full suite: **3211 tests, 0 failures**.

## Also

Adds `scripts/test-local.sh` + `docs/context/local-backend-testing.md` for fast
local `mix test` against the docker Postgres (per-worktree `deps.get`,
`DATABASE_URL` unset, optional `--teardown`/`--reset`).

Refs #777

### DIFF
--- a/docs/context/local-backend-testing.md
+++ b/docs/context/local-backend-testing.md
@@ -1,0 +1,62 @@
+# Running the backend ExUnit suite locally
+
+Fast local `mix test` against a docker Postgres — no CI round-trip. The
+`scripts/test-local.sh` wrapper encodes everything below; reach for the doc only
+when the script misbehaves or you want to understand the moving parts.
+
+## TL;DR
+
+```bash
+scripts/test-local.sh                         # whole suite (~190s, 3200+ tests)
+scripts/test-local.sh test/engram/foo_test.exs[:42]   # one file / line
+scripts/test-local.sh --teardown              # drop the test DB after
+scripts/test-local.sh --reset -- test/...     # drop+recreate test DB first
+```
+
+Runs from the main checkout OR any worktree (it resolves its own repo dir).
+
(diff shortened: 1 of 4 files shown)


Rasbandit:
Folded into #775 (the consolidated CRDT-default completion) as commit 3905b4dc. #775 adds CRDT-channel test surface, so it needs the sandbox-poisoning fix in-tree to keep its own unit-tests deterministic — folding makes #775 self-contained and green. The local-test script + doc ride along.

Issue #547, opened 2026-06-12
flaky e2e: note-live-update spec times out under runner contention


Rasbandit:
`[local] › e2e/note-live-update.spec.ts:124 › SPA viewer live-update (#277) › viewer re-renders when a remote client upserts the open note` failed (30s timeout on `toBeVisible`, both retries) on two unrelated perf-wave branches:

- PR #540 run 27407235318 (attachment-internals only — no channel/frontend surface)
- PR #543 run (ws-invalidation branch, post main-merge rerun)

Both occurred while several CI runs shared the runner pool, alongside a `prebuild-ci-image` buildkit `graceful_stop` infra failure — strongly suggests resource contention slows the live-update round trip past the 30s budget rather than a product bug. Teardown noise in the same runs: `DB cleanup failed — FK constraints on test users` (notes_user_id_fkey).

Follow-ups:
1. Root-cause the timing (websocket join vs invalidation vs render) with a trace artifact on next failure.
2. Consider bumping this spec's timeout or gating on a network-idle signal.
3. Fix teardown FK cleanup order (delete notes before users).

Filed per the diagnose-flakes-always rule; rerun-passes are not acceptance.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Rasbandit:
Related cluster: #545 (SyncChannel push_note assert_receive timeout) + #550 (PromEx poller starves sync_channel). All three are assert_receive/timeout flakes under runner contention — candidates for a single CI-load/teardown fix.

Rasbandit:
CI-history check: e2e-local had 0 failures across the last 30 main runs — note-live-update has not recurred. Keep open for a short soak; close if still clean.

Rasbandit:
Closing: e2e-local job has had no failures in recent main runs; note-live-update has not recurred. Reopen if it returns.

Rasbandit:
Reopening — this flake is back on `main` (e2e-clerk red ~4/5 recent runs on the same sha), and new evidence points at a real **delivery** race, not just an assertion-timing race.

**What changed the diagnosis:** `test_78` step 1 used `wait_for_file` (returns the instant the file *exists*), which fails fast as `assert 'HashLive v1' in ''` when the file materializes empty before the CRDT flush. Swapping to `wait_for_content` (poll for the body, 30s) changed the failure to:

```
TimeoutError: File E2E/HashOnlyLive.md did not contain 'HashLive v1' within 30s
[...]

Issue #161, opened 2026-05-19
e2e: modal/event lifecycle races in test_46/51/55


Rasbandit:
## Summary

Three e2e tests rerun on every CI run despite PR #159's flake purge. None block merges (covered by `reruns=1`), but each rerun = wasted wall time + masks real regressions.

Two distinct race families:

### Family A: Obsidian modal lifecycle (test_51, test_55)

| Test | Symptom |
|------|---------|
| `tests/test_51_modal_switch.py` | Modal-stack ordering check fires before DOM has settled after `Modal.open()` returns |
| `tests/test_55_destructive_modal.py` | Same shape — destructive-action confirmation modal queried before mount paints |

**Hypothesis.** `Modal.onOpen()` resolves before the DOM is fully painted. Current tests `await asyncio.sleep(N)` or poll `querySelector` immediately. Need a `wait_for_modal_settled(selector)` helper that:
- polls for the selector to be present
- *and* `getComputedStyle(el).opacity === '1'` (Obsidian uses CSS transition)
- *and* `el.matches(':focus-within')` if focusable

One helper, both tests stabilize.

### Family B: Cross-process subscription handshake (test_46)

`tests/test_46_mcp_broadcast.py` asserts an MCP write triggers a WebSocket broadcast received by the plugin. Likely race: MCP publish lands before plugin's Phoenix channel `join_ref` is acked.

**Hypothesis.** Plugin currently has no externally-observable \"channel ready\" signal. Two paths:
1. Expose `channel.isJoined()` on `plugin.channel` (one-liner in src/channel.ts), poll from test before MCP publish.
2. Or use the same diagnostic-ladder pattern as `wait_for_vault_registered` (issue #159) — bounded poll + active re-join on stall.

## Acceptance

- All three tests pass on 5 consecutive CI runs with `reruns=0`.
- New helper(s) in `e2e/helpers/cdp.py` or `e2e/helpers/modal.py`.
- No `asyncio.sleep` for state convergence in the affected tests.

## Context

Surfaced during PR #159's final green run (run 26086124712). Not fixed there to keep scope tight — those fixes were vault-registration + echo-cooldown deterministic races. Modal/channel timing is a separate family.

Related: #160 (the *other* reason `reruns=1` is still on).

Rasbandit:
## Update from PR #162's CI run

[...]

PR #150, opened 2026-05-18
test(e2e): SearchView sidebar command/ribbon/results (#53)

Rasbandit:
## Summary

- Adds `e2e/tests/test_53_search_view.py` covering the SearchView sidebar (3 tests):
  - `test_command_opens_sidebar` — `open-search-sidebar` command mounts the leaf
  - `test_ribbon_opens_sidebar` — ribbon icon (aria-label "Engram search") opens sidebar
  - `test_sidebar_search_returns_results` — sidebar query returns ≥1 result for an indexed seed note
- Bumps `mix.exs` to `0.5.113`

## Stacks on

- #148 (feat/e2e-01-cdp-helpers — CDP helper foundations)
- #149 (feat/e2e-02-search-modal — SearchModal tests)

## Selector verification

| Selector | Source | Line |
|---|---|---|
| View type id: `engram-search-view` | `src/search-view.ts` `SEARCH_VIEW_TYPE` const | 8 |
| `.workspace-leaf-content[data-type="engram-search-view"]` | Obsidian workspace API convention | — |
| `input.engram-search-input` | `src/search-view.ts` `createEl("input", {cls: "engram-search-input"})` | 44 |
| `.engram-search-result-item` | `src/search-view.ts` `createDiv({cls: "engram-search-result-item..."})` | 109 |
| Ribbon aria-label: "Engram search" | `src/main.ts` `addRibbonIcon("search", "Engram search", ...)` | 298 |

**Selector fix vs plan draft:** plan used `.engram-search-result` for sidebar results — this class does not exist. Corrected to `.engram-search-result-item` (matches source). The `click_ribbon()` helper matches `aria-label` containing "Engram", which matches "Engram search" ✓.

## Test plan

- [ ] CI `build-and-test` passes
- [ ] `backend/e2e` passes (test_53 skips gracefully when plugin build lacks `open-search-sidebar`)
- [ ] Syntax check: `python3 -c "import ast; ast.parse(open('e2e/tests/test_53_search_view.py').read())"` → OK

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #846, opened 2026-07-02
fix(crdt): sync durability hardening wave (audit 2026-07-01)

Rasbandit:
Nine fixes from the 2026-07-01 CRDT sync audit, hardening the server-side durability and coexistence seams of the CRDT sync layer. Each fix landed TDD-first with a per-task spec+quality review; the branch passed a final whole-branch review (verdict: ready to merge).

## Fixes

1. **Watermark before encode** — the checkpoint prune watermark was captured *after* `encode(doc)`, so an update logged mid-encode could be pruned without being in any snapshot → permanent loss on node death. Watermark now captured first (degrades to nil = no prune on capture failure).
2. **No version/seq churn on unchanged checkpoints** — hash-equal checkpoints now compact the snapshot only; legacy `/changes` pullers no longer see phantom edits. (Also fixed a latent `{:ok, prev}` double-wrap that made the embed-skip gate always-true on main, masked by Oban dedup.)
3. **unbind runs a full checkpoint** — trailing edits in the settle window now materialize into `content`/`content_hash`/`seq` on room exit; REST, search, and legacy pullers no longer serve stale content indefinitely. `checkpoint/5` is now no-raise (function-level rescue) since it runs in terminate.
4. **Rooms trap exits + `restart: :temporary`** — ECS deploy SIGTERM now runs terminate → unbind → checkpoint instead of killing rooms unflushed. `trap_exit` is set in `bind/3` (guarded on `$initial_call` so direct test calls don't leak the flag). `:temporary` kills the immortal-orphan restart mode; channels recreate rooms on demand.
5. **Channel monitors room pids** — after a room crash, `send_yjs_message` casts silently vanished into the dead pid forever. The channel now monitors rooms, evicts dead entries on `:DOWN`, and transparently re-creates the room on the next frame.
6. **Batch upserts deliver-out + store the projection** — `POST /api/notes/batch` never pushed into live rooms, so the room's next checkpoint *reverted* batch writes. Batch rows also stored raw content while the doc seeded the normalized projection (divergent from birth). Both fixed; digest broadcast, HTTP response, and embed gate all carry the stored projection hash (no phantom syncedHashes changes, no spurious re-embeds).
7. **REST merges replay the update-log tail** — `maybe_merge_crdt` merged against the stale snapshot only, deleting live typing from the settle window. Now: three-way Yjs merge (incoming diffed against the snapshot ancestor, applied onto the tail-replayed doc) so concurrent edits survive; nil-snapshot legacy notes take a two-way merge against the tail-inclusive doc (review caught that the naive path duplicated the entire body).
8. **Frontmatter emit is total** — one non-JSON/unserializable Y.Map value used to brick a note permanently (every checkpoint + REST write raised). `emit/2` now degrades per-key (safe decode → raw value → inspect fallback), never raises.
9. **crdt_msg rate limit + frame caps + log hygiene** — 240 frames/10s per user, 5MB frame cap enforced *before* base64 decode allocates, and note paths moved out of log message bodies into redacted metadata.

Plus: fixed a pre-existing `DecryptFailureTest` interleaving flake (reproduced on main: unfiltered `Enum.find` over cross-test log events).

## Verification

- Full suite: 3327 tests, 0 failures (three full runs, seeds 0/1/7)
- `mix credo --strict` clean, `mix sobelow` clean, format clean
- Every fix has TDD red/green evidence in its commit; review loops caught 1 Critical (nil-snapshot body duplication) before merge

## Deferred (follow-up issues being filed)

Checkpoint stale-version guard (checkpoint-vs-REST LWW window); SyncChannel WS rate limiting; transport max_frame_size; rate-limit denial telemetry; CrdtDoc shutdown grace; bind/3 fail-open on snapshot decrypt failure; frontmatter order-array concurrent-rewrite duplication; audit-deferred set (TOCTOU text/state pair, flatten lineage adoption, netsplit).


Rasbandit:
## e2e-clerk regressions: root-caused + fixed

The reproducible e2e-clerk failures (test_49 rapid_api_edits x2, test_78 hash_only_live_update, test_10 rename) were **content corruption from double-encoded Yjs ops**, confirmed via a temporary diagnostics commit + targeted `[e2e: clerk/...]` run (artifacts show the exact cascade):

```
MERGE  snapshot="…Version 2"  tail_doc="…Version 22"   <- tail replay doubles the changed span
DELIVER room_before="…Version 2" content="…Version 23" <- corrupted store pushed back into a healthy room
```

Every REST write was encoded twice: once by `maybe_merge_crdt` (merge lineage, into the snapshot) and again by `deliver_out`'s `ingest_plaintext` re-diff (room lineage, into the update-log tail via `update_v1`). Fix 7's tail replay then unioned both encodings — Yjs can't dedup semantically identical ops from different lineages — so each write doubled the changed span ("Iteration 67", "Version 10" in the vault artifacts). Pre-PR the same defect only poisoned the room doc (checkpoint clobbers = flake #547); the tail replay made it deterministic in stored content.

**Fix (`2d722540`):** `push_to_live_room` now loads the just-committed CRDT snapshot and `Yex.apply_update`s it onto the room — same lineage as the snapshot, so the room's tail row becomes a subset of stored state and tail replay is idempotent. Plaintext ingest remains only as a degraded fallback when state can't be loaded. TDD'd with a new channel test that reproduces the doubling deterministically (the trick: `SharedDoc.update_doc` is a cast — synchronize on room convergence + tail row before the next write). Full suite 3328/0 locally.

Also on the branch: `docs/context/crdt-lineage-doubling.md` (invariant + debugging recipe), diag commit + revert (squash-merge flattens them).

**Notes for merge:**
- mix.exs is 0.5.605 vs main 0.5.606 — version-check passes (differs), but main has moved (PG idempotency/seq work touching sync) so a rebase + bump before merge is advisable.
- Remaining single-encoder hazard is client-side (plugin `seedOnce` enroll race can inject foreign-lineage full-text ops) — belongs to the plugin "flatten lineage adoption" deferred item, not this PR.

Rasbandit:
## Update after the deliver-out fix: 1 of 2 corruption classes fixed; the remaining one is the deferred lineage-adoption item

Latest run (28593070246): **test_49 rapid_api_edits both PASS** — the deliver-out state-apply fix holds. Remaining failures: `test_10_rename_propagation`, `test_34_folder_rename_old_paths_cleaned`, `test_78_hash_only_live_update` (now empty-file, not mangled).

Evidence from the artifacts:

1. **Lineage fork at note birth.** `RenameNew.md` in vault w1-b contains the body **twice, concatenated**. When a client authors a file on disk, the plugin pushes the same content through BOTH lanes: `seedOnce` inserts it into its ydoc (plugin lineage -> room via `sendUpdateRaw`) AND the legacy batch push (server re-encodes on the merge lineage in `maybe_merge_crdt`). Two encodings of the same text -> Yjs union doubles. This fork predates this PR (it is flake #547's substrate).

2. **This PR materializes the fork into durable state.** Timeline for `RenameOld.md`: DELETE 200 at 13:20:07.9, GET 404 at :08.4, then a plugin `batch_upsert` at :08.7 resurrects it and GETs 200 forever. The plugin echo-pushes because its `syncedHashes` no longer match: fix 3's unbind checkpoint (and the checkpoint-carried hashes of fix 2/6) write the forked room text/hash into `content`/`content_hash`, so the plugin sees a phantom server change and re-pushes stale paths. On main the fork stays in-memory (rooms rarely checkpoint mid-test), which is why main+same-plugin passes.

So the remaining red is not a bug in fixes 2/3/6 per se — they do their job — it is the **pre-existing client-side lineage fork** becoming load-bearing once room state flows into the row. That is exactly the deferred "flatten lineage adoption" item in this PR's description.

**Options** (needs a scope call):
- **A. Fix the fork at the source (plugin):** `seedOnce` must not seed while the server has state — adopt server state from the first STEP2, seed only on a server-empty handshake (extends #150's bootstrap semantics). Kills the fork for authored-file + pulled-file cases; server keeps this PR as-is. My recommendation.
- **B. Scope down this PR:** gate fix 3's unbind checkpoint (and hash-carrying broadcasts) behind the lineage work, merge the rest. Loses the durability win the PR exists for.
- **C. Server-side heuristics** (detect doubled text at checkpoint): brittle, not recommended.

Context doc with the full invariant + debugging recipe: `docs/context/crdt-lineage-doubling.md` (on this branch).

Rasbandit:
**Paired run GREEN across the board** — run 28615799090 (this branch + plugin `fix/crdt-sync-hardening` = PR Engram-obsidian#161's adopt-first seed gate): e2e-clerk, e2e-crdt, e2e-api, e2e-local, e2e-browser, unit-tests all pass. test_10/test_34/test_78 cleared.

Merge order: **merge plugin #161 first** (then this PR's e2e pairs against plugin main, which will carry the gate). Before merging this PR: rebase over main (seq/idempotency work landed) + bump mix.exs (main is at 0.5.606). The TEMP diag commit + its revert squash away.

Issue #902, opened 2026-07-04
CRDT: close the REST→live-room convergence gap (checkpoint can revert a committed write)


Rasbandit:
Re-scoped after a full audit of the CRDT persistence seam. The original framing ("seed Y.Doc from REST on room open") was wrong: `seed_from_content` (`crdt_persistence.ex:65-67`) **already exists** and works — it's only reached by migrated/legacy rows. The real divergence is elsewhere.

## What's actually solid (audit findings)

- REST/MCP upserts **always** merge incoming content into `crdt_state` and write content + snapshot **atomically in one changeset** (`notes.ex:743-753`). No content-without-snapshot on live writes.
- `bind/3` runs **synchronously in the room GenServer `init`** (`y_ex/shared_doc.ex:122-127`); STEP1 queues behind it. **No empty-STEP2 bind race.**
- Rename keeps the same `note_id` + snapshot (`notes.ex:1064-1097`); no content loss.
- All writes (REST single/batch, MCP write/append/patch/update_section) funnel through `Notes.upsert_note` → three-way convergent Yjs merge (`merge_plaintext_relative_to_snapshot`).

## The real bug: the `deliver_out` post-commit gap

REST commit and live-room convergence are **not atomic**:
1. `upsert_note` commits `notes.content` + `crdt_state`.
2. *Post-commit*, `deliver_out` (`crdt_deliver.ex`, called at `notes.ex:3309` / batch `1888-1897`) pushes the snapshot into the **live in-memory room** — best-effort.
3. If the room's **debounced checkpoint** (5s idle / 60s ceiling) or an **`unbind` on last-observer-disconnect** fires *in that gap*, it encodes the **stale pre-commit doc** and `Repo.update!`s content + snapshot — **reverting the committed write** (`crdt_checkpoint.ex:101-127`).

The code already flags this: `notes.ex:1887` — *"without this, a room that has the note open never sees the batch merge and its next checkpoint REVERTS it."* `deliver_out` is the mitigation but isn't airtight (skips on `load_merged_state` decrypt failure, `crdt_deliver.ex:96-99`; loses to writer-process death post-commit).

Secondary: a live room in that gap answers a new peer's STEP1 with **stale (pre-commit) content**.
Latent: `bind/3` has **no staleness check** — once a snapshot exists, `seed_from_content` is skipped forever even if `notes.content` later diverges (`crdt_persistence.ex:65-67`, no `content_hash` comparison).

[...]

PR #904, opened 2026-07-04
fix(crdt): checkpoint staleness guard — stop reverting REST writes in the deliver_out gap

Rasbandit:
Tier 1 of #902. Fixes the primary REST↔CRDT divergence: a checkpoint firing in the `deliver_out` post-commit window could **revert a just-committed REST/MCP write**.

## Root cause

`Notes.upsert_note` commits `notes.content` + `crdt_state`, then `CrdtDeliver.deliver_out` pushes the snapshot into any live in-memory room — **post-commit and best-effort**. If a debounced checkpoint (5s idle / 60s ceiling) or an `unbind` on last-observer-disconnect fired in that gap, `CrdtCheckpoint.checkpoint` encoded the room's **stale** doc and `Repo.update!`'d `content` — overwriting the committed write with old text. The code already flagged the hazard (`notes.ex:1887`).

Affects API/MCP writes too — any REST-path write racing a live-room checkpoint.

## Fix

Before encoding, fold the DB's current `crdt_state` into the room doc via `Yex.apply_update` — a **native CRDT merge**: convergent, artifact-free, and idempotent (a no-op when the room already ingested it via `deliver_out`). This is the same mechanism `CrdtDeliver.push_to_live_room` uses, so it's proven, not novel merge logic. A stale room can no longer clobber a commit; a concurrent room edit + REST write **converge** instead of one dropping the other.

Cost: one extra `Repo.get` + decrypt per checkpoint. Checkpoints are debounced (~1 per dirty streak, not per-keystroke), so this is negligible; Tier 2 removes it by routing REST writes through the live room (single writer).

## Tests

- `checkpoint does NOT revert a REST write that landed while the room was stale` — reproduces the revert (RED without the fix: `left: "before"`).
- `checkpoint converges a stale room's own edit with a concurrent REST write` — both edits survive.
- Full `test/engram/notes/` scope green (221 → 223 tests), credo clean, format clean.

## Deferred

The bind/3 staleness re-seed (audit Q6) — latent defense-in-depth for a divergence that current atomic-write invariants prevent. Noted on #902, not included here to keep this scoped to the active bug.

Context: client-side guards in Engram-obsidian #169/#173/#174 kept e2e green by bypassing the fragile handshake; this closes the server gap they were working around.

### DIFF
--- a/lib/engram/notes/crdt_checkpoint.ex
+++ b/lib/engram/notes/crdt_checkpoint.ex
@@ -62,6 +62,26 @@ defmodule Engram.Notes.CrdtCheckpoint do
           end
       end
 
+    # Deliver_out gap guard: a REST/MCP write may have committed a newer
+    # crdt_state that this live room never ingested (CrdtDeliver.deliver_out is
+    # best-effort and runs post-commit, so a checkpoint firing in that window
+    # would otherwise encode the stale in-memory doc and REVERT the committed
+    # write — see Engram#902). Fold the DB's current crdt_state into the doc
+    # before encoding. apply_update is convergent + idempotent (a no-op when the
+    # room already has it), the same native-CRDT merge CrdtDeliver.push_to_live_room
+    # uses — so a stale room can never clobber a committed write.
+    _ =
+      case Repo.with_tenant(user_id, fn -> Repo.get(Note, note_id) end) do
+        {:ok, %Note{} = db_note} ->
+          case Crypto.decrypt_crdt_state(db_note, user) do
+            {:ok, db_state} when is_binary(db_state) -> Yex.apply_update(doc, db_state)
+            _ -> :ok
+          end
+
+        _ ->
+          :ok
+      end
+
     text = CrdtBridge.text_of(doc)
 
     with {:ok, raw_state} <- encode(doc),
--- a/mix.exs
+++ b/mix.exs
@@ -4,7 +4,7 @@ defmodule Engram.MixProject do
   def project do
     [
       app: :engram,
-      version: "0.5.618",
+      version: "0.5.619",
       elixir: "~> 1.15",
       elixirc_paths: elixirc_paths(Mix.env()),
       start_permanent: Mix.env() == :prod,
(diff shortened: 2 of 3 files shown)


Rasbandit:
## Code review — not mergeable as-is (converted to draft)

An independent review found four real issues; the same-lineage happy path is correct and tested, but the guard fights the invariants `CrdtDeliver` protects and the "can never clobber" claim is overstated.

1. **Silent revert on decrypt failure (critical).** `case Crypto.decrypt_crdt_state(...) do {:ok, bin} -> apply; _ -> :ok end` swallows `{:error, _}` and proceeds to encode the stale doc — reverting the write anyway, with no log. `CrdtDeliver.load_merged_state` treats this error as first-class (logs `:error`, skips). Must log + abort, not fall through to a revert.
2. **Window narrowed, not closed.** The `crdt_state` fold runs in its own `Repo.with_tenant`, separate from the checkpoint write txn. A write committing in the read→write gap is still lost. Real closure needs fold + `prev`-read + write in ONE transaction.
3. **Foreign-process mutation of the live room doc.** `checkpoint` runs in the timer process; `Yex.apply_update(doc, …)` mutates the room's shared doc directly instead of routing through `SharedDoc.update_doc(room, fun)` (single-owner invariant). In the stale case it fires the doc monitor → broadcasts the delta + appends to the durable update-log on every stale checkpoint.
4. **Flatten-lineage double-body, now on idle checkpoints.** If a prior checkpoint flattened (new lineage persisted) while the live room stayed on the old lineage, `apply_update(old_doc, flattened_bytes)` unions two histories → doubled content, now reachable on a write-free checkpoint. Low-probability (needs ≥500KB + ≥1000 client-ids) but widened by this change.

**Conclusion:** #3/#4 show this patch fights the exact invariants `CrdtDeliver` was written to protect. Recommend pivoting to **Tier 2** (route REST/MCP writes through the live room owner so the room is never stale) — it eliminates the whole class by construction. Parking this draft pending that decision.

Rasbandit:
Superseded by #907 (Tier 2: version-fenced checkpoint write). This Tier 1 draft's re-read-and-fold approach had four review problems that #907 avoids by capturing atomically + refusing to overwrite newer state.

Rasbandit:
Superseded by #907 (Tier-2 fix shipped).

PR #921, opened 2026-07-06
feat(obs): attribute socket auth rejections + per-user trace attributes

Rasbandit:
## Why

The plugin bugfix (companion PR below) stops the chronic WebSocket `signature_error` bursts at their source. This PR is the observability half: make any **residual** `signature_error` attributable, and make prod traces filterable per user (the original support ask "follow traces for my account" was impossible because spans carried only HTTP-route attributes).

## Changes

**1. Attribute socket auth rejections** (`lib/engram_web/user_socket.ex` + new `Engram.Auth.TokenDebug`)

The `{:error, reason}` branch logged `auth rejected` with no attribution (rejection happens before the user is identified). It now also logs the presented token's `kid`, `iss`, `alg`, and `sub_hash`. `TokenDebug.metadata/1` is a pure, **non-verifying** peek (`Joken.peek_header/1` / `peek_claims/1`), **never raises** (guards + `is_map` + rescue cover crafted/non-JWT/non-string-`sub` input), and **never logs the raw sub**. `sub_hash` uses the keyed `Engram.Crypto.HMAC.hash_user_id/1` — the same helper used for the `user_id` in existing log metadata — so a rejected internal-JWT `sub` (which equals `users.id`) correlates to that user's other log lines.

**2. Per-user trace attributes** (new `EngramWeb.Plugs.TraceUserAttrs`)

Stamps `app.user_id` (same keyed HMAC, so trace and log `user_id` correlate) and `app.vault_id` onto the request span, added after `Auth`/`VaultPlug` in the authenticated pipelines. No-op when assigns are absent.

## Still open (flagged, not built)

Why a freshly minted internal HS256 token can fail signature verify at all (secret drift across nodes/deploys). Deferred until this attribution data exists to chase it.

## Companion

Plugin bugfix: engram-app/Engram-obsidian — branch `fix/ws-close-token-invalidation`.

## Testing

TDD. `TokenDebug`: HMAC correlation, all-nil non-JWT, no-raise on numeric `sub`. `TraceUserAttrs.attrs_for/1`: HMAC user_id + struct-extracted vault_id, empty when absent. Full `mix test` 3472 / 0 (one confirmed pre-existing `SyncChannel` flake). `mix format`, `credo --strict`, and `sobelow --exit low` all clean.

### DIFF
--- a/lib/engram/auth/token_debug.ex
+++ b/lib/engram/auth/token_debug.ex
@@ -2,15 +2,13 @@ defmodule Engram.Auth.TokenDebug do
   @moduledoc """
   Best-effort, NON-verifying peek at a token's header + claims, for attributing
   auth rejections in logs. Never raises, never verifies a signature, never logs
-  raw sub (hashed with sha256-hex).
+  raw sub.
 
-  The sha256 hash here is intentionally NOT the keyed `Engram.Crypto.HMAC.hash_user_id/1`
-  used elsewhere for our internal `users.id`: this module hashes the JWT `sub`
-  claim from an as-yet-unverified, possibly-rejected token, which may be an
-  external issuer's subject id (Clerk, device flow, etc), not our own user id.
-  Reusing the keyed helper would conflate two different id spaces under one
-  key. Plain sha256-hex still gives stable, joinable, non-reversible labels
-  for correlating repeated rejections from the same presented token.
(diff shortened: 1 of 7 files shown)


Rasbandit:
Companion plugin bugfix PR: engram-app/Engram-obsidian#178

Issue #193, opened 2026-05-21
flake: e2e-browser Clerk test user provisioning sporadic failure


Rasbandit:
**Tests affected:**
- `frontend/e2e/clerk-auth.spec.ts:59 sign in via Clerk → dashboard`
- `frontend/e2e/clerk-auth.spec.ts:63 Clerk UserButton renders in header`
- `frontend/e2e/clerk-auth.spec.ts:69 sign out via Clerk → redirects`

**Symptom:** `Error: Failed to sign in with email e2e-browser-XXXXXXXXXX@test.com: No user found with email: ...`

**First seen:** 2026-05-21 during pricing-v2 §B PR review (run 26216389956, job 77139882021). Passed on rerun → flake.

**Suspected cause:** Clerk test-user provisioning isn't waiting for the user to be queryable before `clerkSignIn` runs. Race between `createTestUser` and Clerk's eventual-consistency replication.

**Suggested fix:**
- In `clerk-auth.spec.ts` `beforeEach`: short retry loop polling `clerk.users.getUser(email)` until found before calling `signIn`
- OR seed user in a global setup so propagation completes before any test runs

Filed per repo `feedback_diagnose_flakes` policy.

Rasbandit:
Addressed in #335 (merged, v0.5.240).

Added a bounded retry in `clerkSignIn` (`frontend/e2e/clerk-auth.spec.ts`) that catches `No user found` and retries with a short delay before failing — handles Clerk's eventually-consistent user creation, where the backend sign-in-token call can briefly 404 the user created in global-setup. Any other error rethrows immediately. Leaving open to verify over the next several runs.

Rasbandit:
Hit again 2026-06-01 — same Clerk test-user provisioning race, this time in `e2e-clerk` (not `e2e-browser`):

- `tests/test_49_cross_auth_sync.py::test_apikey_push_oauth_receives`
- `tests/test_49_cross_auth_sync.py::test_oauth_push_apikey_receives`

Symptom:
```
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://api.clerk.dev/v1/sessions
```

Run: https://github.com/engram-app/Engram/actions/runs/26739862518 (squashed into 4a279165 via #383 after rerun-pass; second CI run was clean).

[...]

PR #947, opened 2026-07-07
fix(crdt): separate rate-limit bucket for sync handshakes (STEP1/STEP2)

Rasbandit:
## The incident trigger (2026-07-07 cross-file overwrite)

Connect-time CRDT enrollment fires one STEP1 per note. On a ~230-note vault that blew the single `240/10s` `crdt_msg` budget on every connect — the victim device got **40× `rate_limited` replies in one second** (20:20:25Z, Loki `[client:channel]` lines), starving its real edit frames and degrading the noteIdMap into the wrong-mint state that became cross-file data loss (see engram-app/Engram-obsidian#193 for the plugin side + full forensics). This gap was already flagged as a hardening follow-up in `docs/context/noteidmap-stale-breaks-sync.md`; the incident promoted it to causal trigger.

Aggravator on the released code (v0.5.636): the bucket was keyed per-user only, so Obsidian + webapp **shared one budget** — main's per-device split helps, but enrollment could still starve a single device's own edits.

## Fix

Classify each frame from its **first 4 base64 chars** (= first 3 bytes; yjs v1 subtypes are all single-byte varints, per `Yex.Sync`):
- `<<0,0,…>>` STEP1 → handshake lane, unconditional (non-mutating)
- `<<0,1,…>>` STEP2 → handshake lane **only if ≤4KB b64** (enrollment's near-empty echo replies); larger STEP2s are state-bearing mutations and pay the edit budget — y_ex routes STEP2 and update into the same `apply_update`, so an unqualified classification would let a client relabel edits and mutate at 10× the cap
- everything else → edit lane, unchanged `240/10s`

Handshake lane: `2400/10s` per device, same account-multiplier ceiling. **Bounded, not exempt.** Classification runs BEFORE the rate check, so over-budget floods are still rejected at ETS-lookup cost, never MB-scale decode cost; the full base64 decode happens only after the limiter allows the frame.

## Known ceiling

A >2400-note vault spills enrollment into a second window; client-side enrollment pacing is the follow-up if that ever bites.

## Verification

- 23 channel tests green, incl. new: handshake frames don't consume the edit budget; handshake lane still bounded; **large STEP2 pays the edit budget** (relabeling test)
- Independent adversarial review: protocol classification confirmed sound against `y_ex` (no varint multi-byte break, no message smuggling — `process_message_v1` decodes exactly one message per frame); both review findings (STEP2 amplification, decode-before-check flood cost) fixed in `1c8a97b5`
- format + credo --strict + sobelow clean; full suite green (1 unrelated red: `RlsCoverageTest` sees the in-flight `issue_reports` table from a sibling worktree's migration in the shared test DB — that feature branch needs its RLS policy)
- v0.5.640

## Companion

Plugin PR: engram-app/Engram-obsidian#193 (map bijection + reconcile correction + destructive-op guard). Both should release together — this PR removes the degraded windows; #193 makes a degraded map unable to destroy data.

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -194,10 +194,26 @@ report "sobelow" sobelow || FAILED=1
 # ── Release smoke (dependency changes only) ──────────────────────────────
 # `mix release` assembly runs a duplicate-module check that `mix test` never
 # does. Two deps vendoring the same module names (for example hackney's h2 and
-# grpcbox's ts_chatterbox) only blow up here. Run it when mix.exs/mix.lock
-# changed vs origin/main, so normal pushes stay fast.
-if git diff --name-only origin/main...HEAD 2>/dev/null | grep -qE '^mix\.(exs|lock)$'; then
-  echo "▸ Release smoke: mix.exs/mix.lock changed, running MIX_ENV=prod mix release --overwrite"
+# grpcbox's ts_chatterbox) only blow up here.
+#
+# Trigger ONLY on changes that can actually break release assembly: mix.lock,
+# rel/, or mix.exs lines OTHER than the version bump. The version policy above
+# forces a mix.exs bump on every deployable branch, so triggering on any
+# mix.exs change made the smoke run on essentially every PR's first push —
+# and in a fresh worktree (_build/prod is cold by design, see post-checkout)
+# that is a ~5 min full prod compile for a change that cannot introduce
+# duplicate modules. A bare version bump is exempt; real dep/release-config
+# changes still pay.
+release_smoke_needed=false
+if git diff --name-only origin/main...HEAD -- mix.lock rel 2>/dev/null | grep -q .; then
+  release_smoke_needed=true
+elif git diff origin/main...HEAD -- mix.exs 2>/dev/null \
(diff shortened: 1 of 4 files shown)


Rasbandit:
Added `16e9327f` — pre-push hook fix (bundled per Todd): the `MIX_ENV=prod mix release` smoke gate triggered on ANY mix.exs change, but the version policy forces a mix.exs bump on every deployable branch, so every PR's first push from a fresh worktree paid a ~5 min cold prod compile for a version-string change that cannot introduce duplicate modules. Now it triggers only on `mix.lock`, `rel/`, or non-version mix.exs hunks.

Measured on this very branch: **5 min → 21 s** for the same push. Real dep/release-config changes still pay the gate.

Rasbandit:
e2e-clerk red on the hook-fix push = the known `test_48_oauth_reconnect_catchup` flake class ('WebSocket channel not connected after 30s'), NOT this diff:
- the previous run of this branch (23:42Z) with **identical runtime code** passed e2e green — the failing run's only delta is `.githooks/pre-push`, which never ships
- main itself failed e2e at 21:58Z and 18:33Z today
- mechanism: this diff touches only `crdt_msg` rate classification; the join path has no limiter and the handshake lane is strictly more permissive

Rerunning the failed job. test_48 remains an open flake class (tracked since the 2026-07-04 forensics).

Issue #949, opened 2026-07-08
e2e: SPA write-direction browser leg (CodeMirror edit → Obsidian) + Clerk-capable browser sign-in


Rasbandit:
## Follow-up: SPA write-direction browser leg + Clerk-capable browser sign-in

This is a follow-up to the Obsidian→web live-render leg added in the test-confidence wave (`e2e/tests/crdt/test_obsidian_to_web_live.py`), which covers Obsidian edit → web SPA render only. The reverse direction — editing in the web SPA's CodeMirror editor and asserting the change propagates back to Obsidian — is not covered.

Two things are needed, and both are their own project:

1. **CodeMirror write-direction leg.** Driving CodeMirror 6 via Playwright (focus, keystroke input, save-trigger, waiting for the CRDT broadcast to land) is materially more involved than the read-only render assertions in the existing live leg. Deserves its own test file and its own driver helpers, not a bolt-on.

2. **Clerk-capable browser sign-in helper.** The current browser sign-in helper is local-auth only, which is why all browser legs (including the existing Obsidian→web leg) run only in the `e2e-crdt` job and not in the main `e2e-clerk` suite. Building a Clerk-capable sign-in helper (real or test-mode Clerk flow) would let browser legs run in the main e2e-clerk suite instead of being confined to the CRDT-only job.

Deliberately out of scope for the test-confidence wave (companion PR).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #948, opened 2026-07-08
test: 2-node PubSub/idempotency coverage


Rasbandit:
## Pattern A: single-node topology blindness

The 2026-07-07 cross-repo testing-surface audit (`docs/context/testing-surface-audit-2026-07-07.md` in engram-workspace) identified that PubSub fan-out across nodes, node-local idempotency caches, and cross-node cache eviction are structurally untested. The unit-test suite runs single-node (no BEAM distribution — `:net_kernel.start` fails with `:nodistribution` on CI's unit-tests runner), so anything that only breaks under real clustering has no automated coverage.

Three real bugs in this class were found by manual audit, never by a test:
- #710 — prod ran two unclustered BEAM nodes for weeks before anyone noticed (Dockerfile dropped `rel/`, fixed by #717)
- #862 — sync correctness issue under clustering (idempotency keys, `seq NOT NULL`)
- #866 — same audit wave, `pull_changes` removal

A 2-node harness is its own design project (peer nodes, PubSub fan-out assertions, distributed-ets rate-limit sync, cache-eviction propagation) and is deliberately out of scope for the test-confidence wave (this PR). This issue tracks scoping and building that harness.

Refs the workspace audit doc: `docs/context/testing-surface-audit-2026-07-07.md`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #945, opened 2026-07-07
test: pin why a pre-#187 plugin passes test_82 — /sync/changes may re-teach forgotten note ids


Rasbandit:
During the testing-confidence wave (Task A2, branch `test/confidence-wave`, commit b7b2c10c), the new resumed-device regression test `test_82_resumed_device_stale_idmap` was bite-checked against a pre-#187 plugin build (plugin 7e73d95, parent of the reconcile fix; `grep -c reconcileNoteIdMapFromManifest main.js` = 0). Expected FAIL; got PASS.

Verified facts (full trace in the workspace `.superpowers/sdd/task-A2-report.md`):
- The fixture provably constructs the resumed state: `data.json` shows `noteIds={}` with `syncCursor` preserved, and the restart path (`async_start(restart=True)`) does not re-prepare the vault.
- On the pre-#187 build, the in-memory noteIdMap after restart is repopulated with ids that EXACTLY match the server manifest — on a build with no manifest reconcile at all. A 5-note churn phase (cursor advanced well past the seed notes) did not change this.
- Push-side masking is explained: `clearConfirmedNoteIds()` on every reconnect structurally routes immediately-post-restart pushes via legacy REST (path-keyed). Pull-side re-teaching is NOT explained.

Hypothesis to check: the current backend's `/sync/changes` response for `since=<already-caught-up cursor>` is more generous than delta-only (or an adjacent mechanism re-sends id↔path pairs), i.e. the backend drifted since the original 2026-07 incident so the bug class may be partially self-healing now. Worth pinning because test_82's value as a regression net depends on knowing what actually guards this state.

Refs plugin #187 (the reconcile fix), workspace doc `noteidmap-stale-breaks-sync.md`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Rasbandit:
[...]

Issue #944, opened 2026-07-07
e2e live-window: test_41 canvas_modify_sync fails without pull-assist — rapid double-write never delivers the second write


Rasbandit:
Found by Task A1 of the testing-confidence wave (branch `test/confidence-wave`; local stack 0.5.639, plugin main 77f0338, `E2E_WORKERS=1 --reruns 0`, reproduces solo).

## Symptom

`test_41_canvas_sync.py::test_canvas_modify_sync` writes a base canvas on A, waits for B to receive it live (passes), then immediately overwrites with a modified version (~2ms after the base push per A's logs) and waits for B to see the new content (`wait_for_content`, 30s, pure disk poll, no pull). B materializes the base write (`[client:pull] Created: ... len=430`) but never the modified 582-byte version.

## Hypothesis

The two rapid back-to-back upserts get coalesced somewhere in the broadcast/hash-dedup path so B only ever sees the first. Candidate suspects: the deliberate hash-equal broadcast-skip (`notes.ex` upsert path) interacting with in-flight state, or client-side echo/dedup on the receiving plugin.

## Repro

Convert-first live assert per the live-window discipline (test file on `test/confidence-wave`), run test_41 solo against a version-matched stack per `local-crdt-e2e-repro.md`.

The test is deliberately left red on the branch (no pull-masking re-added). Full diagnostics: workspace `.superpowers/sdd/task-A1-report.md`, Finding 2.

Refs Engram#942 (attachment live delivery — separate finding from the same wave).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #950, opened 2026-07-08
test: confidence wave — live-window discipline, resumed-device, cross-interface, MCP/CRDT parity, no reruns, web leg

Rasbandit:
## WARNING: 7 e2e tests on this branch are DELIBERATELY red

These tests surface real product gaps that pull-masking previously hid. Merging this PR **accepts these as known-red pending the linked issues** — re-adding pull-masking or reruns to make them green again is NOT the fix:

- `test_33_attachment_sync.py` (2 tests), `test_79_attachment_move_convergence.py` (2 tests), `test_80_web_attachment_upload.py` — attachment live delivery never actually happens. See #942 (evidence commented there).
- `test_41_canvas_sync.py` (canvas modify) — a rapid double-write loses the second update. See #944.
- `test_10_rename_propagation.py` — first-push-of-session cold-start delay. See Engram-obsidian#189 (evidence commented there).

## Context

- Workspace audit: `docs/context/testing-surface-audit-2026-07-07.md` (engram-workspace#116) — six failure-to-catch patterns across the stack, this PR + the companion plugin PR close items 1-4, 6, 7.
- Companion plugin PR: Engram-obsidian#190 (Phase B, done).

## Per-task summary

- **A1** (audit item 1 — pull-masking) — `4473d544` — assert live delivery lands within the live-sync window *before* any test falls back to a pull, so a broken live path can no longer hide behind pull-masking.
- **A2** (audit item 2 — fresh-device-only fixtures) — `b7b2c10c` — resumed-device regression test: a device that reconnects with a stale `noteIdMap` must still sync correctly, not just a freshly-provisioned one.
- **A3** (audit item 3 — cross-interface gaps) — `2bbf23d7` — orchestration test driving MCP + REST + Obsidian against the same vault in one flow, so a change from one interface is asserted visible from the others.
- **A4 + A5** (audit item 4 — untested `note_changed` fan-out) — `165875c7` (MCP write tools must emit `note_changed`) + `5e72cafa` (a CRDT-checkpointed edit must reach the `/changes` feed) — closes the gap where CRDT/MCP writes could silently skip the broadcast path.
- **(audit item 5 — plugin mock-shape asserts)** — handled by the companion plugin PR, Engram-obsidian#190. Not in this PR.
- **A6** (audit item 6 — rerun anti-diagnosis) — `fcc82a4c` — `reruns=0`. A flake is a finding, not noise to retry away.
- **A7** (audit item 7 — no browser-rendered coverage) — `acd0c26c` — new web SPA live-render leg, Obsidian edit → browser (`e2e/tests/crdt/test_obsidian_to_web_live.py`).

## Honest limitations

- **test_82's bite-check is inconclusive.** A pre-#187 plugin also passes this test, because the current backend re-teaches ids via an unpinned mechanism. #945 tracks pinning that mechanism down so the test can actually discriminate pre/post-#187 behavior. The fixture state itself (what's on disk after the flow) is verified independently and is not in question.
- **The web SPA browser leg only runs in the `e2e-crdt` job.** The browser sign-in helper is local-auth only, so it can't run in the main `e2e-clerk` suite yet. Follow-up: #949.

## Deliberately out of scope (per the plan)

- Mass rewrite of the plugin's 141 mock-shape assertions.
- 2-node topology coverage (PubSub fan-out, node-local idempotency caches, cross-node cache eviction) — tracked in #948.
- SPA write-direction leg (CodeMirror edit → Obsidian) + Clerk-capable browser sign-in — tracked in #949.

Refs #942, Refs #944, Refs #945

---

## CI status after the de-flake waves (final)

Four fix waves ran after opening (commits `ca33a2a9`, `45f2c2c3`, `95495be0`, `600a547b`). Result:

- **Fixed (harness/infra, verified green):** test_50 collection-wait budget (was being RERUN-rescued in **4/4 recent green main runs** — proof the rerun net was propping up main), test_47/48 OAuth cold-boot moved into a session warm fixture, test_65 render window, test_82 cursor-persist race.
- **Remaining red = product findings only:**
  - Deterministic (7): test_33 ×2 / test_79 ×2 / test_80 (#942 attachment live delivery), test_41 modify (#944 double-write), test_10 (Engram-obsidian#189 materialize stall).
  - Intermittent (1): test_43 multi-note burst — same `received=yes materialized=no` signature as #189 (evidence commented there). Given a 60s window; it can still exceed it because the stall is unbounded (test_78 forensics showed materialization at teardown-reconnect). It flickers until #189 is fixed.
- Main-branch note: recent green main runs also rerun-rescued test_10 and test_78 — the rescue population this PR's `reruns = 0` makes visible is not specific to this branch.

### DIFF
--- a/e2e/conftest.py
+++ b/e2e/conftest.py
@@ -328,6 +328,66 @@ def cdp_c(obsidian_c):
 
 
 # ---------------------------------------------------------------------------
+# Resumed-device fixture (dedicated instance pair, function-scoped)
+#
+# A stop/mutate-data.json/restart cycle must never touch the session-scoped
+# A/B/C fixtures the rest of the suite depends on, so this is its own pair on
+# its own ports/displays. It shares sync_user/sync_client_id with session A/B
+# so it lands on the SAME server vault api_sync polls (client_id upsert is
+# idempotent — four "devices" on one vault is exactly the multi-device shape
+# under test).
+# ---------------------------------------------------------------------------
+
+RESUMED_CDP_PORT_A = _worker_port("E2E_CDP_PORT_RESUMED_A", "9350")
+RESUMED_CDP_PORT_B = _worker_port("E2E_CDP_PORT_RESUMED_B", "9351")
+RESUMED_DISPLAY_BASE = int(os.environ.get("E2E_DISPLAY_BASE_RESUMED") or "150") - _WORKER * 2
+assert RESUMED_DISPLAY_BASE - 1 >= 1, (
+    f"RESUMED_DISPLAY_BASE={RESUMED_DISPLAY_BASE} too low for worker {_WORKER}"
+)
+
+
+@pytest.fixture
+def fresh_instance_pair(sync_user, sync_client_id):
+    """Dedicated A/B-shaped instance pair for tests that stop/restart a device.
+
+    Function-scoped so a mid-test restart can't poison the session fixtures.
+    """
+    inst_a = ObsidianInstance(
+        name="ResumedA",
+        vault_path=Path(f"{VAULT_PREFIX}-resumed-a"),
+        cdp_port=RESUMED_CDP_PORT_A,
+        display=f":{RESUMED_DISPLAY_BASE}",
+        api_url=API_URL,
+        api_key=sync_user[2],
+        plugin_src=PLUGIN_SRC,
+        obsidian_bin=OBSIDIAN_BIN,
+        client_id=sync_client_id,
+        config_dir=Path(f"{CONFIG_PREFIX}-resumed-a"),
+    )
+    inst_b = ObsidianInstance(
+        name="ResumedB",
+        vault_path=Path(f"{VAULT_PREFIX}-resumed-b"),
+        cdp_port=RESUMED_CDP_PORT_B,
+        display=f":{RESUMED_DISPLAY_BASE - 1}",
+        api_url=API_URL,
+        api_key=sync_user[2],
+        plugin_src=PLUGIN_SRC,
+        obsidian_bin=OBSIDIAN_BIN,
+        client_id=sync_client_id,
+        config_dir=Path(f"{CONFIG_PREFIX}-resumed-b"),
+    )
+    inst_a.start()
+    inst_b.start()
+    cdp_a = CdpClient(port=inst_a.cdp_port)
+    cdp_b = CdpClient(port=inst_b.cdp_port)
+    try:
+        yield inst_a, inst_b, cdp_a, cdp_b
+    finally:
+        inst_a.stop()
+        inst_b.stop()
+
+
+# ---------------------------------------------------------------------------
 # API clients (always use API key — works with any auth provider)
 # ---------------------------------------------------------------------------
 
--- a/e2e/conftest.py
+++ b/e2e/conftest.py
@@ -351,6 +351,11 @@ def _oauth_ws_warm(cdp_a, clerk_client):
         clerk_user_id, tokens = await provision_oauth_tokens(clerk_client, API_URL, label="warm")
         try:
             original = await swap_to_oauth(cdp_a, tokens)
+            # From here on cdp_a wears the throwaway OAuth identity, whose
+            # Clerk user the outer finally deletes. restore_auth must run on
+            # EVERY exit path, or the session-scoped cdp_a serves the rest of
+            # the suite as a deleted user and one diagnosable warm-up failure
+            # cascades into dozens of misleading auth errors downstream.
             try:
                 await wait_for_stream(cdp_a, timeout=120)
             except TimeoutError as e:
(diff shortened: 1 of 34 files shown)


Rasbandit:
## reruns=0 exposed rerun-masked marginal wait budgets

Two systematic CI failure classes surfaced after this branch dropped `reruns=1`. Neither is a regression from this branch's actual changes — they're pre-existing marginal budgets that reruns had been silently rescuing.

**Proof:** the latest green `main` run (28907196463) RERUN-rescued `test_50_qdrant_binary_quantization::test_collection_exists` — a "pass" that only happened because the rerun re-executed the fixture, doubling its effective wait window.

### Failure 1 (e2e-api + e2e-local): test_50 fixture timeout
`seeded_note` polled Qdrant collection creation with a 30s budget. Under CI load, the embed queue's first Oban job (Ollama cold model load + queue contention) can exceed that. Bumped to 120s, matching the already-tuned sibling `wait_for_qdrant_indexed` budget, and the timeout error now reports elapsed time. (`ca33a2a9`)

### Failure 2 (e2e-clerk): test_47/test_48 WS connect-gate timeout
`wait_for_stream`'s 30s budget (token refresh + getMe backoff + WS phx_join) was rerun-masked. Checked plugin issue Engram-obsidian#186 first — it diagnoses a *separate* post-connect delivery race (never-seen CRDT note lost between `crdt:` join and `crdt_doc_ready`), not this connect gate, so no code fix applied there. Bumped the connect budget to 60s, same load-tuned pattern as the sibling `RT_TIMEOUT` bump in #643. (`45f2c2c3`)

Both are wait-budget bumps only — no asserts weakened, no reruns re-added.

**Expected remaining red:** exactly the 7 deliberately-red finding-tests (test_33 ×2, test_79 ×2, test_80, test_41 modify, test_10), pending Engram#942/#944/Engram-obsidian#189.

Rasbandit:
## Product-fix wave: Engram#942 + Engram#944

Two real product bugs the live-delivery test conversion (Task A1) exposed. TDD RED→GREEN throughout.

### Engram#942 — attachment create/upload never broadcasts live — FIXED (backend-only)
`upsert_attachment/3` had no `Broadcast.emit` anywhere in its body (delete and move both had it). Fix: broadcast the same `"upsert"` shape `move_attachment/4` already sends for its new-path leg, after the write commits. No plugin change needed — `handleStreamEvent` already fetches+materializes any attachment `"upsert"` event (proven today by move's new-path leg).

- RED→GREEN: `test/engram/attachments_broadcast_test.exs` (new).
- Gauntlet: format/credo --strict/sobelow clean, `mix test` → 3507 tests, 0 failures.
- e2e: rebuilt local image to 0.5.641, `tests/test_33_attachment_sync.py::test_attachment_push_and_pull` (the exact create→live-deliver assertion from the issue) **PASSED**. Sibling tests in the same file/79/80 hit host-resource-contention flakes on this shared sandbox (Xvfb/CDP startup timeouts under 5+ concurrent Claude sessions), never an assertion failure — recommend re-running in CI to confirm clean.

Commit `240b3b9a`.

### Engram#944 — canvas rapid double-write loses the second update — NEEDS PLUGIN FIX (backend confirmed correct)
Checked the hash-equal broadcast-skip in `notes.ex` first — ruled out: `content_hash` is always derived from the CRDT-merged text, so two upserts with different content always broadcast. Added `test/engram/notes_broadcast_test.exs` regression case (two rapid upserts, different content → two broadcasts) — passed immediately, no backend fix needed.

Reproduced live against the rebuilt stack: `test_41_canvas_sync.py::test_canvas_modify_sync` fails exactly as described. Server-side `#908` breadcrumbs show **both** broadcasts fire (`sync broadcast emit ... op=upsert`, 1.1s apart) — backend is not the culprit.

Root cause is plugin-side (`engram-obsidian-sync/src/sync.ts`), confirmed directly in the plugin's own remote log during the repro: applying a pulled/WS file write re-triggers Obsidian's own vault event → `handleModify` → a self-echo `pushFile`. That push correctly no-ops on hash match ("Echo skip"), but its `finally` block (`sync.ts:1322-1329`) unconditionally calls `markRecentlyPushed(file.path)` even on the no-op branch, opening a 5s `recentlyPushed` suppression window. A real second broadcast arriving in that window gets dropped by `handleStreamEvent`'s echo suppression — logged directly as `Echo skip (recently pushed): E2E/TestCanvasMod41.canvas`. Fix: only call `markRecentlyPushed` on the branches that actually POST new content, not unconditionally in `finally`. Filing against `engram-obsidian-sync` with these line numbers.

Commit `8c765d68` (regression test only, permanent backend guard).

Full writeup with commands/log excerpts: workspace `.superpowers/sdd/task-A8-report.md`, "Product-fix wave" section.

Rasbandit:
## Round 5 (run 28919928915) — Defect C fix, plus A/B land plugin-side

**Defect C — test_65 problem-dir scanner (`89d4fd15`, harness precondition, NOT timing).** The 5s→20s bump attacked the wrong variable: `renderIgnoreWarnings()` scans exactly once, synchronously, at Advanced-tab render (`app.vault.getFolderByPath('node_modules')`) and never re-scans. If Obsidian's vault index hasn't ingested the externally-seeded folder when the tab opens, the warning row can NEVER appear regardless of DOM-poll duration. Fix: enforce the precondition — wait (30s cap) for the vault index to see `node_modules/` BEFORE opening settings, with an assert naming the watcher/index layer on failure. No assert weakened.

**Defect A (test_34)** and **Defect B (test_48)** were plugin root causes — see Engram-obsidian PR #196 round-5 comment (`ensureFolder` check-then-create race dropping a burst-delivered note; `setupNoteStream` short-circuit reusing a never-connected channel built mid-auth-swap). Backend commit `aeafcc7d` was checked first for test_48 and exonerated: its `_assert_owns_vault` passed and the failing interleave (settings-first swap ordering vs the plugin's #169 short-circuit) predates it.

e2e harness unit tests: 19 passed.

Issue #959, opened 2026-07-08
Checkpoint abort on unreadable crdt_state leaks tail-log unboundedly (frozen note)


Rasbandit:
Retro-review finding on merged PR #953 (Important, ~85 confidence).

The Phase-0 abort (correct: never overwrite state we cannot prove we contain) has no bounded remediation for PERMANENTLY unreadable stored state (lost DEK version, botched AAD rebind): every debounce tick appends a tail row via update_v1 and aborts → crdt_update_log grows without bound, notes.content freezes at the last good checkpoint, embeddings never refresh. Only signal is a repeating error log.

Wanted: per-note tail-log depth metric + alert, a distinct greppable key for the permanent case, and a quarantine/repair path (e.g. escalate to ops after N consecutive aborts). Transient KMS blips already self-heal.

Refs: identity-as-CRDT handoff; #953 retro-review F3.

Rasbandit:
Re-prioritized **p2 → p1** by the 2026-08-08 triage sweep, and bound into the checkpoint-seam workstream: **#1320**.

Rationale: this is one of four findings that all land on the same `Engram.Notes.CrdtCheckpoint` pipeline (`union_with_row_state` → `maybe_flatten` → persist, both write paths at crdt_checkpoint.ex:229 and :289). Each was filed separately off a retro-review of an already-merged PR, so each was triaged alone and landed at p2. Read together, three of the four silently destroy or duplicate user note content.

Fix as part of #1320 rather than in isolation — four independent patches to one function is how you get a fifth finding.

Issue #958, opened 2026-07-08
Checkpoint union can double content after a flatten (disjoint lineage)


Rasbandit:
Retro-review finding on merged PR #953 (Important, ~82 confidence, rare trigger).

`maybe_flatten` deliberately breaks Yjs lineage (a reset, not a merge) and persists the fresh-lineage state to the row — but the LIVE room keeps running its old lineage. The Phase-0 union (`union_with_row_state`: row state + live doc state) then merges two disjoint lineages at the next checkpoint: every old-lineage struct integrates as NEW content → union text ≈ T + T' (doubled), materialized into content/content_hash/tags AND re-encoded as the new crdt_state.

Trigger requires both flatten ceilings (≥500KB AND ≥1000 client-ids) plus continued live edits after the flatten — rare, but the population is exactly long-lived heavily-collaborated notes. Pre-#953 this case was a harmless wasted flatten; the union converts it into persisted corruption.

Fix directions: adopt the flattened doc back into the live room at flatten time (single lineage restored), or detect disjoint lineage (state-vector overlap) and skip the union for that checkpoint.

Refs: identity-as-CRDT handoff; #953 retro-review F1.

Rasbandit:
Re-prioritized **p2 → p1** by the 2026-08-08 triage sweep, and bound into the checkpoint-seam workstream: **#1320**.

Rationale: this is one of four findings that all land on the same `Engram.Notes.CrdtCheckpoint` pipeline (`union_with_row_state` → `maybe_flatten` → persist, both write paths at crdt_checkpoint.ex:229 and :289). Each was filed separately off a retro-review of an already-merged PR, so each was triaged alone and landed at p2. Read together, three of the four silently destroy or duplicate user note content.

Fix as part of #1320 rather than in isolation — four independent patches to one function is how you get a fifth finding.

Rasbandit:
## Confirmed. It reproduces.

This was filed at *"Important, ~82 confidence, rare trigger"*. It is real, and the trigger does not need the flatten ceilings to demonstrate. Reproduced against current `main`:

```
[1] after normal checkpoint:        "ALPHA"
[2] row flattened to fresh lineage; live doc still on old lineage
[3] after post-flatten checkpoint:  "ALPHA BETAALPHA"   <- expected "ALPHA BETA"
```

[...]

PR #960, opened 2026-07-08
fix(crdt): rooms must not outlive their vault + review hardenings

### LINKED ISSUE #954: Vault deletion leaves orphaned CRDT rooms; checkpoint raises on deleted users/notes
Observed 2026-07-07 (Loki checkpoint error storms, e.g. 19:03 local: dozens of `crdt checkpoint raised … Ecto.NoResultsError` for `019ef66f-*` ids from deleted vaults).

Two parts:
1. Deleting a vault does not tear down its live CRDT rooms — they linger, and their timers/unbinds keep firing checkpoints against rows that no longer exist.
2. `crdt_checkpoint.ex` uses `Accounts.get_user!` / `Repo.get!` — a deleted user/note turns a benign no-op into caught-but-noisy error storms (Sentry + Loki pollution).

Fix: vault deletion walks CrdtRegistry and stops its rooms (brutal kill is fine — quarantine precedent in CrdtDeliver); checkpoint switches to non-raising gets with nil → no-op return.

Refs: identity-as-CRDT handoff (engram-workspace docs/superpowers/plans/2026-07-08-identity-as-crdt-phases.md, A2).

Rasbandit:
Handoff item A2 (closes #954) bundled with the two mechanical fixes from the #953 retro-review (same initiative: room lifecycle correctness).

1. **Vault deletion tears down rooms** — post-commit (a rolled-back delete must not kill rooms), via new `CrdtRegistry.terminate_room/1`. Orphaned rooms ticking against purged rows caused the 2026-07-07 checkpoint error storms.
2. **Checkpoint: deleted user/note = quiet skip** — `get_user`/`Repo.get` with :warning-level skips replace the caught-raise error storms (Sentry/Loki noise class).
3. **Review F2**: the deliver quarantine's `:global.unregister_name` received the `{:global, …}`-wrapped name — silently a no-op, defeating the lookup-race hardening. `terminate_room` unregisters the INNER `{:crdt_doc, id}` term; test pins that the name is gone synchronously, before :global's async DOWN cleanup.
4. **Review F4**: the apply-failure signal is ref-tagged per deliver call — a stale signal left by a timed-out `update_doc` can no longer be consumed by a later deliver and false-quarantine a healthy room (regression test plants the legacy-shaped stray message).

Remaining #953-review findings filed as design issues: #958 (flatten-lineage union doubling), #959 (tail-log leak on permanently-unreadable state).

TDD throughout (5 new tests, RED first where the behavior existed to flip). Full suite 3510/0, credo --strict clean. v0.5.644 (skips 643 held by open #957).

### DIFF
--- a/lib/engram/notes.ex
+++ b/lib/engram/notes.ex
@@ -635,6 +635,22 @@ defmodule Engram.Notes do
 
     Enum.each(ids, &Engram.Notes.CrdtRegistry.terminate_room/1)
     :ok
+  rescue
+    # Runs in delete_vault's post-commit tap: a raise here would surface as a
+    # failure of an already-COMMITTED delete and skip the GateCache eviction
+    # that follows. Cleanup never fails the write; orphaned rooms are killed
+    # by deliver-time quarantine anyway.
+    e ->
+      Logger.warning(
+        "crdt vault room teardown failed",
+        Metadata.with_category(:warning, :sync,
+          user_id: user_id,
+          vault_id: vault_id,
+          error: Exception.message(e)
+        )
+      )
+
+      :ok
   end
 
   # Id-keyed rename support (Phase I): the plugin renames a note by keeping
--- a/lib/engram/notes.ex
+++ b/lib/engram/notes.ex
@@ -618,6 +618,25 @@ defmodule Engram.Notes do
     end
   end
 
+  @doc """
+  Kill every live CRDT room belonging to `vault_id`'s notes (#954). A room
+  must not outlive its vault: orphaned rooms kept ticking checkpoints against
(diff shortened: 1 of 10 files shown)


Rasbandit:
Pre-merge review complete. Finding addressed in 475a1e01: kill_live_rooms_for_vault now rescues (post-commit cleanup must not fail a committed delete or skip GateCache eviction). Review also confirmed: terminate_room TOCTOU windows are design-tolerable (rooms are row-rebuildable), tenant scoping correct, checkpoint quiet-skip correctly distinguishes deleted-vs-transient, F4 ref-tag sound (update_doc is a sync GenServer.call).

PR #962, opened 2026-07-08
test(e2e): encode the 2026-07-07 incident chain (create-race + missed delivery)

Rasbandit:
## What

Identity-as-CRDT handoff A5 — two live tests encoding the production incident chain, numbered test_84/test_85 (#950 took 82/83):

- **test_84_create_race**: an API writer and plugin A race a create on ONE path. Pins the #197 regression surface — the racing device must adopt the winning id and its RECEIVE path must stay alive (a follow-up API edit arrives on both devices, no restart). Deliberately does NOT pin which racer's content wins: a no-base REST create is by contract a full-content overwrite.
- **test_85_missed_delivery_no_deletion**: B misses a server edit (channel down), then pushes its own edit — the killer push. THE invariant: neither edit silently vanishes (conflict copy is the acceptable surface), and B converges to the server's main-file content after reconnect.

## The tests already earned their keep

test_85 failed against v0.5.643 + plugin v1.11.27 on its first run and exposed a live data-loss hole: the plugin's C1 WS branch recorded no serverHash, so a deaf device's REST-fallback push carried no base_hash, bypassed the CAS gate, and silently erased the server edit (zero note_stale_base_rejected). Fixed in plugin #207 (v1.11.32, merged) — with it, the same run logs exactly one note_stale_base_rejected and both edits survive. Validation of that discriminator on a fresh stack: test_85 fails against plugin main-without-#207 and passes with it.

Both tests are rerun-safe via unique uuid paths. Validated green locally against the CRDT stack (backend 0.5.644 image + plugin with #207/#209).

v0.5.646.

### DIFF
--- a/e2e/conftest.py
+++ b/e2e/conftest.py
@@ -16,6 +16,7 @@ files triggers the plugin's file watcher, causing unexpected sync events.
 
 from __future__ import annotations
 
+import asyncio
 import logging
 import os
 import secrets
@@ -30,6 +31,7 @@ from helpers.billing import grant_test_plan
 from helpers.cdp import CdpClient
 from helpers.cleanup import cleanup_minio_bucket, cleanup_test_data, cleanup_vaults
 from helpers.obsidian import ObsidianInstance
+from helpers.oauth import provision_oauth_tokens, swap_to_oauth, restore_auth, wait_for_stream
 
 
 logging.basicConfig(level=logging.INFO, format="%(levelname)s %(name)s: %(message)s")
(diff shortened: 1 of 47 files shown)


Rasbandit:
Review findings addressed: test_85 now wraps the deaf window in try/finally (guaranteed reconnect of session-scoped B — the poison-the-suite pattern the review flagged) and polls for the no-deletion invariant instead of a fixed 3s sleep; test_84's gather flake surface documented. Both tests re-validated green locally.

PR #963, opened 2026-07-08
test(e2e): force-connect precondition for test_84/85

Rasbandit:
Follow-up to #962. test_84's first CI outing (run 28936382212) failed its `wait_for_stream_connected(timeout=20)` precondition — these tests run last (highest numbers) and inherit the whole suite's socket churn, so the stream can sit mid-backoff at test start. Force a known-good connection (reconnect if down, then wait) instead of waiting out the backoff timer; same primitive test_23/test_48 use.

Test-only; the assertion surfaces are unchanged. Both tests green locally against the CRDT stack. v0.5.647.


Rasbandit:
Mechanical follow-up within the #962 review's framework (that review explicitly flagged connect-state fragility and the test_30/48 reconnect convention; this applies the same primitive to the precondition). No assertion changes; both tests green locally.

Issue #966, opened 2026-07-08
prod: OTP ssl tls_client_ticket_store crash (KMS TLS1.3 session tickets) drove 2h ECS task churn — Jul 3 05:05-07:10 UTC cluster-degraded episode


Rasbandit:
## What happened

The Jul 3 02:00-08:00 UTC `engram-prod-cluster-degraded` episode (two sustained 17m/20m splits, up to 4 simultaneous 0-peer series, `count(distributed==1)` dipping to 1) was **not** a clustering/DNSCluster defect. Prometheus shows stable task pairs until **05:05 UTC**, then ~2h of continuous ECS task churn (dozens of ephemeral task IPs, each alive 1-2 samples at `cluster_peers=0`) from **05:05 to 07:10 UTC**, with no new task definitions in that window (rev93 04:39, rev94 08:38) — ECS was replacing tasks that kept dying, then it self-recovered.

## Root cause: OTP `ssl` bug — `tls_client_ticket_store` crash on KMS TLS 1.3 session tickets

Loki (prod) shows crash bursts at **05:02:50** (exactly at churn onset), **05:37**, and **06:30 UTC**:

```
GenServer :tls_client_ticket_store terminating
** (TryClauseError) no try clause matching: {:data, :undefined, {:aes_256_gcm, :sha384},
   ~c"kms.us-east-1.amazonaws.com", <<...>>, ...}
```

cascading into the TLS-handshake `gen_statem`s (`:negotiated_protocol` / `{:start, :infinity}` calls into the crashed store). `tls_client_ticket_store` is a singleton under the `ssl` application's supervisor; repeated crashes exceed restart intensity → the (permanent) `ssl` app goes down → **VM exit** → ECS replaces the task → the replacement talks to KMS again (BootCanary wrap/unwrap) and can re-trigger. Consistent with the deaths being abrupt: **zero** `drain: starting graceful shutdown` lines in Loki during the entire churn window (the two normal drains at ~04:50 are the rev93 rollout).

We run **OTP 27.1.2** (Dockerfile `OTP_VERSION`), which is many `ssl` patch releases behind; TLS 1.3 client session-ticket handling (`session_tickets` auto path) has had several crash fixes in later 27.3.4.x patches.

## Fix options (either kills the class)

1. **Bump `OTP_VERSION` 27.1.2 → latest 27.3.4.x** in the Dockerfile (preferred: patch-level, picks up all ssl hardening + the 2026-06 ssl CVE fixes).
[...]

PR #967, opened 2026-07-08
feat(cluster): cluster-clean deploys — /health/deep join gate + drain-side disconnect after socket drain

Rasbandit:
## Problem

Rolling deploys still produce windows where a node serves traffic with `cluster_peers=0` — its `Phoenix.PubSub` is per-node, so WS clients on that node silently miss cross-node `note_changed` fan-out. This is what the `engram-prod-cluster-degraded` deploy blips (Jul 3/5/7) trace back to. The alert is deliberately untouched; this fixes the cause.

Evidence + design space: `engram-workspace/.superpowers/sdd/alert-noise-investigation.md`. The Jul 3 02:00-08:00 UTC sustained episode is a **separate defect** (OTP `ssl` `tls_client_ticket_store` crash on KMS TLS 1.3 session tickets → VM exits → 2h ECS task churn) — root-caused and filed as #966, not stretched into this PR.

## What was already true (and therefore NOT re-done)

#742/#673 (stable `RELEASE_COOKIE`, socket drainer, 30s HTTP drain) shipped Jun 24, so old+new tasks can cluster mid-roll, and DNSCluster does its first resolve+connect in `init` (`handle_continue`) — before PubSub/Endpoint even start. Cloud Map TTL (10s), DNSCluster interval (5s), ALB HC (15s/2), and deployment percentages are already tight — **no infra changes needed**.

Two gaps remained, both app-side:

## 1. Boot side: cluster-join readiness gate on `/api/health/deep`

New `Engram.Cluster.Readiness`, consulted by the ALB readiness probe when `DNS_CLUSTER_QUERY` is set (self-host/dev/test: check omitted, response shape unchanged):

| state | probe | when |
|---|---|---|
| `ready` | 200 | ≥1 BEAM peer connected |
| `alone` | 200 | Cloud Map resolves no OTHER node (first task, scale-to-1, DR) — or resolution fails (fail-open: a Cloud Map outage must not block boot) |
| `waiting` | **503** | other nodes discoverable but unjoined, within the 60s boot grace — ECS keeps old tasks serving until the new node joins |
| `grace_expired` | 200 + warn log | still unjoined after 60s of VM uptime |

**Wedge-proofing:** the gate can only hold a deploy for 60s of boot. A permanently-alone node (SG regression, cookie mismatch, the undistributed-boot class from #966) passes after grace with a `lifecycle` warning, and the untouched cluster-degraded alert remains the authority. Grace is measured from VM start, so a *later* real split never yanks tasks from the ALB (that would upgrade "degraded fan-out" to an outage). Timing envelope: 60s grace < ECS `health_check_grace_period_seconds` (120s), and worst case stays well inside the CI `Verify ECS deployment health` 300s budget (~160-200s baseline).

## 2. Drain side: disconnect peers AFTER the socket drain, not before

`prep_stop` used to `Node.disconnect` peers 5s into SIGTERM while the endpoint kept draining WS clients for up to ~25s — clients on the dying node missed cross-node fan-out for that window, and the node kept emitting `cluster_peers=0` samples whose staleness lingers ~5m (the post-deploy alert-blip signature). Peer disconnect now moves to `Application.stop/1`, which runs after the supervision tree (endpoint + UserSocket drainer + PromEx) has stopped: full fan-out until the node stops serving, zero dying-node 0-peer samples, and survivors still get a clean `nodedown` before the VM exits.

## Tests (TDD)

`Engram.Cluster.Readiness.decide/1` is a pure decision function — unit-tested across ready/alone/waiting/grace-expired, plus `check/1` seams (self-IP exclusion, undistributed node, DNS fail-open). Controller tests cover the 503 boot window, all 200 paths, and that the unclustered response shape is unchanged. Drainer tests updated for the split (drain no longer disconnects; `disconnect_peers/1` ordered + no-op-safe).

Gauntlet: `mix format` clean, `credo --strict` 0 issues, `sobelow --exit` clean, full `mix test` green.

### DIFF
--- a/lib/engram/application.ex
+++ b/lib/engram/application.ex
@@ -230,12 +230,26 @@ defmodule Engram.Application do
 
   @impl true
   def prep_stop(state) do
-    # Only drain the cluster when actually clustered (SaaS prod).
+    # Only drain when actually clustered (SaaS prod). Peer disconnect happens
+    # in stop/1 — AFTER the endpoint's socket/HTTP drain — so WS clients on
+    # the dying node keep cross-node fan-out until they've reconnected away.
     if Application.get_env(:engram, :dns_cluster_query) do
       Engram.Drainer.drain()
     end
 
     state
+  end
+
+  @impl true
+  def stop(_state) do
+    # Runs after the supervision tree (endpoint included) has stopped: safe
+    # to leave the cluster now, and survivors observe a clean nodedown
+    # before the VM exits.
+    if Application.get_env(:engram, :dns_cluster_query) do
+      Engram.Drainer.disconnect_peers()
+    end
+
+    :ok
   end
 
   # Tell Phoenix to update the endpoint configuration
--- a/lib/engram/cluster/readiness.ex
+++ b/lib/engram/cluster/readiness.ex
@@ -1,0 +1,97 @@
+defmodule Engram.Cluster.Readiness do
+  @moduledoc """
+  Cluster-join readiness for the ALB readiness probe (`/api/health/deep`).
+
+  During a rolling deploy a freshly-booted task must not take traffic while
+  its `Phoenix.PubSub` is still per-node — WS clients routed to it would
+  silently miss cross-node `note_changed` fan-out until DNSCluster connects
+  peers. This module answers "is this node safe to serve?":
+
+    * `:ready` — at least one BEAM peer is connected.
+    * `{:ready, :alone}` — no peers, but Cloud Map discovery shows no OTHER
+      node either (first-ever task, scale-to-1, disaster recovery — or a
+      discovery outage, which deliberately fails open: an unreachable
+      resolver must not block boot).
+    * `:waiting` — other nodes are discoverable but none is connected yet
+      (the deploy boot window). The probe returns 503 so ECS keeps the old
+      tasks serving until the new node has joined.
+    * `{:ready, :grace_expired}` — still unjoined after the boot grace
+      (#{div(:timer.seconds(60), 1000)}s). Passes WITH a warning so a broken
+      clustering layer (Cloud Map outage, SG regression, cookie mismatch,
+      undistributed node) can never wedge a deploy — the untouched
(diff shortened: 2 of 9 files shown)


Rasbandit:
Applied the final-review findings (TDD — each behavioral fix landed red first):

| Finding | Fix | Commit |
|---|---|---|
| **CRITICAL:** `check/1` resolved Cloud Map DNS on every `/health/deep` probe even with peers connected, and `:inet_res.lookup` default retry/timeout (~6s+) exceeds the ALB HC timeout (5s) — a hanging VPC resolver would pull every task from rotation | Peers evaluated first: steady state short-circuits to `:ready` with **zero** resolver calls (test injects a resolver that raises if invoked). Lookup bounded to `timeout: 1_500, retry: 1`; a hung/unresponsive resolver now fails open to `:alone` — proven with a real `:inet_res` call against a TEST-NET-1 blackhole nameserver, measured ~1.5s, asserted < 4s | 70bb1cf3 |
| **MINOR:** `{:ready, :grace_expired}` warned on every probe (~8/min during a sustained split) | Log-once per state-entry via `persistent_term` flag: first observation at `:warning`, subsequent at `:debug`. Test asserts the second probe emits no warning | 70bb1cf3 (+ 338d08d3 test-order hardening: flag reset in setup so grace-test ordering can't leak) |
| **MINOR:** `String.starts_with?(v, "ok")` could false-pass an "okish" value | Exact taxonomy match: `"ok"` or `"ok: " <> _` (the only shapes this controller produces) | 70bb1cf3 |
| **Version conflict:** main took 0.5.648 via #965 | Merged `origin/main` (d69d8f4e), bumped to **0.5.649** | 1dc09644 |
| **Stale infra comment:** `service_discovery.tf` per-build-cookie claim | Separate comment-only engram-infra PR: engram-app/engram-infra#822 | — |

Gauntlet: `mix format --check-formatted` ✓, `credo --strict` 0 issues, `sobelow --exit low` exit 0, full `mix test` **3543 tests, 0 failures** (re-run under the seed that exposed the log-once test-order flake).

PR #969, opened 2026-07-08
fix: wipe-echo hardening — e2e advanced-sync suite, origin device_id on delete broadcasts, no no-op fanout

### LINKED ISSUE #971: no-op DELETE /api/notes still broadcasts note_changed op=delete with empty note_id
When `DELETE /api/notes` hits a path with no live note (path_hmac lookup finds nothing — already deleted or never existed), `Engram.Notes.delete_note/3` still emits a sync broadcast: `note_changed note_id= op=delete mode=fanout` with an **empty note_id**.

Seen at scale during the 2026-07-08 replace-remote wipe (roughly half of the 33 wipe deletes were no-ops → 15+ empty-id delete events fanned out to every device). Pure noise at best; at worst a client treats an empty-id delete as path-keyed and acts on it.

Fix: skip the broadcast entirely when the delete was a no-op (nothing changed, nothing to announce). Regression test: DELETE on a nonexistent path emits no `note_changed`.

Refs: Engram-obsidian#215 (incident), #969

Rasbandit:
## Why

The 2026-07-08 replace-remote vault wipe: `wipeRemote()`'s own delete fanout came back on the live WS channel and trashed the local vault before the upload phase ran (plugin fix: engram-app/Engram-obsidian#215). This PR carries the e2e coverage plus both server-side halves.

## What

**e2e: `test_86_advanced_sync_choices.py`** — one test per advanced-sync button (push/pull × keep/delete-extras), each seeding divergent state and asserting the final note set on **both** sides via `plugin.runSyncFromChoice(...)`, with settle-window re-asserts for late echoes. test_55 only spies the modal UI — the destructive syncs had zero behavioral coverage. Red on plugin 1.11.34 with the incident signature; green with the paired plugin branch.

**Refs #970 — origin `device_id` on delete broadcasts.** REST-driven changes have no socket pid for `broadcast_from` exclusion, so the fanout reaches the very device that made the change. `EngramWeb.OriginDevice` extracts the plugin's existing `X-Device-Id` header (opaque, ≤64 bytes, valid UTF-8 only, never used for authz) and the notes delete path stamps it into the `note_changed` payload; the plugin (paired branch) drops deletes it originated. Delete-side only: upsert echoes drive client id-relocation and already have hash-based suppression — #970 stays open for those + oracle attribution.

**Closes #971 — no-op deletes announce nothing.** `delete_note/4` of an unknown/already-deleted path no longer broadcasts (the incident fanned out 15+ empty-id delete events). Mirrors `do_delete_attachment`'s existing `if deleted?` guard.

## Verification

- New ExUnit: no-op delete emits nothing; delete carries `origin_device_id`; omitted when absent; controller stamps the header; oversized header ignored.
- `mix format` / `credo --strict` / `compile --warnings-as-errors` / sobelow clean; notes+broadcast+controller suites green (201 tests). Bump 0.5.648.
- Local e2e: test_86 (all 4) + test_05 + test_10 green against the paired plugin build.


Rasbandit:
Review round applied in `d62e51e1`: attachment delete broadcasts now carry `origin_device_id` (review F7 — wipeRemote deletes attachments too), and `OriginDevice` documents the accepted same-account spoofing bound (F6; session-bound identity stays in #970). Reviewed-and-accepted, not changed: the no-op-delete broadcast removal matches the attachments precedent — a never-synced local-only copy on another device was only ever cleaned up by accident (noted for #970's convergence work).

PR #979, opened 2026-07-09
test(e2e): force-connect gate for test_82 (Resumed-pull-seed)

### LINKED ISSUE #977: e2e: extend #963 force-connect churn gate to test_82 (Resumed-pull-seed)
test_82 failed on plugin PR #216's backend/e2e run (28985769854) at its Phase 1 assert (line 76): plain live A→B delivery of a fresh seed note on a brand-new fresh_instance_pair never arrived within 30s.

Forensics (from the PR #216 verification pass):
- Failure is at ~97% of the suite: fresh ResumedA/ResumedB instances boot under end-of-suite churn — the same class PR #963 force-connect-gated for test_84/85. test_82 is not covered by that gate.
- Exonerated the #216 mint-refusal guard: it is push-side only (cannot affect inbound delivery), and `recentlyFlushed` is a private in-memory Map with 5s expiry, never serialized — a resumed device always starts with it empty, so the guard provably cannot fire in the wiped-map window.
- Local: test_82 solo on the #216 branch = 5/5 pass.

Fix: apply the same force-connect gate from #963 to test_82's fixture.

Rasbandit:
## Summary
- test_82 failed at its Phase 1 assert on plugin PR #216's e2e run (28985769854): plain live A→B delivery on a fresh `fresh_instance_pair` at ~97% of the suite, under end-of-suite socket churn — the stream can sit mid-backoff at test start.
- Same failure class PR #963 gated for test_84/85; test_82 wasn't covered. This mirrors the exact #963 pattern: `check_stream_connected` → `reconnect_stream` → `wait_for_stream_connected(20)` on both instances before the Phase 1 seed writes.
- The #216 mint guard was already exonerated in the issue forensics (push-side only; `recentlyFlushed` is in-memory, never serialized). test_82 solo = 5/5 pass locally per the issue.

Closes #977

### DIFF
--- a/e2e/tests/test_82_resumed_device_stale_idmap.py
+++ b/e2e/tests/test_82_resumed_device_stale_idmap.py
@@ -69,6 +69,15 @@ async def test_resumed_device_with_stale_idmap_syncs_both_ways(
 ):
     inst_a, inst_b, cdp_a, cdp_b = fresh_instance_pair
 
+    # This test runs near the END of the suite and inherits the whole suite's
+    # socket churn — the stream can sit mid-backoff at test start (same class
+    # PR #963 gated for test_84/85; CI run 28985769854 failed Phase 1 here).
+    # Force a known-good connection instead of waiting out the backoff timer.
+    for cdp in (cdp_a, cdp_b):
+        if not await cdp.check_stream_connected():
+            await cdp.reconnect_stream()
+        await cdp.wait_for_stream_connected(timeout=20)
+
     # Phase 1: normal sync so B earns a cursor + populated map, and learns the
     # ids of two pre-existing notes (one per direction we'll test post-resume).
     write_note(inst_a.vault_path, "E2E/Resumed-pull-seed.md", "# PullSeed\noriginal")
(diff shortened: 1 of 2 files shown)


Rasbandit:
Review round applied in `a0f2b271`: fresh_instance_pair boots brand-new instances — there is no suite churn to inherit, so the copied test_84/85 force-reconnect gate would mask genuine fresh-boot connect regressions on the one e2e that exercises resumed-device boot. Replaced with a plain `wait_for_stream_connected(30)` (tolerates slow boot, masks nothing). Lands third (0.5.656). Follow-up worth doing when a fourth test needs the gate: extract the legit churn-gate into a shared cdp helper (it's now copy-pasted in test_84/85).

Rasbandit:
Fresh evidence for the wait-gate: run 28992179975 (PR #981's branch, which carries NO gate) failed test_82 Phase 1 with `Resumed-pull-seed.md did not contain 'original' within 30s` — the fresh instances' stream wasn't connected before the Phase 1 writes. The reworked `wait_for_stream_connected(30)` gate covers exactly this without masking a broken connect (a dead connect still fails the gate loudly).

PR #981, opened 2026-07-09
fix(ci): block pngquant-bin postinstall via trustedDependencies

### LINKED ISSUE #975: CI: pngquant-bin postinstall flake — GitHub 429 on binary download kills frontend bun install
Seen on run 28983318642 job 86006918979 (e2e-browser, cold node_modules on runner-7): `bun install --frozen-lockfile` failed in 31s because `pngquant-bin`'s postinstall got `Response code 429 (Too Many Requests)` downloading its prebuilt binary from GitHub, then the source-build fallback died on missing `pkg-config` / `libimagequant.h`. Any frontend CI job (`e2e-browser`, `frontend-lint`, `frontend-audit`) that runs `bun install` on a runner with cold state can hit this — the runner pool shares an egress IP, and GitHub binary downloads bypass the local NPM mirror (:4873 serves the tarball, not the postinstall's GitHub fetch).

`pngquant-bin` is a devDependency used ONLY by `frontend/scripts/render-email-mark.ts` (one-off email-logo asset render) — the CI jobs never need the binary.

Fix directions (pick one):
- Drop `pngquant-bin` from the trusted-postinstall set so bun skips its lifecycle script; whoever runs render-email-mark locally does `bun pm trust pngquant-bin` first.
- Or install `pkg-config` + libimagequant on the runner image so the source fallback works.
- Or cache `frontend/node_modules` keyed on `bun.lock` so cold installs are rare.

Rasbandit:
## Summary
- Three CI hits on 2026-07-08 alone: `pngquant-bin`'s postinstall fetches its prebuilt binary from GitHub, which 429s on the runner pool's shared egress IP (the download bypasses the local NPM mirror), then the source-build fallback dies on missing `pkg-config`/`libimagequant.h` — killing `bun install --frozen-lockfile` for any frontend job with cold node_modules.
- The binary is only used by `frontend/scripts/render-email-mark.ts`, a one-shot email-logo render whose output PNG is committed. CI never needs it.
- Fix: pin `trustedDependencies` to today's postinstall set (`core-js`, `esbuild`, `msw`, `protobufjs`, `workerd`) minus `pngquant-bin`, so bun blocks its lifecycle script. Documented the one-time `bun pm trust pngquant-bin && bun install` step in the script header. No `|| true`, no retries — the flaky network fetch just no longer runs.

## Verification (local A/B, bun 1.3.11)
- **Control** (no change, cold install): reproduces the exact CI failure — `error: postinstall script from "pngquant-bin" exited with 1`.
- **With this change** (cold install): `bun install --frozen-lockfile` succeeds, `bun pm untrusted` state for all other packages unchanged from main, and `bun run build` (tsc + vite) passes.

Closes #975

### DIFF
--- a/docs/context/bun-postinstall-trust.md
+++ b/docs/context/bun-postinstall-trust.md
@@ -1,0 +1,54 @@
+# Context Doc: Bun lifecycle-script trust (pngquant-bin CI flake fix)
+
+_Last verified: 2026-07-08 (bun 1.3.11, against `frontend/package.json`; issue #975, PR #981)_
+
+## What This Is
+
+Why `bun install --frozen-lockfile` was flaky on cold-node_modules frontend CI jobs, and what we
+learned about bun's lifecycle-script trust model fixing it. Read this before touching
+`trustedDependencies` or re-adding a package with a network-fetching postinstall.
+
+## The Problem (#975)
+
+- `pngquant-bin`'s postinstall downloads its binary from GitHub. The CI runner pool's shared
+  egress IP gets 429'd, and the download bypasses the local `:4873` NPM mirror entirely.
+- The source-build fallback dies on missing `pkg-config`/`libimagequant`.
+- Net effect: any frontend job with a cold `node_modules` failed at install time.
+- The binary is only used by `frontend/scripts/render-email-mark.ts` — a one-shot script whose
+  output is committed. Nothing at build/test time needs it.
+
+## The Fix (PR #981)
+
+`trustedDependencies` in `frontend/package.json`, pinned to today's actually-needed postinstall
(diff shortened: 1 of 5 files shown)


Rasbandit:
Review round applied in `12205b3a`: regen recipe now warns that `bun pm trust` mutates package.json (committing it would reintroduce the #975 flake); dead core-js/protobufjs entries pruned (never execute under bun 1.3.11 — listing them misrepresented the trust surface); closed the A/B gap by running the full frontend vitest suite (761 green) against the pruned list. Lands second (0.5.655). Suggested follow-up (not blocking): a CI assert that pngquant's postinstall stayed blocked, since the exclusion semantics are empirical and version-coupled.

Issue #983, opened 2026-07-09
[p0] CRDT unbind-checkpoint pool exhaustion: bound + batch checkpoint fan-out


Rasbandit:
## Incident

A single user syncing phone + laptop on a ~230-note vault put prod into a self-sustaining sync loop (2026-07-09 ~07:26–07:53 UTC, backend `0.5.653`). DB query queue time went from a flat 9.9ms baseline to **4858ms**, driven by `DBConnection.ConnectionError` storms (30+ notes/second) from CRDT checkpoints. Loop stopped only when the user disconnected. Scope was one vault; no data lost (edits survive in the `crdt_update_log` tail-WAL), but the plaintext projection went stale and other tenants' requests were at risk during the pool starvation.

Paired with plugin issue: converge live-bound catch-up + lazy enrollment (see Engram-obsidian). Client re-handshake loop is the *trigger*; this issue is the *amplifier*.

## Root cause (B1): unbind checkpoints have no backpressure

- Each note is a `Yex.Sync.SharedDoc` GenServer with `auto_exit: true` (`lib/engram/notes/crdt_doc.ex:46`). When the last observer (socket) leaves, the room terminates and runs a **synchronous** checkpoint.
- `terminate/2 → CrdtPersistence.unbind/3` (`lib/engram/notes/crdt_persistence.ex:129-133`) → `CrdtCheckpoint.checkpoint/5`. Each checkpoint takes **two** pool checkouts (`lib/engram/notes/crdt_checkpoint.ex:76` and `:103`).
- Pool is **10 connections/node** (`config/runtime.exs:611`), shared with all REST/search/Oban work. No semaphore, no queue, no dedicated connection, no timeout override.
- On a socket drop / re-handshake, **all** of that client's rooms lose their observer at once → up to ~N-notes concurrent checkpoint txns fight 10 connections → checkout queue times out at ~2.8s → `DBConnection.ConnectionError`.
- The one **unguarded** DB call, `Accounts.get_user` at `lib/engram/notes/crdt_checkpoint.ex:50`, is outside the `do_checkpoint` rescue and raises straight out of `terminate/2` (the frame in the reported stack). Inside `do_checkpoint` the error is caught and the flush silently abandoned (`crdt_checkpoint.ex:188-196`).

## Scaling context

[...]

PR #984, opened 2026-07-09
fix(crdt): bound unbind-checkpoint fan-out (DB pool exhaustion)

Rasbandit:
## What

Bounds CRDT unbind-checkpoint concurrency so a reconnect storm can no longer exhaust the DB connection pool. This is the backend half of the 2026-07-09 sync-loop incident (see workspace incident doc + engram-app/Engram-obsidian#221 for the client half).

## Why

A client socket drop with `auto_exit: true` terminates ALL of that client's CRDT rooms at once, and each runs a full synchronous checkpoint transaction on the terminating process, holding one `Engram.Repo` connection for its duration. On a ~230-note vault that is up to ~230 concurrent checkpoints fighting a 10-connection pool. The checkout queue times out and raises `DBConnection.ConnectionError` (prod queue time went 9.9ms → 4858ms). Combined with the client re-handshake loop (#221), it was self-sustaining.

## How

A hybrid bound, so the common single-note-close case keeps its current synchronous timing and only a storm is offloaded:

- **`Engram.Notes.CheckpointGate`** — a process-global `:atomics` counter (init at boot) caps concurrent inline unbind checkpoints at 5 (under `POOL_SIZE`).
- **`unbind/3`** — under the cap, checkpoints synchronously as before (`try/after` release). Over the cap, overflows to a new bounded Oban queue instead of piling onto the pool.
- **`crdt_checkpoint` queue (concurrency 3)** + **`Engram.Workers.CheckpointNote`** — `unique: [keys: [:note_id]]` collapses a same-note storm into one job; excess jobs wait in Postgres, not as BEAM processes holding connections. The worker rebuilds the doc from durable snapshot + tail-log (`bind/3`'s recipe) and checkpoints.

### Loss-free

The tail-WAL (`crdt_update_log`) is pruned only on a *successful* checkpoint, so deferring or dropping a checkpoint loses zero edits — they replay on the next room bind. The worker also skips when a live room already owns the note (avoids racing the live timer) and when the note has no CRDT state (never blanks `notes.content`).

### Behavioral change to note

Under a storm (over the gate), unbind materialization of `notes.content`/search becomes **asynchronous** (drains via the queue). Self-healing and bounded. The uncontended path is unchanged. `get_user` is already the nil-safe variant (#954), so no raise escapes `terminate`.

## Tests

- `CheckpointGate`: acquire/refuse/release invariants.
- `unbind` routing: under the limit → synchronous checkpoint, no job; over the limit → job enqueued, no synchronous write.
- Worker: rebuilds from a real tail-log row and materializes the edit; does NOT blank content when the note has no CRDT state.

All green (`7 tests, 0 failures`); `mix format`, `credo --strict`, `sobelow` clean.

Refs #983 (incident fix complete; the lazy-enrollment scaling item in that issue is client-side in #221).

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -76,6 +76,11 @@ config :engram, Oban,
     export: 1,
     cleanup: 1,
     indexing: 2,
+    # Overflow lane for CRDT unbind checkpoints under a reconnect storm (see
+    # Engram.Notes.CheckpointGate + Engram.Workers.CheckpointNote). Kept well
+    # under POOL_SIZE (10) so draining jobs never starve REST/search of DB
+    # connections; excess jobs wait in Postgres, not on the pool.
+    crdt_checkpoint: 3,
     default: 1
   ],
   plugins: [
--- a/config/config.exs
+++ b/config/config.exs
@@ -77,9 +77,11 @@ config :engram, Oban,
     cleanup: 1,
     indexing: 2,
     # Overflow lane for CRDT unbind checkpoints under a reconnect storm (see
-    # Engram.Notes.CheckpointGate + Engram.Workers.CheckpointNote). Kept well
-    # under POOL_SIZE (10) so draining jobs never starve REST/search of DB
-    # connections; excess jobs wait in Postgres, not on the pool.
+    # Engram.Notes.CheckpointGate + Engram.Workers.CheckpointNote). Combined
+    # peak DB-connection budget on this path is inline (CheckpointGate default
+    # 3) + this lane (3) = 6 of POOL_SIZE (10), leaving headroom for REST/search/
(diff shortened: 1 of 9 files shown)


Rasbandit:
Addressed the scale/reliability code review (high-effort, 18-agent):
- **CheckpointGate → monitored GenServer** — fixes the permanent slot-leak on brutal `:kill` (deploy shutdown-timeout) that would silently pin the gate and defeat the sync fast path; also makes the limit exact (no acquire race).
- **CheckpointNote snoozes** instead of dropping when a live room owns the note.
- **Inline limit 5→3** + honest pool-budget comment (inline 3 + queue 3 = 6/10).
New tests incl. a :kill self-heal test. All green + credo/sobelow.

PR #988, opened 2026-07-10
feat(otel): drop health-check + scrape traces at the root sampler

Rasbandit:
## What

Add `Engram.Observability.TraceSampler` as the `:parent_based` **root** sampler: drop traces whose `url.path` is a health-check/scrape path, delegate everything else to the existing ratio sampler.

Dropped paths: `/metrics`, `/api/health`, `/api/health/deep`, `/` (bare-root health ping), `/socket/origin-probe/websocket`.

## Why

Measured in Tempo: **~96% of all server spans we ship are infra noise** (0.40 of 0.417 spans/sec) —

| source | path | cadence |
|--------|------|---------|
| ALB readiness | `/api/health/deep` | 15s/task (+ Ecto `SELECT 1` child span) |
| container liveness | `/api/health` | 30s/task |
| Prometheus scrape | `/metrics` | scrape interval |
| ALB origin-probe | `/socket/origin-probe/websocket` | continuous |
| root health ping | `/` | continuous |

All at **100% sampling** (`ENGRAM_OTEL_SAMPLE_RATIO=1.0`). None has diagnostic value — it's pure Tempo ingest/storage cost and dashboard noise.

## How it's correct

- **`url.path` is present at sample time.** `opentelemetry_bandit` sets it in the attribute map at span *start* (`opentelemetry_bandit.ex:226` → `start_span`), so the sampler can match it. Verified in the dep source.
- **Root `:drop` cascades.** Under `:parent_based` the sampler is consulted only for the root span; children follow the parent's `sampled` flag. So dropping the root Bandit span silently drops the Phoenix child and the `/deep` Ecto grandchild — no orphaned spans, and the app never builds/exports them (saves CPU + bandwidth + ingest, not just storage).
- **Real traffic untouched.** Decision is root-only, so a real `POST /api/mcp` trace that happens to carry `url.path=/` on a *child* span is kept. Exact-match only: `/api/health` drops, `/api/healthz` does not.
- **Total predicate.** Runs on every root span's hot path; any unexpected attribute shape returns `keep` rather than raising.

## Tests

`test/engram/observability/trace_sampler_test.exs` — 12 tests: each noise path drops, real paths + prefixes kept, missing-`url.path`/list/garbage shapes tolerated, `should_sample/7` drop + ratio-delegation (1.0 and 0.0). Green; `mix format` + `credo` + `sobelow` clean (pre-push gate).

## Scope note

Dropping `/` was confirmed safe: only pure `GET /` root traces (0.59ms single-span probes) hit the backend origin — real SPA loads serve from Cloudflare Pages per the host-split.

No infra change — entirely app-side head sampling.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -868,13 +868,18 @@ end
 # unless the OTLP endpoint is set (prod points it at the Alloy sidecar
 # on loopback; Alloy forwards to Tempo). ENGRAM_OTEL_SAMPLE_RATIO tunes
 # head sampling without a code deploy (1.0 = every trace).
+#
+# The root sampler drops health-check / scrape traffic outright (~96% of
+# span volume — see Engram.Observability.TraceSampler), then delegates the
+# rest to the ratio sampler. Under :parent_based a root :drop cascades, so
+# probe traces never build child spans or hit Tempo.
 if otlp_endpoint = System.get_env("OTEL_EXPORTER_OTLP_ENDPOINT") do
   ratio = Engram.Observability.Otel.sample_ratio(System.get_env("ENGRAM_OTEL_SAMPLE_RATIO"), 1.0)
 
   config :opentelemetry,
     span_processor: :batch,
     traces_exporter: :otlp,
-    sampler: {:parent_based, %{root: {:trace_id_ratio_based, ratio}}},
+    sampler: {:parent_based, %{root: {Engram.Observability.TraceSampler, %{ratio: ratio}}}},
     resource: [
       service: [name: "engram-backend"],
       deployment: [environment: to_string(config_env())]
--- a/lib/engram/observability/trace_sampler.ex
+++ b/lib/engram/observability/trace_sampler.ex
@@ -1,0 +1,79 @@
+defmodule Engram.Observability.TraceSampler do
+  @moduledoc """
+  Root-span sampler that drops health-check / scrape traffic outright, then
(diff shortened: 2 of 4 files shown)


Rasbandit:
Ran a high-effort code review pass; applied 3 of 6 findings, documented the rest:

**Fixed**
- **[correctness] Removed `/` from the drop list.** On self-host the backend serves the SPA index at `/` (`get "/", SpaController`), so dropping it would silently lose real page-load traces once tracing is enabled — its safety was coupled to SaaS topology (CF Pages), not to the request being noise. The three probe paths already carry the ~96% volume, so `/` wasn't needed.
- **[cleanup] Collapsed redundant `drop_paths` config threading** → `drop?/1` (it's a compile-time constant).
- **[cleanup] Removed the speculative `is_list`/proplist branch** — the `:otel_sampler` contract only delivers a map; the catch-all clause keeps the hot path total.

**Documented as deliberate head-sampling limitations (not fixable without tail sampling)**
- Probe paths are dropped regardless of status — a slow/5xx health check emits no trace. Health-check failures are diagnosed via the ALB `UnHealthyHostCount` alarm + 5xx metrics/logs, not traces.
- Only root spans are evaluated; a probe carrying an inbound `traceparent` would bypass the drop. Our probes (ALB, Prometheus/Alloy scrape, Grafana synthetic) send no trace context, so this doesn't arise.

PR #998, opened 2026-07-12
feat: frontmatter resilience: degrade cleanly and surface parse reasons

Rasbandit:
## What
Clobbered frontmatter now degrades cleanly instead of 500ing the sync pipeline, and the server surfaces *why* a note's frontmatter failed to parse so clients can show it.

## Why
Prod incident: a note with exotic YAML frontmatter (yamerl key tuples/charlists that `Jason.encode` raises on) 500'd the whole `POST /api/notes/batch`, which the plugin then retried forever. Root cause: `Frontmatter.encode_values/1` let `Jason.encode` raise uncaught, taking down the batch.

## Changes (8 tasks, subagent-driven + reviewed)
- `encode_values/1` is now total and lenient: returns `{values, bad_keys}`, never raises.
- `parse/1` returns `{:ok, order, values, degraded}`; degraded keys carry `%{key, line, snippet}` with the snippet redacted to key-only (<=200 chars), so no plaintext frontmatter leaks into the reason.
- Raw source spans for degraded keys are preserved out-of-band in a `frontmatter_raw` Y.Map (CRDT), so they round-trip verbatim through emit/project instead of being dropped.
- New `parse_status` + `parse_reason` columns (expand-phase migration). Plaintext on purpose: they describe our parser's behavior, not note content, so no encryption/HMAC.
- Per-entry batch rescue with real SAVEPOINT isolation: a failed SQL statement in one entry no longer poisons the shared transaction and falsely fails its siblings (partial success is honest again).
- Audited other bad-parse-kills-pipeline sites; `okf_fields` extraction made lenient (a degraded sibling no longer drops good keys).

## Review
Built-in code review surfaced 3 must-fixes beyond the per-task gates (normalize_doc data loss, emit projection collision, plaintext leak in the reason). All fixed. SAVEPOINT isolation independently verified RED/GREEN against a real 25P02.

## Ships with
The plugin PR (engram-app/Engram-obsidian) consumes `parse_status`/`parse_reason` + the `frontmatter_raw` map. **Merge together** — the producer here is inert-to-harmful without the consumer.

## Test
`mix test` 3707/0; dialyzer clean; `mix compile --warnings-as-errors` clean.

### DIFF
--- a/.dialyzer_ignore.exs
+++ b/.dialyzer_ignore.exs
@@ -17,17 +17,5 @@
   # binary-shape union from the leading three pattern matches, but the spec
   # has to remain `term()` so future callers don't fail type-check at the
   # boundary. Same pattern as the AAD helpers above.
-  {"lib/engram/crypto/key_provider.ex", :contract_supertype, 71},
-
-  # `decode_value/1`'s catch-all clause handles a Y.Map value written directly
-  # over the Yjs wire protocol (apply_update/2) by a buggy or hostile peer,
-  # bypassing Frontmatter's own JSON-string encoding entirely — e.g. a raw
-  # number/bool/nil/nested-map. Every ACTUAL lib/ call site happens to route
-  # through our own encode path (always a binary), so dialyzer's success
-  # typing (correctly, for the code it can see) concludes the clause is dead.
-  # It isn't: it's the only thing standing between a hostile CRDT write and a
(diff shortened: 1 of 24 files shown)


Rasbandit:
Ships with engram-app/Engram-obsidian#239 (the plugin consumer). Merge together.

PR #1003, opened 2026-07-14
fix(crdt): pace vault-channel fan-out so genesis can't starve live edits (#1002)

### LINKED ISSUE #1002: P0: vault-channel genesis fan-out is un-throttled — large-vault reconnect starves live edits ~30s+
## Summary
The per-note vault-channel CRDT fan-out (`note_yjs_update`) added in #997 (`d01f8711`) fires on **every** update with no throttling or prioritization. When a device joins/enrolls a vault, every note it enrolls emits a fan-out to all other devices on the `sync:<user>:<vault>` topic. For a large vault this is a burst of N+ broadcasts, and a **concurrent live edit's update queues behind that genesis backlog** — so a real user reconnecting with a big vault sees live edits from another device delayed tens of seconds (~30s+).

## Impact (prod)
- Large-vault user reconnects (or a second device joins) → live edits are starved behind the full-vault replay for ~30s+. Degraded real-time sync UX.
- Surfaced in e2e as the test_82 flood (now *mitigated* in the harness by isolating the resumed pair onto a private near-empty vault), but **the prod issue is independent and unfixed** — the test fix only stopped the harness from hitting it.

## Root cause / emit sites
Two per-note emit legs, both `Broadcast.emit("sync:<user>:<vault>", "note_yjs_update", %{note_id, b64, head})`, firing per note:
1. `lib/engram/notes/crdt_persistence.ex` — update_v1 chokepoint, broadcasts the **delta** on every update (channel / REST / deliver-out).
2. `lib/engram/notes/crdt_deliver.ex` `fanout_idle/3` — broadcasts **full state** for non-CRDT-origin writes.

On genesis enrollment of N notes → N+ emits flood the single per-vault topic; PubSub is ordered per-topic, so a live-edit fan-out waits behind the backlog.

## Fix direction (needs design + review — live-sync path)
- Prioritize / interleave **live-edit** updates ahead of the **genesis** backlog (a priority path), OR chunk / rate-limit the genesis fan-out.
- **Coalescing / debounce is WRONG** here: it would add latency to live edits, which is the opposite of what we want.
- Backend `crdt_channel.ex` topic-auth guard is correct and unrelated — do not touch.

## Provenance
Flagged independently by two investigations during the 2026-07-13 #355 e2e-stabilization wave. Refs #997. Deferred out of the frontmatter ship (#998/#239, now merged) to unblock that; tracked here as its own reviewed change.

Rasbandit:
Closes #1002

## Problem
The per-note vault-channel CRDT fan-out (`note_yjs_update`) added in #997 fires on **every** update with no throttling. On a large-vault reconnect / second-device join, N+ per-note broadcasts flood the single `sync:<user>:<vault>` topic; because the sync channel is fastlaned (no `handle_out`), those frames sit ahead of a later live-edit frame in every subscriber's socket FIFO mailbox → live edits delayed ~30s+.

## Fix
New `Engram.Notes.FanoutPacer` classifies each `note_yjs_update` by **per-note recency** (a server-side proxy for live-vs-genesis: genesis touches each note once; live editing hits one note repeatedly):
- **Hot** (fan-out within `hot_window_ms`, ~2s) → emit **immediately**, bypassing the pacer (live edits pay zero overhead — O(1) ETS read + inline broadcast).
- **Cold** (first touch / bulk enrollment) → per-vault queue drained at a bounded rate, **room-free and never dropped** (preserves #997's room-free convergence, just spread over time).

Both emit sites (`crdt_persistence.ex` `update_v1/4` delta, `crdt_deliver.ex` `fanout_idle/3` full-state) route through `FanoutPacer.emit/4`. A periodic ETS sweep bounds the recency table. Server-only; no plugin change; `crdt_channel.ex` topic-auth untouched.

## Config (defaults; all `Application.get_env(:engram, …)`)
`fanout_hot_window_ms=2000`, `fanout_drain_batch=20`, `fanout_drain_interval_ms=100`, `fanout_sweep_interval_ms=30000`, `fanout_pacing_enabled=true` (flip false → inline emit = instant rollback; also the test-env default so existing broadcast/e2e timing is unaffected).

## Verification
- New `FanoutPacer` tests: disabled-passthrough, cold paced drain, **hot-bypass #1002 regression**, per-vault fairness; wiring test on the real emit path.
- Full suite: 3719 tests, 0 failures (seed 0). format / credo / sobelow / dialyzer all clean.

Spec (Engram vault): `50 Engineering/_Superpowers Specs/2026-07-13-vault-channel-fanout-pacer-design.md`.


Rasbandit:
**Final whole-branch review (opus): READY WITH MINOR FOLLOW-UPS — 0 Critical, 0 Important.**

Independently re-verified: the hot/cold split defeats the #1002 starvation (call-site ETS classification keeps the live-edit path out of the GenServer mailbox); drain/timer discipline yields exactly one drain timer in prod (`ensure_draining` arms once, `handle_info(:drain)` reschedules iff non-empty); the sweep match-spec (`$1 =< now - hot_window_ms`) evicts only stale entries, never hot ones; pacing-off passthrough is byte-identical (test default off, so existing timing assertions hold); no other broadcast (`note_changed`, `crdt_doc_ready` announce, room relay) disturbed.

All findings are Minor and non-correctness (pacer telemetry, bounded-queue alarm, rollback env wiring) → tracked in #1004. The FIFO-order test gap is fine as-is (structurally guaranteed by `:queue`). Merge-ready pending CI.

Issue #242, opened 2026-05-22
Paddle catalog v1→v2 product/price update before first paying customer


Rasbandit:
## Context

Pricing v2 (Free / Starter \$10 / Pro \$20) is live on marketing site engram.page (2026-05-21), but the Paddle catalog still holds v1 product/price IDs. Per `[[project_pricing_model]]` memory: *"Paddle catalog still v1, must update before new customer."*

## Scope

Update Paddle dashboard product + price IDs to v2 catalog defined in `docs/context/paddle-v2-launch-runbook.md`:

- New product/price IDs created in Paddle dashboard (Starter \$10, Pro \$20, annual variants)
- Backend `runtime.exs` / billing config maps new price IDs to plan slugs
- Founding-member discount code configured in Paddle
- Verify webhook ingestion still parses new IDs (no field renames)
- Smoke: checkout overlay loads v2 prices in staging

## Refs

- Launch runbook: `docs/context/paddle-v2-launch-runbook.md`
- Pricing model state: `[[project_pricing_model]]`
- Design spec: `docs/superpowers/specs/2026-05-20-pricing-tiers-v2-design.md`

## Acceptance

- [ ] v2 product/price IDs live in Paddle dashboard
- [ ] Backend config maps v2 IDs → Free/Starter/Pro
- [ ] Founding-member discount tested end-to-end
- [ ] Staging checkout overlay verified

Part of v1-launch milestone. Blocks first paying customer.

Rasbandit:
## Audit finding (2026-05-22)

Backend `lib/engram/billing.ex:150` documents:

> Users must start a 7-day trial (card on file) via the Paddle overlay before syncing — the trial is configured on the Paddle price.

**Add to acceptance:** each Paddle v2 price (Starter + Pro, monthly + annual once shipped) must be created with a 7-day trial period configured at the Paddle dashboard level. This isn't optional — code expects status="trialing" on subscription creation.

Coordinate with #270 (founding-member offer mechanics — affects whether founding members get extended trial or different terms) and #276 (Free tier implementation — affects whether trial is the only path or coexists with a no-card Free option).

Rasbandit:
**Unblocked by founding-member decisions** (workspace PR #25 — engram-app/engram-workspace#25).

Per `docs/context/founding-member-offer.md`, launch with **OG-waitlist v1 grandfather only**. General `FOUNDER30` discount is **deferred**.

Catalog scope for this issue:
[...]

PR #1008, opened 2026-07-14
fix(crdt): ack crdt_msg + browser logs/traces + cut Oban trace noise

Rasbandit:
### 1. `crdt_channel` ACKs every successfully routed `crdt_msg`

With no ack, a client that attaches a reply/timeout handler treats every successful push as a timeout. Captured live: the web SPA re-handshook every open note **every ~3.5s, forever** (STEP1/STEP2/STEP2/awareness round, byte-identical, ~17 rounds/min per note). Ack point is routed-to-room: the frame is in the owner process's mailbox and `update_v1` persists it. Timeouts are now real non-delivery signals; the web SPA's timeout handler comment updated to match.

### 2. Browser logs + traces (the browser was a black box)

Diagnosing the incident required attaching a CDP WebSocket sniffer to the user's live tab. Now:

- **`frontend/src/observability/remote-log.ts`** — batched web `rlog` shipping CRDT lifecycle breadcrumbs to `POST /api/logs` (same `client_logs` pipeline as the plugin; warn+ re-emits to Loki): channel join ok/fail, push rejection/timeout with reason + note_id, re-handshake scheduled, reconnect resync, dropped inbound frames. Bounded queue, keepalive fetch, pagehide flush, never throws.
- **Failure-only `web.crdt.push` beacons** join Tempo through the existing dark-launched beacon buffer (OK pushes are keystroke-volume and would blow the 60/min beacon budget; failures are the signal).
- **`VITE_TRACING_ENABLED=true`** in the saas build env — the entire web tracing skeleton (traceparent header on authFetch + beacon buffer) existed but was dark in prod (`config.json` never set the flag).
- **BeaconSanitizer** allowlists the web span names plus `engram.note_id` (UUID-validated: anything else is a PII/cardinality risk and is dropped), `engram.route`, `engram.reason`.

### 3. TraceSampler drops orphan `engram.repo.query*` root spans

Oban's ~1s staging poll emitted a junk single-span trace per query (`oban_jobs`/`oban_peers`/source-less begin/commit), drowning Tempo — paging through them cost real time during the incident. Request-bound Ecto spans always hang off the Bandit server span and are untouched; only parentless query roots drop.

Pairs with plugin PR (live-bound REST converge + fan-out skip logging). Format/credo/sobelow/dialyzer green; full `mix test` green; frontend vitest (787), build, biome green.

https://claude.ai/code/session_01YPsYQart6AXhihbSjaqTqV

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -1226,6 +1226,7 @@ jobs:
           ENGRAM_PLUGIN_SRC: ${{ github.workspace }}/plugin-src
           CI_POSTGRES_CONTAINER: ${{ env.CI_PROJECT }}-postgres-1
           CI_MINIO_CONTAINER: ${{ env.CI_PROJECT }}-minio-1
+          CI_ENGRAM_CONTAINER: ${{ env.CI_PROJECT }}-engram-1
           E2E_CLERK_SECRET_KEY: ${{ secrets.E2E_CLERK_SECRET_KEY }}
           AUTH_PROVIDER: clerk
           E2E_VAULT_PREFIX: ${{ env.E2E_VAULT_PREFIX }}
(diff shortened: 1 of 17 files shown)


Rasbandit:
e2e proof added (commit 3d543517): `test_deaf_live_bound_note_converges_via_rest_catchup` stages the deaf-note incident deterministically (room killed via backend_rpc + handshake budget zeroed). Verified locally on the CRDT stack (this branch's image, v0.5.670):

- plugin **main**: FAILS — `File E2E/Crdt/DeafLiveBound.md did not contain 'EDIT-WHILE-B-DEAF' within 30s` (the incident)
- plugin **#242**: PASSES in 41s (REST converge)

The pre-existing xfail on `test_remote_edit_converges_while_note_is_live_bound_in_obsidian` is flipped to a hard gate (2/2 green on #242).

**Merge order: engram-app/Engram-obsidian#242 first.** Until it merges, this PR's e2e pairs with the same-name plugin branch (`fix/crdt-ack-observability`, pushed pointing at #242's head).

Issue #355, opened 2026-05-30
flake: e2e-clerk randomly fails different tests across runs on main


Rasbandit:
Observed during PR #352 (merge SHA `3dedf89`) post-merge CI: `e2e-clerk` failed `tests/test_26_large_file_rejected.py::test_large_file_rejected` with `Large note should not be pushed to server`. The PR scope was email-only; the failure is in the large-file rejection path. Re-running passed.

Cross-referencing the last five main CI failures shows a different test failing each time:

| Run | Failing test |
|---|---|
| 26678698273 | test_26_large_file_rejected.py::test_large_file_rejected |
| 26614331856 | test_47_oauth_websocket_live_sync.py (+ test_44, test_48) |
| 26612368637 | same as 26614331856 (same SHA, rerun) |
| 26609394517 | test_70_stale_file_skip.py::test_stale_local_accepts_remote |

That pattern — different tests, different runs, all under `e2e-clerk` only — looks like infra-level flake (timing, resource contention on the runner VM, container startup race) rather than per-test bugs. test_70 even self-describes the issue: "Obsidian's cached stat.mtime never reflected the backdated value within ~10s — the file-watcher didn't pick up os.utime()."

Per workspace memory feedback, rerun-pass is not acceptable on its own. Root-causing this is the work.

Likely investigation angles:
- Is e2e-clerk allocating insufficient time for the CI stack / Obsidian to settle before tests run?
- Are pytest-xdist workers (`gw0`, `gw1`) contending for shared state (DB rows, vault dirs, Clerk sessions)?
- Does the runner VM have memory pressure during the e2e-clerk job? Check the runner-vm-setup docs for the 8-runner topology — concurrent jobs across runners may starve each other.
- test_26 specifically: does the large-file rejection path use a timeout that's tight under load?

Rasbandit:
Related: #225 (e2e-clerk systemic timeout flake). Both are 'random different tests fail' patterns pointing at infra/runner contention rather than a specific test bug. Consider consolidating the infra investigation.

Rasbandit:
[...]

PR #1009, opened 2026-07-14
refactor(ci): fold e2e-api into e2e-clerk

Rasbandit:
## What

Same consolidation as #1006 (e2e-local → e2e-crdt), on the Clerk side: `e2e-api` ran the **identical Clerk-auth stack** as `e2e-clerk`, so its `api_only` suite becomes a step inside `e2e-clerk`. The split wait runs it right after stack health, hidden behind the plugin-build + Playwright background installs → ~0 added wall-clock, one fewer stack bring-up/teardown/runner-slot per run.

**e2e stacks: 5 (two days ago) → 3** (clerk, crdt, browser — one per stack shape).

## Semantics preserved

- Full `api_only --ignore local` with xdist, no `-k` (the standalone job never took a target pattern, even on `[e2e: clerk/...]` runs)
- Runs on `repository_dispatch`, skips on Dependabot (both jobs already shared those gates)
- `always()` on downstream waits + harness step: an api_only failure can't mask the Obsidian suite's signal
- `e2e-browser` rechains off clerk+crdt (e2e-api's old #355 chain slot)
- `ci` + `report-e2e-to-plugin` drop the e2e-api result (report loop is shape-agnostic)
- No fingerprint changes — e2e-api always piggybacked `skip-e2e-clerk`

## Verification

- YAML parses (26 jobs), `groups_test.sh` green, zero stale `e2e-api` refs (grep-clean)
- This PR's full-suite run is the proof; the merged step must show `api_only` green inside e2e-clerk

Refs #355

https://claude.ai/code/session_019omnopWqAww7rfxG9Yp47i

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -816,13 +816,19 @@ jobs:
           echo "Version bumped: $MAIN_VERSION → $BRANCH_VERSION"
 
   e2e-clerk:
+    # Clerk-auth E2E suite: the Obsidian plugin↔backend contract tests PLUS
+    # tests/api_only/ (absorbed from the deleted e2e-api job — same Clerk
+    # stack, one bring-up instead of two; the api_only step runs right after
+    # stack health, hidden behind the plugin-build + Playwright background
+    # installs, mirroring e2e-crdt's local-auth fold).
+    #
     # Also gate on the heavy non-e2e jobs (unit-tests ~6min, n1-compat ~3min).
     # They saturate the shared runner VM's CPU, and when they run concurrently
     # with e2e they starve its 30s live-delivery budgets — a rotating-victim
     # flake (#355) that within-suite serialization alone can't fix (e.g. test_34
     # materialized=no, test_edit_after_discovery). e2e needs a quiet VM, so it
-    # waits for the heavy jobs to finish. crdt gates on the same set; api/browser
-    # inherit the wait by chaining off clerk+crdt.
+    # waits for the heavy jobs to finish. crdt gates on the same set; browser
+    # inherits the wait by chaining off clerk+crdt.
     needs: [fingerprint, prebuild-ci-image, unit-tests, n1-compat]
     # Serialize the two heavy full-Obsidian suites (this + e2e-crdt) via a shared
     # per-ref concurrency group so they don't co-tenant the runner pool and starve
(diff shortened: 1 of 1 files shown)


Rasbandit:
## Code review (structured, post-hoc for the fold)

**Verdict: no blocking findings.** Wiring verified against the workflow graph + fingerprint machinery + a fully green solo run (api_only 63 ✓, harness 32 ✓, Obsidian 95 ✓ on one stack).

### Checked and clean
- **Job-graph rewiring**: `e2e-api` deleted with zero dangling refs (`ci` needs + `EA` check, `report-e2e-to-plugin` needs/RESULTS, `e2e-browser` rechain to clerk+crdt). RESULTS loop is shape-agnostic, so the 4→3 result string is safe.
- **Fingerprint**: no changes needed — e2e-api never had its own hash (piggybacked `skip-e2e-clerk`); `groups_test.sh` green.
- **Semantics preserved**: full `api_only --ignore local` + xdist, no `-k` (matches old job even on targeted-clerk runs); runs on dispatch, skips on Dependabot (both inherited from the clerk job's gates, identical to the old job's).
- **Shared-state audit** (the real risk of stack-sharing): `test_50_qdrant_binary_quantization` + `test_67_filtered_search_hmac` only READ the now-shared Qdrant collection (config probe + point queries — no drop/recreate). DB/user state is per-user-isolated, same pattern already proven by the #1006 crdt fold.
- **No new secret exposure**: `E2E_CLERK_SECRET_KEY` was already in this job (orphan reap step). The api step adds no template-interpolated input (no injection surface).

### Nits (non-blocking)
1. **Slower failure path on stack death**: `Wait for plugin + Playwright` has `if: always()`, so if the stack wait exits 1 the job can spend up to ~5 min waiting on plugin markers before the (always-run) Obsidian step fails fast. Pre-existing pattern from the crdt fold; acceptable, but a `stack.marker` short-circuit in the second wait would tighten it.
2. **Duplication debt grows**: the split-wait machinery is now copy-pasted in clerk + crdt. Known debt — the queued composite-action extraction (`.github/actions/e2e-stack`) is the fix; don't let a third copy appear before that lands.
3. Job id `e2e-clerk` now undersells the job (it also carries the api_only slice). Kept deliberately for marker/ruleset continuity; the header comment carries the truth.

https://claude.ai/code/session_019omnopWqAww7rfxG9Yp47i

PR #1018, opened 2026-07-15
fix(web): refresh socket token on connect error, not just wake triggers

Rasbandit:
## Problem (prod, 2026-07-15)

Clerk session tokens live 60 seconds. The SPA socket's `params()` re-reads `latestToken` on every phoenix reconnect attempt, but `latestToken` only changed in `connectChannel` / `reconnectWithFreshToken`, whose triggers are focus / online / visibilitychange. Those triggers can simply never fire (unattended tab; and observed live: even an actively used tab), so any phoenix-initiated reconnect >60s after the last explicit connect replays an expired token forever: ~9,300 socket 403s/day from single tabs, one frozen token replayed for 8 hours at ~7s cadence, zero live sync until page reload. Server-side these land as the misleading `signature_error` auth-reject bursts (label fix tracked separately).

## Fix

Refresh the token where the failure actually surfaces: `socket.onError` fires on every failed (re)connect attempt, so refresh `latestToken` there. The next backoff step (phoenix caps at 5s) reads `params()` and the loop self-heals. Properties:

- **Single-flight**: one refresh in flight regardless of error cadence; clerk-js caches tokens ~60s so per-attempt refresh is cheap.
- **Generation-guarded**: a refresh resolving after a teardown/vault switch does not touch the next connection's token (test pins this).
- **Fail-safe**: if Clerk can't mint (session refresh failing, network down), the previous token is kept and the next error retries — no crash, no `no_auth` regression.

Full triage context: `engram-workspace docs/context/spa-stale-socket-token-loop.md`.

## Tests

`frontend/src/api/channel-token-on-error.test.ts`: refresh-on-error updates `params()`; failed mint keeps prior token; post-teardown refresh can't pollute the next connection. Sibling channel test mocks gained the `onError` surface. Frontend suite 793/793, `tsc` and `biome lint` clean.

https://claude.ai/code/session_019omnopWqAww7rfxG9Yp47i

## Also in this PR: drain hygiene (`connection: close` while draining)

Half-open connections — the root of the wedged-request class fixed client-side in Engram-obsidian#248 — are born during deploy drains: the ALB deregisters the task but clients keep reusing pooled keep-alive connections. Once `Engram.Drainer.drain/1` starts, `EngramWeb.Plugs.DrainConnClose` stamps `connection: close` on every response (Bandit closes after the response), so a client's next request opens a fresh connection routed to a live task instead of hanging on a doomed one. Outside the drain window the plug is a single `:persistent_term` read. Tested in `drain_conn_close_test.exs`; full suite 3733/0.

### DIFF
--- a/frontend/src/api/channel-crdt.test.ts
+++ b/frontend/src/api/channel-crdt.test.ts
@@ -39,6 +39,7 @@ vi.mock("phoenix", () => ({
 		connect() {}
 		disconnect() {}
 		onOpen(_cb: () => void) {}
+		onError(_cb: () => void) {}
 		channel(topic: string, params?: any) {
 			return mkChannel(topic, params);
 		}
--- a/frontend/src/api/channel-onopen.test.ts
+++ b/frontend/src/api/channel-onopen.test.ts
@@ -16,6 +16,7 @@ const {
 		Object.assign(this, {
 			connect: vi.fn(),
 			disconnect: vi.fn(),
+			onError: vi.fn(),
 			onOpen,
 			channel: vi.fn(() => ({ on: channelOn, join, leave: vi.fn() })),
 		});
--- a/frontend/src/api/channel-reconnect.test.ts
+++ b/frontend/src/api/channel-reconnect.test.ts
@@ -13,6 +13,7 @@ interface MockSocket {
 	disconnect: ReturnType<typeof vi.fn>;
 	isConnected: ReturnType<typeof vi.fn>;
 	onOpen: ReturnType<typeof vi.fn>;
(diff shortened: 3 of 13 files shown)


Rasbandit:
Review wave applied (commit 4f2e1b9f): tokenRefreshInFlight resets on disconnectChannel (a hung getToken across a vault swap could permanently re-freeze the token loop — CONFIRMED); new test file biome-formatted (frontend-lint was red); Drainer's :persistent_term flag now isolated in tests (full mix test was red — now 3733/0). Frontend 794/0, tsc + biome green, format + credo --strict clean.

Issue #251, opened 2026-05-23
Customer support inbox (support@engram.page) + SLA in ToS


Rasbandit:
## Context

Paid customers need a place to send "I paid but didn't get my plan" / "I want to cancel" / "Found a bug" messages. No support inbox today.

## Scope

**Minimum:**
- `support@engram.page` MX → forwarding to a mailbox you check daily
- `privacy@engram.page` (already referenced in #250) routed too
- SLA documented in ToS (#248): "We aim to respond within 2 business days"
- Auto-reply confirming receipt (optional, low-effort)

**Future (post-v1):**
- Helpscout / Plain / Front integration when volume warrants
- Knowledge base / FAQ page

## Acceptance

- [ ] DNS MX for engram.page routes support@ + privacy@
- [ ] Mailbox set up + tested with a send
- [ ] SLA language added to ToS (#248)
- [ ] Auto-reply (optional) configured

## Refs

- ToS umbrella: #248
- GDPR runbook (uses privacy@): #250

Rasbandit:
Master v1 launch-day procedure now lives at `docs/context/launch-day-procedure.md` in engram-workspace (engram-app/engram-workspace#27, merged). This issue is **row 9 (support inbox + SLA)**.

Update the runbook inline if anything here changes the pre-flight order or adds an abort signal.

PR #1026, opened 2026-07-17
feat(crdt): deleted_since tombstone cursor in catch-up heads

Rasbandit:
## Why

`crdt_catchup_heads` conveyed only surviving notes, so a client could not learn about notes **deleted on another device while it was offline** (delete == absence, which is ambiguous vs. a first-discovery note). This adds a bounded tombstone cursor so the plugin can authoritatively trash cross-device deletes on catch-up — closing the offline-delete resurrection gap for the CRDT-sockets-only sync (#251).

## Contract

Request gains optional `deleted_since: integer | nil` (last per-vault `seq` the client reconciled). Reply:

```elixir
{:ok, %{heads: %{id => %{path, head}}, deleted: [note_id, ...], seq: <current vault change_seq>}}
```

`vault_catchup/3` snapshots `Vaults.current_seq` first, then bounds BOTH the heads read and the deleted read by that snapshot (no lost/double-counted rows). `deleted` = tombstoned note_ids in `(since, current]`. Fresh client (`nil`) → `deleted: []` + current `seq` to seed the cursor. `vault_heads/2` behavior unchanged (backward-compatible; plugin ignores the new fields until its consumer lands).

Backend-only. Plugin consumer (trash `deleted` ids, persist `seq`) is a separate task on #251. Uses `notes.seq` (per-vault monotonic, stamped on every change incl. delete). Local: compile (warnings-as-errors), format, credo, `mix test` (75/75) all clean.

### DIFF
--- a/lib/engram/notes/crdt_transport.ex
+++ b/lib/engram/notes/crdt_transport.ex
@@ -14,7 +14,7 @@ defmodule Engram.Notes.CrdtTransport do
   import Bitwise
   import Ecto.Query
 
-  alias Engram.{Crypto, Notes, Repo}
+  alias Engram.{Crypto, Notes, Repo, Vaults}
   alias Engram.Logger.Metadata
   alias Engram.Notes.{CrdtBridge, CrdtPersistence, CrdtRegistry, CrdtUpdateLog, Note}
   alias Yex.Sync.SharedDoc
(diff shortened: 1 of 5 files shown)


Rasbandit:
Closing as superseded by #1032 (now merged).

#1032's `crdt_catchup_since` serves `Notes.list_changes_by_seq`, which does **not** filter `deleted_at` — so cross-device delete tombstones already ride the seq-ordered catch-up op-log as delete ops ("Tombstones ride the same feed"), applied via `applySyncChange`. That closes the exact offline-delete-resurrection gap this PR targeted.

Additionally, #1032 added a `complete` completeness contract to `crdt_catchup_heads` (`{heads, complete}`) as a second defense that the plugin's destructive offline-delete reconcile gates on. This PR's diff would **revert** that contract, so merging it now would regress #1032. Hence: close, not rebase.

PR #1056, opened 2026-07-21
feat(crdt): seq on live fan-out + gap-heal drop seam (single-path D2)

Rasbandit:
Phase D2 of the single-path CRDT migration (backend half; pairs with plugin PR of the same branch name).

## What
- Attach the vault-global change seq (`Note.seq`) to both `note_yjs_update` fan-out payloads (`crdt_deliver.ex` idle full-state + `crdt_persistence.ex` per-delta). Additive and backward-compatible: old plugins ignore the key; a missing row yields `"seq" => nil`.
- Seq fetch costs one query: folded into the existing crdt_head-invalidation `update_all` via `select: n.seq`, with a select-only fallback read when the guard skips (never drags `crdt_state_ciphertext`).
- `FanoutPacer.test_drop_next/2`: e2e-only drop seam (persistent_term toggle, production no-op) to stage a lost live broadcast deterministically.
- New deterministic e2e `test_seq_gap_heal.py`: drops note X's fan-out, edits note Y, asserts X converges on device B via socket seq-replay and that the plugin logged the heal (`gap-heal fired`), pinning the mechanism rather than eventual content.

## Design (user-approved: seq is a heal-trigger, never a gate)
The plugin applies every live Yjs delta unconditionally and uses seq purely as a behind-detector: an observed integer seq > cursor means unconsumed feed entries exist, firing its single-flight socket catch-up. Live ops never advance the cursor (the replay is the sole cursor writer) — this survives stale seqs between checkpoints, seq aliasing across write kinds, and the unordered cross-node multi-note stream.

This is the prerequisite that lets Phase E delete the plugin's REST recovery path without regressing deaf-note.

## Testing
- Full `mix test` 3779/0; format, credo --strict, dialyzer clean.
- New/extended unit tests for both fan-out payloads (seq value asserted, both select and fallback branches).
- Paired e2e runs against the same-name plugin branch.

https://claude.ai/code/session_01VK9qUcRtS3tiW9wNhw83y7

### DIFF
--- a/e2e/tests/crdt/test_seq_gap_heal.py
+++ b/e2e/tests/crdt/test_seq_gap_heal.py
@@ -1,0 +1,102 @@
+"""D2 seq-gap-heal proof (single-path CRDT sync, Phase D2).
+
+A live fan-out that never reaches a device must be healed by the NEXT
+delivered op: every live op carries the vault-global `seq`, the plugin
+compares it to its persisted catch-up cursor, and a `seq` ahead of the
+cursor fires the single-flight socket seq-replay (`crdt_catchup_since`).
+
+Deterministic stage: `FanoutPacer.test_drop_next/2` swallows note X's
+fan-out server-side (a lost broadcast — the real-world class behind the
+"received=no materialized=no" delivery flakes). A later edit to note Y
+fans out normally with a later seq; B is behind, heals, and the replay
+carries X's missed edit. No `trigger_full_sync` after the drop — the heal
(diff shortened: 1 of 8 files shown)


Rasbandit:
Superseded by #1058 (merged), which ships seq gap-heal on the same wire contract. Auditing the merged implementation against this branch's findings (hot-path seq fetch cost, fan-out payload shape, deterministic drop-seam e2e); gaps land as a follow-up PR. Branch retained for reference.

PR #1085, opened 2026-07-23
feat(ci): auto-alert on chronically flaky e2e suites

Rasbandit:
## What

Closes the loop on the nightly flake ledger (#1078). The ledger recorded suite pass/fail to the `ci-ledger` branch, but nothing read it — "monitoring" meant a human remembering to open the branch. Now it pokes us.

- **Auto-alert (`ledger-append`):** opens a `flaky-suite` GitHub issue for any suite red on **≥40%** of the trailing 14 nights (**min 5 nights** of data, so it can't scream off one bad night), and auto-closes it on recovery. Best-effort — a `gh` hiccup emits `::warning::`, never fails the measurement job.
- **`scripts/flake_ledger_rates.jq`:** extracted the trailing-14 dedup+window math so the step-summary table and the alert gate share one definition.
- **`scripts/flake_ledger_rates_test.sh`:** offline self-check (dedup, windowing, rate math). Runnable: `bash scripts/flake_ledger_rates_test.sh`.
- **`docs/context/testing-architecture-migration.md`:** the report-only-e2e rationale `verify.yml` referenced twice but that was never written.

## Notes

- The `ledger-append` job is schedule-only, so this can't be exercised in PR CI — hence the offline test. First real data lands on the next 06:00 UTC nightly; alerts stay silent until ≥5 nights accrue (warm-up guard).
- Gotcha handled: the job checks out the `ci-ledger` orphan branch mid-run (no `scripts/` there), so the jq filter is `cat`'d into a var before the branch switch.
- e2e suites are report-only, so this gates on unit/lint/etc.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -3588,6 +3588,7 @@ jobs:
     runs-on: [self-hosted, linux, x64, isolated, light]
     permissions:
       contents: write
+      issues: write   # auto-open/close a flaky-suite issue when a suite crosses the threshold
     env:
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       GH_REPO: ${{ github.repository }}  # gh CLI can't infer repo from the LAN insteadOf-rewritten origin, see docs/context/gh-cli-runner-url-rewrite-and-release-repair.md
@@ -3608,6 +3609,13 @@ jobs:
           DATE=$(date -u +%F)
           RUN_ID="${GITHUB_RUN_ID}"
           SHA="${GITHUB_SHA}"
+
+          # Read the rate filter NOW, while the working tree is still on the
+          # default branch — the append step below checks out the ci-ledger
+          # orphan branch, which does not contain scripts/, so the file would
+          # vanish. Both consumers (summary table + alert gate) use this one
+          # program, so they can never disagree on a suite's flake rate.
+          RATES_JQ="$(cat scripts/flake_ledger_rates.jq)"
 
           # result must be one of success/failure/skipped for the ledger schema;
           # anything else (cancelled, timed_out, ...) folds into failure — it's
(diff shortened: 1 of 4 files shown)


Rasbandit:
### Code review

Ran a 5-agent review (CLAUDE.md compliance, bug scan, git history, prior-PR patterns, code-comment guidance). Two real issues found and **already fixed in 2bbf251a**; one minor note deferred.

**1. `gh issue list` failure could file duplicate issues** (flagged independently by two reviewers)

The `|| echo '[]'` fallback meant a transient `gh` blip made every over-threshold suite look un-tracked, so the pass would `create` a second `flaky-suite` issue instead of `comment`-ing the existing one — a slow issue leak over a nightly horizon.

https://github.com/engram-app/Engram/blob/b65b95593416fef60ae8cb381cb5b6a4d8e5031f/.github/workflows/verify.yml#L3718-L3720

Fixed: the pass now skips (and warns) when the list call fails, rather than assuming empty state.

**2. The new self-check wasn't wired into CI**

`scripts/flake_ledger_rates_test.sh` guards the extracted rate filter, but nothing ran it — the comment claimed protection that didn't exist, and the `ledger-append` job is schedule-only, so a broken filter would hide until the next nightly.

https://github.com/engram-app/Engram/blob/b65b95593416fef60ae8cb381cb5b6a4d8e5031f/.github/workflows/verify.yml#L3692-L3693

Fixed: added a `flake-ledger rate logic self-check` step to the `static-checks` job (runs every push).

**Deferred (minor, follow-up):** if a suite is ever renamed/removed from the ledger, its old rows never prune, so a `flaky-suite` issue open at removal time could never auto-close. Not worth building until a suite is actually removed.

Everything else (dedup/window semantics unchanged from #1078, `set +e` best-effort scoping, `GH_REPO`/orphan-branch handling, threshold/doc agreement, YAML block-scalar indentation) checked clean.

Issue #1088, opened 2026-07-23
Dead inbound surface: sync_channel push_note/delete_note/rename_note/pull_changes + GET /vault/heads


Rasbandit:
From the plugin CRDT audit: the plugin sends NONE of sync_channel's inbound ops (it joins sync: only for broadcasts), and plugin PR #303 removed the last GET /vault/heads caller (capability probe deleted — capability is the crdt: join ack now). If web/MCP don't use them either: delete the four handle_in clauses (~120 lines), the /vault/heads route (router.ex) + CrdtSyncController.vault_heads. Verify web SPA + MCP first.

PR #1095, opened 2026-07-23
fix: gate crdt: channel writes on DEK rotation (T3.7, #1092)

### LINKED ISSUE #1092: T3.7 gap: crdt: channel has no RotationGate — DEK rotation doesn't gate the actual write path
Found during #1088's review: the deleted sync_channel inbound ops were the only RotationGate-gated channel writes — but they had no callers. The real write path (crdt_msg / crdt_create / crdt_delete on the crdt: channel) has NO rotation gating: during a DEK rotation a live WS client keeps writing and encrypting mid-rotation, exactly what the T3.7 gates exist to prevent (the HTTP plug blocks even reads). Enforcement today: HTTP RotationLockCheck plug + Oban worker snoozes only. Needs: gate crdt: channel writes (join-time check + per-op check, or drain/kill sockets on rotation start in user_dek_rotation.ex).

Rasbandit:
Closes #1092

## Problem

Found in #1088's review: the deleted `sync_channel` inbound ops were the only RotationGate-gated channel writes — but they had no callers. The real write path (`crdt_msg` / `crdt_create` / `crdt_delete` on the `crdt:` channel) had **no** rotation gating. During a DEK rotation a live WS client kept writing and encrypting mid-rotation — exactly what the T3.7 gates exist to prevent (the HTTP `RotationLockCheck` plug blocks even reads).

## Fix — two guards

1. **Join gate** — `join_authenticated` calls `RotationGate.check/1` (re-reads the user row, so a reconnect mid-rotation is caught even when the socket's user struct predates the lock) and refuses with `reason: "rotation_in_progress"` while the lock is held.
2. **Socket drain** — `rotate_user` broadcasts `"disconnect"` on the `user_socket:{id}` topic the instant the lock is acquired (the proven `SessionInvalidator` pattern; Phoenix's `Socket.__info__` terminates the socket). Drained clients reconnect → hit the join gate → refused until the lock releases.

Timing is safe: `DekCache.invalidate` and the lock release both happen at `final_flip` (rotation end), so during the sweep there is no live writer AND no reconnect can succeed; the first successful rejoin sees the new DEK as authoritative.

### Rejected: per-op check

A `RotationGate.check` on every `crdt_msg` frame would add a DB round-trip to the hottest edit path to close only a single in-flight frame at the drain instant. Drain-on-start is the issue's own suggested approach and is right-sized.

## Tests (TDD, RED first)

- `crdt_channel_test.exs` — join refused while the user lock is held (fresh struct, as a reconnect carries).
- `user_dek_rotation_test.exs` — `rotate_user` broadcasts the disconnect once the lock is acquired; does NOT broadcast when the lock is already held.

## Gauntlet

format · credo --strict (no issues) · dialyzer (passed) · `mix test` 3767 tests, 0 failures.

### DIFF
--- a/lib/engram/crypto/user_dek_rotation.ex
+++ b/lib/engram/crypto/user_dek_rotation.ex
@@ -32,6 +32,7 @@ defmodule Engram.Crypto.UserDekRotation do
   import Ecto.Query, only: [from: 2]
 
   alias Engram.Accounts.User
+  alias Engram.Auth.SessionInvalidator
   alias Engram.Crypto
   alias Engram.Crypto.{DekCache, Envelope, RotationLock}
   alias Engram.Crypto.KeyProvider.Resolver
@@ -77,9 +78,9 @@ defmodule Engram.Crypto.UserDekRotation do
          {:ok, _locked_at} <- RotationLock.acquire(user_id) do
       # T3.7 (#1092): with the lock held, no new crdt: socket can join
       # (RotationGate on join). Drain the ones already open so live WS clients
-      # stop writing/encrypting mid-rotation — the socket id topic disconnects
-      # every socket for this user; they reconnect and hit the join gate.
-      _ = EngramWeb.Endpoint.broadcast("user_socket:#{user_id}", "disconnect", %{})
+      # stop writing/encrypting mid-rotation — they reconnect and hit the join
+      # gate. Same force-disconnect the entitlement/auth paths already use.
+      SessionInvalidator.disconnect_user(user_id)
 
       new_dek_version = (user.dek_version || 1) + 1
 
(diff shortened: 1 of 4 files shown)


Rasbandit:
Addressed review in f7a64b50:
- **Major (test didn't pin the re-read):** the join-refused test now builds the socket from the **unlocked** user struct and locks the DB row *after* — refusal proves `check/1` re-reads (a stale-struct `check_user` would wrongly allow it). Added a lock-clears → join-succeeds test so the gate can't go sticky.
- **Minor (dup):** drain now calls `SessionInvalidator.disconnect_user/1` instead of an inline broadcast.
- **Nit:** gate matches `{:error, :rotation_in_progress}` explicitly; a deleted user falls through to `unauthorized`.

The rotation-end `DekCache.invalidate`-after-txn window you flagged is pre-existing + identical on the HTTP plug path — out of scope here, agreed.

PR #1118, opened 2026-07-25
fix(editor): render Obsidian callout icons + full alias map

Rasbandit:
## What

Rendered mode in the SPA editor only styled four callout types and never hid the raw `> [!type]` marker — so `> [!important]` showed literal source in a grey box, while Reading mode already drew a lucide flame for the same note.

## How

`@portaljs/remark-callouts` (already a dep, already what Reading mode renders from) exports `defaultConfig.types`: the full Obsidian map of base types (`tip: {keyword, color, svg}`) plus alias strings (`important: "tip"`). `callout-decoration.ts` now resolves through it, so both views share one source and can't drift.

- Header line becomes a `Decoration.replace` widget: lucide icon + title, falling back to the capitalized keyword when there's no custom title.
- `CALLOUT_START_RE` matches the fold marker (`[!note]-`) so it stops leaking through as a stray `-`.
- Per-type color arrives as an inline `--callout-color`; **deleted** the hardcoded four-color list in `callout.css`. All ~25 Obsidian types and aliases are now covered instead of 4.

Still strictly view-only — no `EditorState.doc` mutation, so the yCollab `Y.Text` binding is untouched. The existing reveal-raw-on-cursor path covers the widget (replace is only emitted when the block is inactive).

## Testing

6 new tests in `callout-decoration.test.ts` (alias→flame, marker hidden, custom title kept, keyword fallback, fold marker, unknown type doesn't crash). Full frontend suite green: **892 tests / 153 files**. `tsc --noEmit` + prod build clean.

Manual smoke still owed: open a note with `> [!important]` in Rendered mode and confirm the flame + teal border, and that clicking into the block reveals raw source.

https://claude.ai/code/session_013fXPYtAKeUBLKjW9QmVRpZ

### DIFF
--- a/frontend/src/viewer/editor/callout-decoration.test.ts
+++ b/frontend/src/viewer/editor/callout-decoration.test.ts
@@ -42,6 +42,89 @@ describe("calloutDecoration", () => {
 		expect(view.dom.querySelectorAll(".cm-callout").length).toBe(0);
 	});
 
+	test("renders the lucide icon for an alias type (important -> tip -> flame)", () => {
+		const doc = "before\n\n> [!important] Read this\n> body\n";
+		view = new EditorView({
+			state: EditorState.create({ doc, extensions: [calloutDecoration] }),
+			parent: document.body,
+		});
+		expect(view.state.doc.toString()).toBe(doc); // view-only
+		const title = view.dom.querySelector(".cm-callout-title");
+		expect(title?.innerHTML).toContain("lucide-flame");
+	});
+
+	test("hides the raw marker and keeps a custom title", () => {
+		const doc = "> [!note] My Title\n> body\n\ntail\n";
+		view = new EditorView({
+			state: EditorState.create({
+				doc,
+				selection: { anchor: 30 },
+				extensions: [calloutDecoration],
+			}),
+			parent: document.body,
+		});
+		const title = view.dom.querySelector(".cm-callout-title");
+		expect(title?.textContent).toBe("My Title");
+		expect(view.dom.textContent).not.toContain("[!note]");
+	});
+
+	test("falls back to the capitalized keyword when there is no custom title", () => {
+		const doc = "> [!important]\n> body\n\ntail\n";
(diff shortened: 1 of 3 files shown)


Rasbandit:
### Code review

Found 2 issues, both fixed in 23985fa8:

1. Crash on a callout type that names an inherited Object property. `defaultConfig.types` is a plain object, so `types["__proto__"]` returns `Object.prototype` — it passes the `typeof === "object"` check and produces a style with `keyword`/`color`/`svg` all `undefined`, so the no-title path called `capitalize(undefined)`. CM6 catches a ViewPlugin exception and disables the plugin, so the real symptom is every *other* callout in the note losing its decorations. Lookup is now own-property-only.

https://github.com/engram-app/Engram/blob/deffcc7f285b62df653a58986e578867bbda1a06/frontend/src/viewer/editor/callout-decoration.ts#L30-L36

2. `.cm-callout`'s `border-left` was dead CSS. Every callout line is also an Atomic blockquote line, and `blockquote-depth.css` targets those with `.cm-line.cm-atomic-blockquote { border-left: none }` — 2 classes vs 1, so specificity won regardless of import order. Predates this PR (the four hardcoded colors it replaced were equally inert). Selector is now `.cm-line.cm-callout`, and the nesting rail is dropped on callout lines so the per-type bar is what shows.

https://github.com/engram-app/Engram/blob/deffcc7f285b62df653a58986e578867bbda1a06/frontend/src/viewer/editor/callout.css#L1-L9

Checked and clear: no repo guideline violations; the replace decoration stays inline and never spans a line break, so it does not re-trigger the block-decoration-in-ViewPlugin crash fixed in 2b8adec9; still view-only, so the yCollab `Y.Text` binding is untouched; `innerHTML` only ever receives the trusted SVG constant, titles are appended as text nodes; alias resolution is one hop deep, which matches the actual map.

The CSS fix has no automated coverage (happy-dom does not resolve the cascade) and needs a manual check in a real browser.

PR #1126, opened 2026-07-26
ci(actions): vendor runs-on/cache to stop per-job codeload re-download

Rasbandit:
Parked from a worktree cleanup — no PR, local-only branch. Draft so it's recoverable; decide later.

Vendors the `runs-on/cache` action into `.github/actions/runs-on-cache` so CI jobs stop re-downloading it from codeload, plus the `verify.yml` wiring.

Audit notes:
- Absent from `main`.
- **Large**: ~362,900 added lines, almost entirely committed `dist/` JavaScript. That's the main reason it was worth preserving rather than deleting — it's tedious to reconstruct — but also the main reason to think before merging.

🤖 Parked by [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/actions/runs-on-cache/LICENSE
+++ b/.github/actions/runs-on-cache/LICENSE
@@ -1,0 +1,22 @@
+
+The MIT License (MIT)
+
+Copyright (c) 2018 GitHub, Inc. and contributors
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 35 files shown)


Rasbandit:
Closing after measuring the problem this solves. **The cost is ~0.6s of critical-path time per CI run — not worth 362,900 committed lines.**

### Measurement
Parsed the runner logs of a real full run (30228376076), timing every `Download action repository 'runs-on/cache@...'` against the next log line:

| | |
|---|---|
| Distinct downloads per run | **5** (one per job that uses it, not per call site) |
| Aggregate wall-clock | **1.94s** |
| Mean per download | **0.39s** |
| **Critical-path cost** | **~0.61s** (the 5 jobs run in parallel, so the slowest one is the real cost) |

Per job: migration gates 0.61s, lint 0.48s, unit-tests 0.35s, prebuild-mix 0.30s, e2e-browser 0.19s.

The ~18 `uses:` call sites in `verify.yml` are misleading — the runner downloads an action once per job and reuses it across steps, so the fan-out never materialized.

### Why that changes the verdict
Vendoring buys back **0.6s per run** and costs:
- 362,900 lines of committed third-party `dist/` JavaScript, paid by every clone and checkout forever
- No Dependabot coverage — security updates become a manual re-vendor
- An unreviewable diff, permanently

That trade is upside-down.

### If the download ever does become a real cost
Fix it at the runner layer, not in the repo — cache or mirror the action tarball on the self-hosted runner VMs (`docs/context/runner-vm-setup.md`). That fixes *every* action at once, including `actions/checkout` (12 downloads/run) and `actions/create-github-app-token` (10/run), which are both downloaded more often than `runs-on/cache` is.

Branch `ci/vendor-runs-on-cache` is left in place — nothing is lost if this needs revisiting.

Issue #1087, opened 2026-07-23
Seed the Y.Doc from note content server-side on first room open — kills the client getNote race-closer


Rasbandit:
From the plugin CRDT audit: plugin's materializeEmptyDiscovered does a REST getNote inside the CRDT hot path, self-documented as 'Proper fix is server-side (seed the Y.Doc from REST on first room open); this is the client-side race-closer' (empty-STEP2 vs REST race, test_38/43 class). Do the seed in crdt_channel.ex room open, then delete the client cross-check + the legacy hash-only getNote branch in plugin handleStreamEvent.

PR #1141, opened 2026-07-29
refactor(crdt): the server stops authoring note content

Rasbandit:
Refactor #1 of the Relay-parity plan: remove the third writer.

## The problem

`CrdtPersistence.bind` seeded a bound room's doc from `notes.content` whenever the doc projected an empty body:

```elixir
if applied == [] and CrdtBridge.body_of(doc) == "" do
  seed_from_content(doc, note, user)
end
```

That made the **server** a third writer of note content, alongside the client and the disk flush. Reconciling three representations — doc / `notes.content` / disk — is the shape of every "which copy is authoritative right now" bug we've shipped, including the `#1087` empty-STEP2 class this seed was itself tuned to dodge.

Relay has no equivalent: its server (a y-sweet fork) is a document store that never authors content.

## The change

The seed is gone. `notes.content` becomes a **derived projection** maintained by the checkpoint materializer; the doc is the only authority.

## Why it's safe

`upsert_note` **already** merges incoming plaintext into the note's persisted CRDT state, roomlessly (`notes.ex:1395+` — `doc_from_state` → `replay_tail` → `merge_plaintext_*`). Every REST / MCP / web write therefore leaves the body in `crdt_state` without a room ever binding — exactly what the bind-time seed was compensating for.

A new test pins that property directly, since it's the load-bearing reason the seed can go:

```
a plaintext write leaves the body in the note's persisted CRDT state
```

## What was tried and rejected

An earlier pass also made `CrdtDeliver` start a short-lived observed room per non-CRDT write, to ingest plaintext at write time. It was **redundant** (the write path already covers it) and **harmful**: `CheckpointNote` deliberately snoozes while a live room owns a note, so a room-per-write starved checkpointing — which would have left the derived `notes.content` permanently stale. 23 test failures made it obvious. Reverted; the no-room branch now documents why nothing is needed there.

## Tests

Rewritten rather than deleted — those shapes are real historical hazards, so they now pin the **new** invariant:

| Test | Asserts |
|---|---|
| `bind/3 does NOT seed the doc from notes.content` | the seed is gone |
| `bind/3 does NOT resurrect notes.content over an empty-projecting snapshot (#1087)` | doc wins over a divergent row |
| `bind/3 does NOT resurrect notes.content over a frontmatter-only snapshot` | same, body-emptiness shape |
| **`a plaintext write leaves the body in the note's persisted CRDT state`** | **why the removal is safe** |
| `a plaintext write routes frontmatter into Y.Map, not body Y.Text` | codec path preserved |

Those divergent snapshot-vs-content shapes are now only constructible by hand — the write path keeps the two consistent.

**Backend 2578 tests / 0 failures.** `compile --warnings-as-errors`, `mix format`, `credo --strict`, `sobelow` all clean.

## Behaviour change worth knowing

A note whose content exists **only** in `notes.content` with no CRDT state — a genuinely pre-CRDT row — will now bind to an empty doc rather than being rescued at bind time. `upsert_note` has written `crdt_state` for a long time, so this affects only rows predating that, and the DB is wipeable pre-launch. Flagging it rather than burying it.

https://claude.ai/code/session_01N62e5eq31DoffeCH2LJDFM


Rasbandit:
## ⛔ Converted to draft — data-loss bug found in review

Do not undraft until the two items below are done.

### The bug

Removing `seed_from_content` means a note with NULL `crdt_state` binds to an **empty** doc. On room close, `unbind/3` (`lib/engram/notes/crdt_persistence.ex:239`) calls `CrdtCheckpoint.checkpoint/4` **unconditionally**, which materializes that emptiness into `notes.content` and fires `announce_ready` — telling every other device to pull the emptied note.

The async worker already guards against exactly this (`lib/engram/workers/checkpoint_note.ex:96`):

> "A note opened but never edited has no snapshot and no tail, so the doc rebuilds EMPTY; checkpointing an empty doc would blank notes.content."

The live path has no such guard. That asymmetry was harmless **only** because bind-time seeding guaranteed a non-empty doc — which is what this PR deletes.

### The population is larger than this PR's description claims

`priv/repo/migrations/20260706210000_clear_crdt_state_id_keying_cutover_migrate.exs:37` ran `UPDATE notes SET crdt_state_ciphertext = NULL, crdt_state_nonce = NULL` for **every note**, and its header states the contract:

> "clear it pre-launch and let every note re-seed cleanly from `notes.content` on next bind."

That re-seed is the code removed here. So the affected set is not "rows predating `upsert_note`" — it is every note not opened and not written since 2026-07-06.

### Required before undrafting

1. **Guard the live checkpoint path** — mirror the worker's `has_state?` check so checkpointing a doc with no snapshot and no tail is a no-op.
2. **Backfill** — a one-shot `phase/migrate-data` migration populating `crdt_state` for `crdt_state_ciphertext IS NULL` rows via `CrdtBridge.merge_plaintext(nil, content)`. Without it those notes still bind empty and read blank in the web app until someone writes them.

Note this removes the "wipe `crdt_state`, let bind re-seed" tool that has been used twice. Any future codec cutover now needs a paired backfill.

### Also fix in this PR (one-liners, files already touched)

- `crdt_persistence.ex:55-57` — comment still references "the seed below"; there is no seed below.
- `crdt_persistence.ex:320-329` — orphaned doc comments for the deleted `seed_from_content/3` and `markdown?/1`.
- `test/engram/notes/crdt_persistence_test.exs:126` — `=~` weakened from `==`; passes on `"hello worldhello world"`, and full-body doubling is a previously-shipped failure mode of this exact merge path.

### Correct the claim

The commit subject "the server stops authoring note content" overstates it. The server still authors content at `crdt_deliver.ex:200` and `notes.ex:1383`. The comment at `crdt_deliver.ex:151-155` states unconditionally a premise the fallback 45 lines below exists because it is not unconditional.

### Confirmed clean

The reverted room-per-write / `CheckpointNote`-starvation hazard is **not** reintroduced — this diff opens no rooms.

Rasbandit:
## ✅ Blockers resolved — undrafting

Both required items from the review are done, plus the three one-liners.

### 1. Guard: `CrdtCheckpoint.ensure_projection_safe/2`

A row with no persisted CRDT state whose doc projects nothing is now skipped instead of materialized. Mirrors the `has_state?` gate `CheckpointNote` already applies off the live path; the `.canvas` branch never projects text at all, so markdown-on-unbind was the one path missing it.

A genuine "user deleted all the text" is unaffected — those deletion ops are themselves persisted state, so `crdt_state_ciphertext` is non-nil and real emptying still materializes.

RED-verified before the fix: `assert reloaded.content == "IMPORTANT"` failed with `left: ""`.

### 2. Backfill: `Engram.Workers.BackfillCrdtState`

**The guard alone was not sufficient**, which I verified rather than assumed. With only the guard, the note still binds EMPTY, so it opens blank — and the first character typed gives the doc real state, at which point the checkpoint *legitimately* materializes that one character over the whole body. Probed directly: content `"IMPORTANT"` became `"x"`.

So the data needed repairing, not just defending. The worker seeds `crdt_state` from `content` for NULL-state rows, cursor-batched per (user, vault), mirroring `BackfillCrdtHead`.

- **Representation only** — never writes `content`, `content_hash`, `version`, or `seq` (pinned by a test).
- **Idempotent** — the `is_nil` predicate is re-asserted in the UPDATE, so a note that gains state mid-run (concurrent write, room checkpoint) is never overwritten. This matters: re-seeding a live note would discard its CRDT history.
- **Never raises per-note** — one unseedable note cannot strand the rest of the batch.

Enqueue post-deploy:

```
bin/engram rpc 'Engram.Workers.BackfillCrdtState.enqueue_all()'
```

Seven tests, including an end-to-end one asserting the same `bind/3` yields `""` before the backfill and the real body after — the guarantee `seed_from_content` used to provide, now provided by the data.

### 3. Review one-liners

- Comment rot in `bind/3` referencing a seed that no longer exists — rewritten.
- Orphaned doc comments for the two deleted functions — removed.
- `=~` restored to `==` (it passed on `"hello worldhello world"`, and full-body doubling is a previously-shipped failure mode of this merge path). It passes as `==`, which also confirms no doubling.

### Not done here

The commit subject's "the server stops authoring note content" **does** overstate it — `crdt_deliver.ex:200` and `notes.ex:1383` remain. Left as-is deliberately: `notes.ex:1383` (`maybe_merge_crdt`) moved from bind time to write time, which is the correct design, and removing the `crdt_deliver.ex:200` fallback is a separate change. Flagging rather than silently expanding this PR.

### Verification

3805 tests / 0 failures, `credo --strict` clean, `dialyzer` clean. Dialyzer caught a real bug in the new worker (wrong module path for the logging metadata helper) before it shipped.

Issue #1148, opened 2026-07-31
p1: implement CIMD (Client ID Metadata Documents) — provable identity for loopback MCP clients, and get off the deprecated DCR path


Rasbandit:
## Why

Two separate problems, one fix.

**1. Local-first clients can never be verified today.** Claude Code, Cline, OpenCode, Cursor and Windsurf all redirect to loopback. Verification currently requires a vendor-owned HTTPS redirect host, which is un-spoofable because the auth code gets delivered to the vendor rather than the forger. Loopback has no such host, so those clients are unverifiable *by construction*, not by misconfiguration. PR #1147 got them attributed (brand mark + slug + onboarding row ticks) via the self-reported `client_name`, but attribution is not proof, and the connections page says so in as many words.

**CIMD fixes exactly this.** The `client_id` becomes an HTTPS URL the vendor owns, and the AS fetches the metadata document from it. Nobody else can serve a document at `claude.ai`, so the URL host is un-spoofable for the same reason a vendor redirect host is. It gives us a **provable identity that the redirect can never supply**.

**2. We are on the deprecated path.** MCP `2025-11-25` adopted CIMD (IETF `draft-ietf-oauth-client-id-metadata-document`) and demoted DCR from SHOULD to MAY. The July 2026 spec update **deprecates DCR**, keeping it only for servers that have not migrated.

## We keep both, permanently

This is additive, not a migration. Of the nine connectors observed in prod, seven do not use CIMD, and self-hosted clients (Open WebUI, LobeChat) never will — there is no vendor to publish a document. DCR stays as the floor.

The two paths are independent: a CIMD client sends a URL-shaped `client_id` and never touches `/oauth/register`, so the existing flow is untouched.

## Precondition already half-met

Anthropic's docs state Claude selects CIMD only when AS metadata advertises **both**:

- `token_endpoint_auth_methods_supported` containing `"none"` — ✅ already do (`well_known_controller.ex`)
- `client_id_metadata_document_supported: true` — ❌ missing

With the flag absent, Claude Code falls back to DCR, which is precisely why the observed grant is an anonymous loopback client. Claude Desktop / Web / Cowork appear to use CIMD via shared infra.

No MCP client sends an RFC 7591 `software_statement` (signed JWT), so that is not an alternative route to cryptographic vendor attribution today.

## Scope

- [ ] Advertise `client_id_metadata_document_supported: true` (behind config, see rollout)
[...]

PR #1163, opened 2026-07-31
fix(auth): stop a self-asserted software_id granting a vendor logo

### LINKED ISSUE #1156: p2: a self-asserted software_id still grants a vendor logo (delete the 4 unvalidated entries)
## Problem

```json
```

## The actual fix

```elixir
```

## Scope

## Why not in #1147

Rasbandit:
Closes #1156.

## The problem

#1147 stopped it granting `verified: true` and made a proven redirect host outrank it for identity — but it still supplied `logo`, `display_name` and `slug`. So a rogue client registering:

```json
```

still appeared in the victim's `/settings/connections` as **Claude Desktop**, with Anthropic's logo, and with **no `unverified` chip** — the chip is suppressed whenever a slug resolves, deliberately, so that Claude Code isn't badged as suspect.

Not an authorization-time phishing vector (the consent screen shows only `client_id` + `client_name`). The risk is that a rogue grant looks trustworthy enough not to revoke.

## The fix

Four of the five `@software_id` entries were **never observed on a real registration** and were documented in the moduledoc as guesses. Deleted:

- `anthropic-claude-desktop`
- `cursor.sh` — Cursor actually registers `cursor://` with no `software_id`
- `openai-chatgpt` — ChatGPT is attributed by host
- `vscode-engram`

All nine connectors observed in prod are attributed by **redirect host** or **client_name**; none sends `software_id`. The map was carrying attribution for nobody except our own plugin, while remaining a free vendor-logo grant for anyone who read the source.

`engram-vault-sync` **stays**: it redirects to a custom scheme, so no host can attribute it, and its `client_name` derives no catalog slug. It has nothing else to identify it.

A self-asserted `software_id` now resolves to the unverified placeholder, so the rogue grant renders with the "Unverified client" chip.

## Test changes, and why they aren't loosening

Two existing tests asserted vendor identity *from a guessed id*, so they had to change. Both got **stronger**:

- `a self-asserted software_id cannot buy the verified badge` → `a self-asserted vendor software_id buys no identity at all`, now asserting `display_name: nil, logo: nil, slug: nil` rather than just `verified: false`.
- Added `the deleted guessed software_ids resolve to the placeholder`, naming all three others individually, so **re-adding any key fails the suite** rather than silently restoring the logo grant.

One needed a subtler fix. `a verified host wins over a conflicting self-asserted software_id` used `"openai-chatgpt"` — with that key deleted the lookup returns empty, so the assertion would still pass while testing *nothing about precedence*. Switched to `engram-vault-sync`, the only id that still resolves, so it actually exercises host-over-software_id again.

`connections_controller_test` asserted `slug == "claude"` off the deleted key. Rather than pick another magic string, the fixture now models a **real** loopback client — `client_name: "Claude Code (engram)"`, no `software_id` — which also exercises `normalize_name/1`'s parenthetical strip end-to-end. That is the only attribution path such clients have left.

## Seeds

The two demo Claude clients redirect to `claude.ai`, so the host already attributes them (its `display_name` already overrode the `"Demo: "` name) — the `software_id` was redundant and is dropped.

The demo Cursor client keeps its `"Demo: "` prefix and now renders unattributed, which is accurate: real Cursor is loopback with no `software_id`. **I nearly renamed it to `"Cursor"` so its slug would derive** — that would have broken the seed cleanup, which deletes by `like(client_name, "Demo: %")` and would have orphaned the row on every reseed. Left as-is with a comment recording the constraint.

## Not changed

`e2e/tests/api_only/test_71_connections.py` still registers these ids. It uses `software_id` as opaque registration metadata and asserts only on revoke behaviour and 402 limits — nothing about identity — so it is unaffected. Arguably more realistic now: it *is* an unknown client.

## Gate

| check | result |
|---|---|
| `mix format --check-formatted` | pass |
| `mix credo --strict` | pass |
| pre-push hook (format, compile --warnings-as-errors, credo, sobelow, …) | pass |
| `mix test` (connections + controller) | **55 tests, 0 failures** |

Full suite runs in CI.

https://claude.ai/code/session_01CvUvadaneVBQXax8NfeNJU


Rasbandit:
**Heads-up: this duplicates part of #1167.**

Flagging rather than acting, since #1167 isn't mine to close and the ordering is your call.

#1167 (`feat/oauth-cimd`, opened ~an hour after I started this) deletes **the same four `@software_id` entries**, with the same reasoning — we independently reached the same conclusion from the same issue text. It also rewrites the same `logo_allowlist.ex` moduledoc and the same `docs/context/connections-client-identity.md` section, so these two will conflict textually, not just semantically.

### Options

- **Close this, keep #1167.** #1167 does strictly more — it closes #1156 *and* #1148 (CIMD), and its version of the moduledoc already integrates the deletion into the four-source precedence story rather than bolting it on.
- **Merge this first, rebase #1167.** This PR is the smaller, surgical change with its own revert point, which is what #1156 explicitly asked for (*"it is a behaviour change ... so it deserves its own change and its own revert point"*). #1167 then drops the duplicated hunk on rebase.

I'd lean the **first** — the duplicate revert point isn't worth a hand-resolved conflict in a security-sensitive file, and #1167's moduledoc is the better end state.

### One thing worth salvaging either way

The test changes here are not in #1167 and are the part I'd be sorry to lose:

- `the deleted guessed software_ids resolve to the placeholder` — names all four ids individually, so **re-adding any one fails the suite** rather than silently restoring the logo grant.
- The precedence test was switched from `"openai-chatgpt"` to `"engram-vault-sync"`. This matters: with the guessed key deleted, that assertion still *passes* while testing **nothing at all** about precedence, because the lookup returns empty either way. Worth porting to #1167 if this one is closed.

### Also checked

#1167 touches two other files this session's PRs touch, but in different regions, so they should merge cleanly:

- `frontend/src/api/queries.ts` (adds a `cimd_url` field) vs #1169's `RawFolder` work
- `frontend/src/onboarding/checklist-widget.test.tsx` (adds `cimd_url: null` to fixtures) vs #1162's catalog-parity test

PR #1167, opened 2026-07-31
fix(oauth): stop granting vendor identity on unprovable claims (CIMD + software_id deletion)

### LINKED ISSUE #1148: p1: implement CIMD (Client ID Metadata Documents) — provable identity for loopback MCP clients, and get off the deprecated DCR path

### LINKED ISSUE #1156: p2: a self-asserted software_id still grants a vendor logo (delete the 4 unvalidated entries)

Rasbandit:
Implements **CIMD** (#1148): a provable identity for loopback MCP clients, and the way off the deprecated DCR path.

| | Before | After |
|---|---|---|
| Claude Code (loopback redirect) | Unverifiable **by construction** | Verified, if it publishes a CIMD document |

**CIMD goes live on deploy — there is no feature flag.** See [Rollout](#rollout).

> **Scope changed mid-flight.** This started as #1148 + #1156. While it was in progress, **#1163 shipped the `software_id` deletion and closed #1156** — independently derived and substantively identical, down to the same observation that the precedence test would otherwise pass trivially on an empty lookup. `origin/main` is merged in here and **main's version of that work won** on every overlapping file; my duplicate commit survives only in history, not in the tree. So this PR is now CIMD-only. Section 1 below is kept for context on *why* CIMD matters, since the two fixes attack the same root problem — identity granted on claims rather than proof.

---

## 1. Context: `software_id` no longer names any vendor (#1156, shipped separately in #1163)

`software_id` is an RFC 7591 field the client sends **about itself** in the DCR body. PR #1147 stopped it granting `verified: true`, but it still supplied `logo`, `display_name` and `slug`, so:

```json
{ "software_id": "anthropic-claude-desktop",
  "redirect_uris": ["http://127.0.0.1:9999/cb"] }
```

appeared in the victim's `/settings/connections` as **Claude Desktop**, wearing Anthropic's logo, with **no `unverified` chip** — the chip is deliberately suppressed whenever a slug resolves, so that Claude Code is not badged as suspect.

Not an authorization-time phishing vector (the consent screen shows only `client_id` + `client_name`), but it made a rogue grant look trustworthy enough not to revoke.

Four of the five entries had never been observed on a real registration and were documented in the moduledoc as guesses. All nine connectors observed in prod omit `software_id` entirely. So the map carried attribution for **nobody** except our own plugin, while remaining a free vendor-logo grant to anyone who read the source. `engram-vault-sync` stays — it is ours, and it redirects to a custom scheme, so neither the host layer nor name derivation can attribute it.

The test-fixture reasoning below was mine; **main's equivalents are what is in the tree now**, and where they differ main's is the better call — its `connections_controller` fixture models a loopback Claude Code rather than a vendor-host Claude, and name-derivation is the only attribution path a local-first client has left, so it is the path most worth pinning.

<details><summary>My original notes on those fixtures, kept for the reasoning</summary>

- The precedence test ("a verified host wins over a conflicting self-asserted `software_id`") claimed `openai-chatgpt`. With that entry deleted it would have passed *trivially* and stopped proving precedence, so it now claims `engram-vault-sync`, the surviving real entry.
- `connections_controller_test` asserted `slug == "claude"` via the deleted `anthropic-claude-desktop`, and it is the **only** test proving slug reaches the JSON. Rather than let that coverage lapse, its fixture is now shaped like the real Claude registration observed in prod: no `software_id`, attributed and verified by its `claude.ai` redirect host.
- The demo seeds claimed the deleted ids. The two Claude rows already carried the vendor host that now attributes them; the Cursor row is loopback, so it is renamed to the product name Cursor actually registers, deriving its slug by name exactly as prod does.

</details>

## 2. CIMD (#1148)

Verification is granted by exactly one thing: proof that the party we are talking to controls a host the vendor owns. A vendor-owned HTTPS **redirect** host proves that — a forger can *claim* `redirect_uri=https://claude.ai/...`, but the auth code is then delivered to Anthropic, not to them.

Loopback clients have no such host. Claude Code, Cline, Cursor, OpenCode and Windsurf are therefore unverifiable **by construction**, not by misconfiguration. #1147 got them attributed via `client_name`, but attribution is not proof.

CIMD supplies the proof the redirect never can: the `client_id` becomes an HTTPS URL the vendor owns, we fetch the metadata document from it, and **nobody but Anthropic can serve a document at `claude.ai`**. It also gets us off a deprecated path — MCP `2025-11-25` demoted DCR from SHOULD to MAY, and the July 2026 update deprecates it.

**DCR stays, permanently.** Seven of nine observed connectors do not use CIMD, and self-hosted clients (Open WebUI, LobeChat) never can — there is no vendor to publish a document. The paths do not interact: a CIMD client sends a URL-shaped `client_id` and never touches `/oauth/register`.

### The document is bound, not trusted

One check carries the entire security argument: **the document's own `client_id` MUST equal the URL it was served from.** Without it, a vendor could serve a document claiming any `client_id`. Everything else the document says is then attributable to whoever controls that host, which is the point.

Also refused: no usable `redirect_uris`, and any confidential `token_endpoint_auth_method`. That last one is refused rather than downgraded on purpose — a CIMD client never registered, so no secret exists. Honouring it leaves the client unable to authenticate (nil hash); silently downgrading to `none` makes it send a secret that `authenticate_client/2` must then reject *for being present at all*. Both failures are opaque; refusing the document is legible.

Document metadata is validated by `registration_changeset/2` rather than a parallel CIMD path, so the redirect-URI rules hardened in #1147 (the `https:///cb` host-less trap, array bounds, unsafe schemes) apply verbatim and cannot drift. `software_id` from the document is deliberately dropped: it would be attributable, but after #1156 it names only our own plugin, and storing it would re-grow the surface that issue just removed.

### The `client_id` type seam

The PK is **not** widened. `structure.sql` confirms `oauth_authorization_codes.client_id` and `oauth_refresh_tokens.client_id` are bare `uuid` columns with no FK, so a unique partial-indexed `cimd_url` column carries the wire identity and UUIDs stay internal.

What that leaves is three places where a bare `==` against a stored UUID would reject the **legitimate** client presenting its URL: `check_code_client/2`, refresh rotation, and revocation. All three normalize through `internal_client_id/2`, which short-circuits when the wire id already matches, so the DCR path pays for no extra query.

`get_client/1` stays a pure DB lookup. **Only `/oauth/authorize` may fetch a document** — if token exchange or revocation could, a vendor outage would break already-granted access.

The round-trip is tested end to end (mint → exchange → rotate → revoke, all with the URL as the wire `client_id`), plus the negative: another client's URL cannot redeem this client's code.

### SSRF: most of the work

There was **no SSRF guard anywhere in `lib/`** before this. `/oauth/authorize` is unauthenticated, so this is an unauthenticated-request-triggered outbound fetch — an SSRF primitive *and* a traffic amplifier aimed at third parties.

`Engram.Http.SsrfGuard` enforces: https only, port 443 only, no userinfo (`https://claude.ai@169.254.169.254/` reads as claude.ai — the ambiguity is the bug), no fragment, and public addresses only against the IANA special-purpose ranges. That includes CGNAT `100.64/10`, link-local `169.254/16`, IPv6 ULA/link-local, and the v4-embedding transition ranges 6to4 `2002::/16` and Teredo `2001::/32`, which a relay forwards to an internal v4 host. IPv4-mapped IPv6 is unwrapped and judged as its inner v4 address — `::ffff:169.254.169.254` is the metadata service in a v6 costume. If **any** address for a name is non-public the whole name is refused, which covers a split-horizon resolver returning one public and one internal answer.

**Pinning is the subtle part.** `resolve/1` returns a URL with the host replaced by the address it validated; the caller connects to *that* while passing the original hostname for SNI and certificate verification (`connect_options: [hostname: ...]`). Handing the *hostname* to the HTTP client would re-resolve it, and the second answer is free to be `127.0.0.1` — DNS rebinding, where the check and the connection disagree about what they are talking to.

**Redirects are not followed.** A redirect is a new URL that would bypass the guard that approved the first one. A vendor that redirects its metadata document does not work, and that is correct. Body is capped **mid-stream** at 64 KB — checking `byte_size` afterwards means already holding a hostile vendor's whole body in memory.

Two rate-limit buckets, per-host (10/min) and global (60/min). Per-host alone is useless against an attacker varying the host, which is exactly the amplification case.

### Caching, and a vendor being down

The `oauth_clients` row **is** the document cache, with `cimd_fetched_at` as its clock (24h TTL). No ETS cache: the redirect allowlist derives from the document and has to be durable anyway, so a second copy would only add a way for the two to disagree.

A failed refresh keeps serving the stale row — **including when the new document fails validation.** A vendor's five-minute outage, or a vendor that starts serving something broken, must not lock out every user of that client. A first-contact race is resolved by re-reading the winner's row rather than `on_conflict:`, because ON CONFLICT cannot infer a *partial* unique index from a bare column list.

### The connections page

`LogoAllowlist.resolve/4` takes the CIMD host as a fourth, top-precedence identity source, and grants `verified: true` **without needing to recognise the host** — fetching the document already established who serves it. The allowlist is consulted only to dress the row; an unrecognised vendor shows its host and takes its checklist slug from `client_name`, the same mechanism loopback clients already use.

The Identity row had one "Verified" string: *"Sign-in redirects to a domain the vendor owns."* For a CIMD client that is **plainly false** — Claude Code redirects to localhost. So the copy branches, and the two strings stay distinct on purpose: if they ever collapse into one, one of them is a lie. A test pins each branch. The Identifier row now shows the CIMD URL, which unlike an opaque UUID the user can check by visiting it.

## Rollout

**There is no feature flag, deliberately.** An earlier revision of this PR gated CIMD on `ENGRAM_CIMD_ENABLED` (default off) so the capability could be advertised staging-first. That was removed before merge, for two reasons:

- It was config for a value that never changes — a knob set to `true` once and never touched again.
- A default-off flag nobody remembers to set is the *worse* failure mode. CIMD would look shipped while silently doing nothing, and no tripwire would fire, because no CIMD traffic would ever arrive to trip one.

**So CIMD is live the moment this deploys.** Claude Code stops choosing DCR as soon as it sees `client_id_metadata_document_supported`, and it does **not** silently fall back, so if `Engram.OAuth.Cimd` is broken then *new* Claude Code connections fail rather than degrade. Existing DCR grants are separate rows and are unaffected. Backing this out is a revert of the advertisement line plus a deploy, not a config change.

Worth stating the residual risk plainly: no test can cover the real Claude Code handshake, so the first genuine CIMD exchange happens in prod. Everything up to that boundary is covered — the document binding, the TTL and stale-retention behaviour, the transport's caps and refusals, the SSRF ranges, and the full wire round-trip through mint → exchange → rotate → revoke.

**Tripwires, which matter more now that there is no flag to blame:** `mcp_cimd_rejected` (a document was refused; reason + host, host only because `:lifecycle` ships to Loki) and `mcp_cimd_stale_retained` (a refresh failed, serving yesterday's document). Rate-limit denials are deliberately *not* logged — the fetch is bounded but the log line would not be, so that volume is a metric (`purpose: :cimd_fetch, result: :deny`) instead.

## Testing

`mix test` (full suite), `mix format --check-formatted`, `mix credo --strict`, `mix sobelow --exit`, `mix dialyzer`; frontend `vitest` (1142 tests), `tsc --noEmit`, `biome ci`.

The transport sits behind `Cimd.Fetcher` — the same behaviour + Mox shape as `Engram.Paddle.Client` and `Engram.Storage` — so policy is tested without a socket while the transport is tested against a **real local server**: body cap enforced mid-stream, non-JSON refused, non-200 reported, and a 302 to `169.254.169.254` **not** followed. The seam carries transport only; which URLs may be fetched stays in `SsrfGuard`, so a mocked fetcher cannot quietly widen what production requests.

Three things the tests caught that review would probably not have:

- **IPv6 double-bracketing.** `URI.to_string/1` brackets a host containing `:` itself, so bracketing the literal produced `https://[[::1]]/`.
- **The rate limiter, by breaking the first run.** Exactly ten fetches passed and everything after failed `:rate_limited` — the per-host budget working. Both buckets are now asserted explicitly.
- **Nine test fixtures missing the new required `cimd_url` field**, via `tsc`. `vitest` alone passes, since it does not typecheck.

Boundary tests that earn their keep: `172.16/12` (a /12 implemented as a /16 comparison wrongly rejects `172.32`; a byte-prefix one wrongly accepts `172.15`) and `100.64/10`.

## Docs

`docs/context/connections-client-identity.md` — precedence table now four sources, trust model now two proofs, the CIMD section rewritten from "not yet implemented" to how it is built, the UI states table from three states to four, and a note recording why there is no flag (so nobody re-adds one without reading the argument). No new env var.

Closes #1148

(#1156 was closed by #1163; this PR carries its work only as a merge of `origin/main`.)


Rasbandit:
### Code review

Five independent passes: guidance-file compliance, bug scan, git history vs the invariants #1147 established, prior review feedback on these files, and code-comment fidelity.

Three findings, all fixed in this PR rather than deferred:

1. **One global fetch budget let discovery starve refresh.** Refresh of an already-known CIMD client shared the 60/min global bucket with discovery, which any unauthenticated `/oauth/authorize` can drive. An attacker cycling hosts could consume it and every legitimate refresh would fall back to its retained stale document — the anti-amplification control turned into a DoS vector against the vendors real users are connected to. Refresh now has its own per-host bucket, unreachable without a row that already passed discovery. Pinned by a test that burns the discovery budget and asserts a known-stale client still refreshes.

https://github.com/engram-app/Engram/blob/f9fb0e2e19cbd19d7b9a6c256dca801c1033a4b8/lib/engram/oauth/cimd.ex#L200-L212

2. **`LogoAllowlist.resolve/4`'s @doc was stale.** It still said identity comes from "the redirect host, then `software_id`, then the normalized `client_name`" and that `verified` is "decided by exactly one thing". The moduledoc and `docs/context/connections-client-identity.md` were updated to the four-source / two-proof model; the docstring immediately above the function was not. That is the failure mode the context doc names as the reason an earlier gap survived six weeks.

https://github.com/engram-app/Engram/blob/f9fb0e2e19cbd19d7b9a6c256dca801c1033a4b8/lib/engram/connections/logo_allowlist.ex#L115-L132

3. **The connections page branched on `cimd_url` truthiness, not `verified`.** Equivalent in every reachable path today, but it restated the backend's verification rule in TypeScript with nothing keeping them in sync, so loosening `SsrfGuard` or `lookup_by_cimd` could have the UI assert a proof the server never granted. `verified` now gates both "Verified." strings and `cimd_url` only selects which proof to name; a test pins the unreachable-today drift case.

https://github.com/engram-app/Engram/blob/f9fb0e2e19cbd19d7b9a6c256dca801c1033a4b8/frontend/src/settings/connections-page.tsx#L166-L181

Also checked and clean: the SSRF guard's CIDR boundary math and reserved-range tables, address pinning against DNS rebinding, the `internal_client_id` normalization at token exchange / refresh rotation / revocation, the document-binding check, the partial-unique-index race handling, and the three invariants from #1147 (redirect-URI validation, host-outranks-`software_id` precedence, the RFC 8252 loopback port exemption) — all preserved, with the CIMD path reusing `registration_changeset/2` rather than introducing a parallel validator.

Rasbandit:
Heads-up from an overnight autonomous pass: **#1163 has merged**, which deletes the four guessed `@software_id` entries this PR also deletes. You'll hit conflicts on update-branch.

I measured the surface with `git merge-tree` rather than guessing — **7 files conflict**:

```
lib/engram/connections/logo_allowlist.ex
test/engram/connections/logo_allowlist_test.exs
test/engram/connections_test.exs
test/engram_web/controllers/connections_controller_test.exs
priv/repo/seeds_connections.exs
docs/context/connections-client-identity.md
frontend/src/onboarding/checklist-widget.test.tsx
```

### What's already on main, so you can drop it

- The four `@software_id` deletions themselves (`anthropic-claude-desktop`, `cursor.sh`, `openai-chatgpt`, `vscode-engram`). **Take main's side** — the outcome is identical.
- The seeds change: the two demo Claude clients lost their redundant `software_id` (they're host-attributed anyway). The demo Cursor client deliberately **keeps** its `"Demo: "` prefix — renaming it to `"Cursor"` so its slug derives would break the seed cleanup, which deletes by `like(client_name, "Demo: %")` and would orphan the row on every reseed. There's a comment on it.
- `checklist-widget.test.tsx` also conflicts with **#1162** (merged), which added a catalog-parity test. Yours adds `cimd_url: null` to fixtures — both wanted, different lines.

### Worth keeping from your side

Your moduledoc is the better end state: it folds the deletion into the four-source precedence story (CIMD host → redirect host → software_id → client_name) instead of describing it as a standalone tightening. **Take yours** for `logo_allowlist.ex` prose and `connections-client-identity.md`, then re-check that the `@software_id` map itself matches main.

### Please don't lose these two test changes

They're on main now and are the part most easily clobbered by taking your side wholesale:

1. `the deleted guessed software_ids resolve to the placeholder` — names all four ids individually, so **re-adding any one fails the suite** rather than silently restoring the vendor-logo grant.
2. The precedence test moved from `"openai-chatgpt"` to `"engram-vault-sync"`. This one matters: with the guessed key deleted, the old assertion still **passes while testing nothing** — the lookup returns empty either way, so it no longer exercises host-over-software_id at all.

Sorry for the collision — we were working the same issue (#1156) in parallel without knowing.

Issue #1158, opened 2026-07-31
flake: CrdtTransportTest setup times out on DBConnection checkout at max_cases: 20


Rasbandit:
## Symptom

`Engram.Notes.CrdtTransportTest` intermittently fails in its **setup block**, before any assertion runs:

```
** (DBConnection.ConnectionError) connection not available and request was dropped from queue after 15000ms
```

Every test in the module fails together, which is the signature of a setup failure rather than a logic bug.

## Diagnosis

Pool contention, not a defect in the module. Observed 2026-07-30 during a full `mix test` at the default `max_cases: 20`. The module passes 22/22 when run in isolation.

The Ecto sandbox pool is shared across concurrent async cases; when enough of them hold checkouts, a setup that needs a connection waits out the 15s queue timeout. Nothing about this module is special except that its setup is connection-hungry at exactly the wrong moment.

Related but distinct: a genuinely different flake in the same full-suite run, `Engram.Billing.OverrideCacheTest`, turned out to be a real cross-test race (an undrained PubSub `evict_all` broadcast landing in the next test). That one was **fixed** in PR #1147. This one was not, because it is a resource-shape problem rather than a correctness bug.

## Options

1. Raise the checkout timeout for this module only (`@moduletag`), the smallest change, but it hides rather than removes the queueing.
2. Move the connection-hungry work out of `setup` so the checkout happens inside the test with the normal ownership path.
3. Reduce what the setup does, if it is establishing more state than the tests need.
4. Leave it and accept the occasional red, which conflicts with the never-merge-on-red rule.

(2) is the real fix if the setup is doing avoidable work.

## Notes

- Not reproducible on demand; needs a loaded machine or a full-suite run.
- Do **not** confuse with a rerun-pass being proof of health. Isolation passing is evidence of the *cause* (contention), not evidence that it is harmless.
- If CI starts hitting this, it will look like an unrelated CRDT failure on an innocent PR.

Rasbandit:
## Second instance, same class: `Engram.Cluster.ReadinessTest`

Hit twice on 2026-07-31 during full-suite runs on a loaded dev box; passes 3/3 in isolation.

```
test "resolve_a/2 ... fails open within a bounded budget against an unresponsive resolver"
assert elapsed_us < 4_000_000
left: 5_500_356   (heavy contention)
left: 4_076_698   (light contention, only 2% over)
```

### Why it is brittle rather than merely unlucky

[...]

PR #1176, opened 2026-07-31
fix(connections): delete the guessed software_id entries, link Antigravity docs

### LINKED ISSUE #1156: p2: a self-asserted software_id still grants a vendor logo (delete the 4 unvalidated entries)

### LINKED ISSUE #1157: p3: Antigravity checklist row has no docs link (DOC_URLS entry missing, comment is stale)
## Problem

`frontend/src/onboarding/checklist-widget.tsx` has a label for `antigravity` but no `DOC_URLS` entry, so the onboarding checklist row renders without its "Learn how" link. Every other connector has one.

The comment explaining the omission is now false:

```ts
// No `antigravity` entry yet, engram-marketing has no integrations/antigravity/
```

The page exists and is live:

```
$ curl -sL -o /dev/null -w '%{http_code}' https://engram.page/docs/integrations/antigravity/
200
```

`src/content/docs/docs/integrations/antigravity.mdx` is in engram-marketing. The comment was accurate when written during #1147 and went stale before that PR merged.

## Fix

```ts
antigravity: "https://engram.page/docs/integrations/antigravity/",
```

and delete the comment.

## While in there

Worth an eyeball on the QC gallery (`/__qc/connectors`, dev only) for two suspected brand-wordmark mismatches carried over from `@lobehub/icons-static-svg`:

- `chatgpt` wordmark may render "OpenAI" rather than "ChatGPT"
- `lobechat` wordmark may render "LobeHub" rather than "LobeChat"

Unverified: the assets are vector paths, so it needs a visual check rather than a grep. Ignore if they render correctly.

Rasbandit:
Two follow-ups from the #1147 review. Closes #1156, closes #1157.

## 1. A self-asserted `software_id` no longer buys vendor branding (#1156)

`@software_id` carried four vendor guesses — `anthropic-claude-desktop`, `cursor.sh`, `openai-chatgpt`, `vscode-engram` — that had **never** been seen on a real registration. All nine connectors observed in prod are attributed by redirect host or by `client_name`, and not one sends `software_id` at all.

They were previously documented as "harmless". They were not.

`software_id` arrives in the DCR request body, so it is self-asserted. Each entry was a standing offer: register claiming `anthropic-claude-desktop` and the connections list hands you Claude's **logo and display name**, with no `unverified` chip — because the chip is deliberately suppressed whenever a slug resolves, so that Claude Code isn't badged as suspect. So they attributed nobody real, while giving anyone who read the source a free vendor identity.

#1147 stopped `software_id` granting the `verified` badge. This closes the rest of it.

One entry remains: our own `engram-vault-sync`, which genuinely needs it — the plugin redirects to a custom scheme, so no host identifies it and name derivation doesn't cover it.

A test pins all four deleted ids as conferring **nothing**: no badge, no logo, no name, no slug. If vendor branding comes back, someone has re-added a speculative entry.

### Test fixture change worth reviewing

`connections_controller_test` used `software_id: "anthropic-claude-desktop"` to assert the endpoint serializes `slug`. That fixture depended on a deleted guess, so I re-pointed it at a **real observed shape** (no `software_id`, loopback redirect, `client_name: "Claude Code (engram)"`). It proves the same thing — the endpoint serializes `slug` — without asserting a mapping that no longer exists.

## 2. Antigravity's checklist row had no docs link (#1157)

Adds the `DOC_URLS` entry and deletes the comment claiming the marketing page doesn't exist. It does, and returns 200 — the comment went stale during #1147.

**The entry is the trivial part; the guard is the point.** `DOC_URLS[slug] ?? DOC_FALLBACK` renders the generic docs index for an unmapped slug, so a missing entry fails *invisibly*. That is precisely why nobody noticed.

Added a test asserting every selectable catalog connector resolves to its own URL rather than the fallback, and verified it catches the defect by reverting the entry:

```
AssertionError: Antigravity (antigravity) has no DOC_URLS entry,
so its row falls back to the docs index
```

`unavailable` entries are excluded — Gemini is listed to explain that it *can't* connect, and deliberately has no guide.

## Verification

- Backend: **3852 tests, 0 failures**; credo `--strict` + format clean
- Frontend: `tsc --noEmit` + `biome ci` clean; 186 tests across onboarding + settings
- Guard proven to fail without the fix (above)

Context doc updated: the section that called the entries "harmless" now records why that was wrong, and carries a do-not-re-add rule.

### DIFF
--- a/docs/context/connections-client-identity.md
+++ b/docs/context/connections-client-identity.md
@@ -233,9 +233,15 @@ SELECT software_id, client_name, redirect_uris FROM oauth_clients;
 
 `@lobehub/icons-static-svg` is already a dependency and already covers every catalog slug. Use `ToolMark slug={...}` (`frontend/src/onboarding/tool-icon.tsx`). The backend `logo: "/assets/clients/*.svg"` field is a legacy parallel system still needed only for `engram-vault-sync` and `vscode`; `grok.svg`/`mistral.svg` were never created and don't need to be.
 
-## Known stale: the 4 guessed `software_id` entries
+## The 4 guessed `software_id` entries were DELETED (2026-07-31, #1156)
 
-`anthropic-claude-desktop`, `cursor.sh`, `openai-chatgpt`, `vscode-engram` are UNVALIDATED guesses. Prod data now **proves** they never fire (the real ChatGPT and Claude grants both arrive with `software_id: null`). Harmless, but they read as coverage, do not add more speculative entries. The only proven-real `software_id` is our own `engram-vault-sync`.
+`anthropic-claude-desktop`, `cursor.sh`, `openai-chatgpt` and `vscode-engram` were unvalidated guesses. Prod proved they never fire: all nine observed connectors are attributed by redirect host or by `client_name`, and **not one sends `software_id` at all** (the real ChatGPT and Claude grants both arrive with `software_id: null`).
+
+This section used to call them "harmless". They were not. Because `software_id` is self-asserted, each entry was a standing offer: register claiming `anthropic-claude-desktop` and the connections list hands you Claude's logo and display name, with no `unverified` chip (the chip is suppressed whenever a slug resolves, so Claude Code is not badged as suspect). They attributed nobody real, while giving anyone who read the source a free vendor identity.
+
+`@software_id` now holds exactly **one** entry, our own `engram-vault-sync`, which genuinely needs it: the plugin redirects to a custom scheme, so no host identifies it and name derivation does not cover it.
+
+> **Do not re-add a vendor entry speculatively.** An entry here is a claim we have agreed to render, not a fact. Add one only after `mcp_dcr_unattributed_client` shows a real client that needs it AND nothing else can attribute it. `logo_allowlist_test.exs` pins all four deleted ids as conferring no branding.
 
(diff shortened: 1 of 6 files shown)


Rasbandit:
Closing as superseded. #1162 and #1163 landed on main while this was in CI and cover both fixes, so there is nothing left here worth merging.

Checked rather than assumed, and their version is **better than mine** in one respect worth recording.

I wrote the DOC_URLS coverage guard as a render loop: mount the widget once per catalog slug and assert the rendered link is not the fallback. #1162 asserts map parity directly and says why:

> Deliberately a map-parity check and NOT a render: rendering all 14 rows costs enough CPU to starve a sibling suite's real-timer chain (it tipped `agreement-page.flow.test.tsx` over its `waitFor` budget under full-suite parallelism). Nothing here needs the DOM.

That is exactly right, and my version would have introduced the failure mode. Rendering 14 widgets to check a lookup table is CPU spent for no coverage, and this repo already has several timing assertions that go red under suite parallelism (#1158) — I hit two of them myself tonight. Same guard, none of the cost.

The `software_id` deletion in #1163 is equivalent to mine, including naming each deleted id in a test so a re-added key fails loudly rather than silently restoring the logo grant.

No action needed. Recording it because the duplication was mine to avoid: I should have re-checked `origin/main` immediately before opening, not just at branch time.

PR #1188, opened 2026-08-01
fix: security & correctness fixes from backend-wide review

Rasbandit:
## Summary

PR A of the review-remediation series (A: security/correctness, B: web-layer DRY, C: core DRY, D: perf). Bundles the context-boundary and correctness findings from the full-backend best-practice review.

- **Attachments**: MIME/extension whitelist enforced inside `upsert_attachment/3` — it was controller-only, so MCP tools, Oban jobs, and console callers bypassed it entirely (and the whitelist is declared-type/extension only, so bypass meant zero validation). Controller gains defense-in-depth 415 clauses. `batch_delete/3` capped at 500 paths.
- **Notes**: batch delete/move/upsert capped at 500 entries at the context boundary. The old comments claimed a controller-side 100 cap that **never existed**; >500 rows stamped on one server timestamp can make the legacy change feed re-serve the same page forever (`changes_server_time` bound). 500 matches that bound; plugin chunks at ≤100, SPA selections realistically stay far under. Controllers map the new error to 422 `batch_too_large`.
- **BackfillContentHashHmac**: decrypt-failure logs now route through `Engram.Logger.DecryptFailure` (the module whose documented invariant this worker violated) and skip-telemetry `reason` is bounded via `Telemetry.error_kind/1`.
- **DeviceFlow**: `authorize_device/3` now rejects soft-deleted vaults (`is_nil(deleted_at)`), matching `OAuth.verify_vault_ownership/2`. TDD test proved the live bug: the old code returned `{:ok, authorized}` against a deleted vault.
- **Onboarding**: `status/1` read `:billing_enabled` twice with opposite defaults (`false` line 164, `true` line 174). Now read once. Currently dead (runtime.exs always sets the key) but a latent self-host/SaaS gating flip.
- **Search**: unknown `:mode` returns `{:error, :invalid_mode}` (was `FunctionClauseError` mid-pipeline); public `clamp_limit/1` bounds MCP/console-supplied limits to 1..100 (controller keeps its own 50).
- **CrdtBridge**: `client_count/1` now uses `CrdtTransport`'s bounded LEB128 reader; deleted the module's own unbounded reimplementation of the exact decode that already needed a DoS cap once.
- **Storage.Database**: `exists?/1` no longer rescues everything to `false` — a dead DB raises loudly instead of reading as "object missing".
- **CrdtChannel**: the one unlogged `rescue` (checkpoint materialize on create) now logs a warning; still best-effort.
- **CleanupVault**: documented why an Oban `unique` key is deliberately absent (restore→re-delete needs a second scheduled job; uniqueness over `:scheduled` would silently drop it and lose the cleanup — duplicate-run safety comes from the transaction + `StaleEntryError`).

## Test plan

- TDD throughout: 11 new tests written first and watched fail (device-flow one confirmed a live authz bug), then made green.
- Gauntlet: `mix format` clean, `credo --strict` 0 issues, dialyzer clean, full `mix test --warnings-as-errors`: 3949 tests green.
- Known unrelated flake surfaced during one full run: `readiness_test.exs` DNS wall-clock bound (2.26s vs 2.0s under full-suite load) — the #1158/#1177 lineage; passes in isolation and in the other full run. Not touched here per its own "do not raise the threshold" comment; if it keeps recurring on loaded boxes, the next lever is probably an `:integration` tag on the real-network test.


Rasbandit:
### Code review

Found 3 issues (multi-agent review pass; all three fixed in 2afa20e1):

1. `Storage.Database.exists?/1` — the PR made it raise on DB errors, but the comment's claimed parity with S3 was backwards: `S3.exists?/1` deliberately logs-and-returns-false (decision in 98a94976), and the `@callback exists?(key) :: boolean()` contract promises a boolean. Fixed to log-and-return-false, matching the sibling adapter; the original sin (a bare silent `rescue _ -> false`) stays fixed.

https://github.com/engram-app/Engram/blob/dc0a9c0a44712e5924409908014468638ff8a26e/lib/engram/storage/database.ex#L65-L75

2. New `CrdtChannel` checkpoint rescue interpolated `Exception.message(e)` into the log message — the rescued call wraps crypto decrypt/encrypt whose raw errors can carry secrets, the exact invariant `Engram.Logger.DecryptFailure` documents and this same PR enforces two files over. Fixed to log only the bounded `error_kind`.

https://github.com/engram-app/Engram/blob/dc0a9c0a44712e5924409908014468638ff8a26e/lib/engram_web/channels/crdt_channel.ex#L431-L442

3. Off-by-one in the stamp chunking: the legacy `since` boundary is inclusive, so a same-stamp run of EXACTLY page size (500) wedges the feed too — the delete path chunked at 500, and `batch_upsert_notes` still stamped up to 500 creates with one shared `now` (the `changes_server_time` comment's "fewer than limit" bound). Fixed: both paths stamp through `@stamp_chunk` (400); stale controller comment rewritten; tests now assert max same-stamp run < 500.

https://github.com/engram-app/Engram/blob/dc0a9c0a44712e5924409908014468638ff8a26e/lib/engram/notes.ex#L2050-L2072

PR #1190, opened 2026-08-01
refactor(web): dedupe plug halts + batch-op helpers

Rasbandit:
## Summary

PR B of the review-remediation series (stacked on #1188 — merge that first; GitHub will auto-retarget this to main). Web-layer DRY pass, behavior-identical on the wire except one deliberate change:

- **`EngramWeb.Plugs.Halt.json/3`** — the shared put_resp_content_type + send_resp + halt idiom, replacing 16 hand-rolled copies across the plugs directory (incl. `require_admin`'s two mismatched idioms in one file). Extra headers (`Retry-After`, `X-RateLimit-*`) stay at call sites; `LimitResponse` keeps its own header semantics, untouched. `AccountLifecycle`/`AccountDeleted` moved off `Phoenix.Controller.json` — wire-identical, pinned by the new halt test.
- **`format_errors` dedup** — deleted the private reimplementation in `vaults_controller` and the inline `traverse_errors` in `users_controller`; both call the canonical `EngramWeb.format_errors/1`.
- **BEHAVIOR CHANGE:** `PATCH /api/me` 422 now returns the sibling shape `%{errors: ...}` instead of the one-off `%{error: "validation_failed", details: ...}`. Frontend audited: the SPA only reads the generic `body.error` fallback, nothing parses the old shape. OpenAPI schema updated, unused `Schemas.ValidationError` deleted. TDD (test flipped first, watched red).
- **`EngramWeb.BatchOps`** — byte-identical `parse_uuid_list/parse_uuid/parse_move_target` plus event-parameterized `broadcast_batch` extracted out of the notes/folders controllers. `send_move_result` deliberately NOT unified: success bodies and error-clause sets genuinely differ (`moved_attachments`, `cycle`/attachment-leg errors vs `batch_too_large`) — documented in the moduledoc.

Net −33 lines.

## Test plan
- New halt-helper unit test (status/body/content-type/halted) pins the wire format.
- Users 422 shape change done test-first.
- Gauntlet: format, compile --warnings-as-errors, credo --strict clean; full suite 3952 tests, 0 failures.

### DIFF
--- a/e2e/tests/api_only/local/test_62_local_account_settings.py
+++ b/e2e/tests/api_only/local/test_62_local_account_settings.py
@@ -127,7 +127,9 @@ class TestProfileUpdate:
             timeout=10,
         )
         assert patch.status_code == 422
-        assert patch.json()["error"] == "validation_failed"
+        # Sibling-controller 422 shape (errors map keyed by field) since the
+        # web-layer DRY pass unified PATCH /me with every other controller.
+        assert "display_name" in patch.json()["errors"]
 
 
 # ---------------------------------------------------------------------------
--- a/frontend/src/api/client.test.ts
+++ b/frontend/src/api/client.test.ts
@@ -77,6 +77,22 @@ describe("api client 402 handling", () => {
 		await expect(api.get("/x")).rejects.toBeInstanceOf(ApiError);
 	});
 
+	it("derives the ApiError message from a 422 errors map when error is absent", async () => {
+		// The unified controller shape is %{errors: %{field => [messages]}} with
+		// no top-level `error` key — the message must surface the field detail,
+		// not degrade to the generic statusText.
+		globalThis.fetch = vi
+			.fn()
+			.mockResolvedValue(
+				new Response(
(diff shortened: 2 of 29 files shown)


Rasbandit:
### Code review

Found 1 issue (fixed in dbf42789):

1. The `PATCH /api/me` 422 shape unification silently degraded the SPA's validation toasts — the generic error surface only read `body.error`, which the unified `%{errors: ...}` shape doesn't have, so `err.message` fell back to the bare statusText ("Unprocessable Entity"). This was the one endpoint where the old shape actually surfaced field detail to the user. Fixed at the root in the shared client: the ApiError message now falls back to the first "field: message" entry of the errors map, which also un-degrades every sibling controller's 422s (they all already had this gap).

https://github.com/engram-app/Engram/blob/a8b2a0e4a30a19a55925906427f16b25f9ac2a41/frontend/src/api/client.ts#L54-L58

Also verified clean by the other review passes: all 16 plug conversions are wire-identical (headers/assigns preserved, charset parity pinned by test), the no-op-fanout broadcast gating from #987/#969 survived the BatchOps extraction, and the openapi.json change is generator-produced.

PR #1192, opened 2026-08-01
fix(mcp): answer 405 for the stream-open GET instead of 406

Rasbandit:
Found while diagnosing a failed **Cursor** connection to staging.

PR #340 added `GET/DELETE /api/mcp` → **405 + `Allow: POST`** so a Streamable-HTTP client learns "POST-only, carry on" instead of treating the endpoint as dead. **That handler has been unreachable for the only clients it was written for.**

A Streamable-HTTP client opens the server→client stream with `Accept: text/event-stream`. Those routes piped through `:api`, whose first plug is `plug :accepts, ["json"]` — so the request is rejected **406** before `McpController.unsupported_transport/2` ever runs.

## Measured against live staging (0.12.3)

| Method | Accept | Result |
|---|---|---|
| GET | `text/event-stream` | **406** |
| GET | `application/json` | 405 `Allow: POST` |
| GET | `*/*` | 405 `Allow: POST` |
| GET | `application/json, text/event-stream` | 405 `Allow: POST` |

Cursor's log, 2026-08-01: Streamable-HTTP POST failed → fell back to the legacy HTTP+SSE transport → `SSE error: Non-200 status code (406)` → `Connection failed`.

## Fix

A second pipeline, `:api_any_accept` — identical to `:api` minus content negotiation — used only by these two routes. Negotiating a representation for a request being rejected on **method** alone is meaningless; the answer is 405 + `Allow` regardless of what the client would have accepted.

## Deliberately narrow

`POST /api/mcp` with `Accept: text/event-stream` **alone** also 406s, and I left that alone:

- the MCP spec requires clients to accept `application/json` on POST, so a conformant client sends both types and already works;
- that route lives in the big vault-scoped authed scope shared with notes/search/folders, and widening negotiation there to accommodate an out-of-spec Accept is not worth the blast radius.

Happy to revisit if a real client turns out to need it.

## Why the existing tests missed it

`mcp_transport_test.exs` uses a bare test conn, which sends **no Accept header** — so it never exercised the header a real client sends. The two new negative tests fail on the old router with `Expected one of ["json"]`. A third pins that the spec-conformant dual Accept keeps working, so this widens what is tolerated rather than moving the answer.

## Verification

`mix format`, `mix compile --warnings-as-errors`, `mix credo --strict` (no issues), 3941 tests.

One unrelated failure in the full run: `Cluster.ReadinessTest`'s wall-clock resolver bound (2.66s vs 2.0s). It reproduces only under the load of back-to-back full suites and passes **3/3 in isolation**; #1177 documents that exact cause. Nothing in this diff touches DNS resolution.

> Note: the 406 was **not** the primary blocker in the reported Cursor session — that was a burst of Cloudflare **523 origin-unreachable** errors between 14:46 and 16:00 local. No request reached the app or NPM during that window, so the break was between Cloudflare and the origin, upstream of the proxy. Separate, infrastructure-side, and self-recovered. This fix matters for the next attempt, once the origin is reachable.

https://claude.ai/code/session_01BQ4aJGK1eWxoghJLrw8Fuc

### DIFF
--- a/lib/engram_web/router.ex
+++ b/lib/engram_web/router.ex
@@ -10,6 +10,19 @@ defmodule EngramWeb.Router do
     }
   end
 
+  # Same as `:api` minus content negotiation. Used only by the MCP
+  # method-not-allowed routes below: a Streamable-HTTP client opens the
+  # server→client stream with `Accept: text/event-stream`, which
+  # `plug :accepts, ["json"]` answers 406 — before the controller that would
+  # have told it "POST-only". Negotiating a representation for a request we are
+  # rejecting on method alone is meaningless, so this pipeline simply does not.
+  pipeline :api_any_accept do
+    plug :put_secure_browser_headers, %{
+      "x-content-type-options" => "nosniff",
+      "x-frame-options" => "DENY"
+    }
+  end
+
   pipeline :rate_limit_auth do
     plug EngramWeb.Plugs.RateLimit, limit: 10, period: 60_000
   end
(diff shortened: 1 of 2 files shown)


Rasbandit:
Superseded by #1189 — this exact commit is now the second commit there, bundled with the Devin/LobeHub attribution fix. Same session, same theme (third-party MCP clients reaching us and getting the wrong answer), so one PR per the repo's one-initiative-one-PR convention.

No content lost: #1189 carries `cc2d884f` verbatim plus its three tests.

PR #1194, opened 2026-08-02
perf: batch attachment deletes, folder-rename writes, parallel creates

Rasbandit:
## Summary

PR D (final) of the review-remediation series — stacked on #1188 (merge order: #1188 → #1190/#1191 → this; GitHub auto-retargets). The structurally-unbounded write paths from the review:

- **`Attachments.batch_delete/3`**: was one transaction + one seq-allocation UPDATE + one synchronous storage call PER PATH. Now: one transaction, one contiguous seq-block allocation (per-row seqs stay unique + monotonic for the `(seq, id)` feed), one multi-row soft-delete via VALUES join, per-path broadcasts moved post-commit (same events/payloads). `delete_folder/3` inherits the win.
- **`Notes.batch_move_folders/4`**: was O(K × vault) — a full vault decrypt-scan per folder marker. Now one shared scan for all K markers, mirroring the delete-folders path. Per-marker result/rollback semantics preserved.
- **`do_rename_folder/5`**: per-note `update_all` (thousands of round trips in one held-open transaction for a big folder) → chunked multi-row `UPDATE ... FROM (VALUES ...)` with explicit casts, 500 rows/statement, inside the same transaction/tenant context.
- **`crdt_create_batch`**: up to 100 serial DB-write + SharedDoc round-trips inside the channel process → `Task.async_stream` (bounded concurrency, ordered results; socket-state mutations applied serially from task results). Reply contract unchanged (after all entries, input order).
- **`seed_genesis_if_empty/3`**: the `receive ... after 0` rendezvous (zero latency tolerance — could misreport `:error` after the doc mutation already applied) → bounded receive with a monitor on the room, so a dead room can't hang and a slow reply isn't misread.

Plus dialyzer-driven cleanup surfaced by the rewrite (dead error clauses) and the tenant-raw-SQL allowlist test update for the new VALUES-join statements.

## Test plan

- Gauntlet on the rebased branch: compile --warnings-as-errors, credo --strict, dialyzer, full suite — 3954 tests, 0 failures.
- New tests pin: distinct monotonic seqs across a batch delete (seq-feed correctness via `list_changes_by_seq`), folder-rename correctness over multiple notes, batch-create reply shape/order.
- ⚠️ **Run the Obsidian e2e suite before merging** — plugin-side e2e does not gate backend CI, and this touches the CRDT create path and sync feeds directly (`make e2e` after `make ci-up`, or let the full pipeline run and hand-check `e2e-crdt`/`e2e-clerk`).


Rasbandit:
## e2e-clerk failure is real, not a flake — root-caused

Three failures ([run 30726811237](https://github.com/engram-app/Engram/actions/runs/30726811237)) that look independent are one cascade:

```
test_77 bulk first sync   converged only 140/1000 notes in 121.5s
test_81 remote logging    trigger.md not on server after 120s
test_85 missed delivery   received=no materialized=no
  └─ crdt_create failed … {"reason":"unmatched topic"}
```

From the CI debug artifact (`docker-compose.log`), 14 × `DBConnection.ConnectionError`:

```
Task #PID<0.8307.0> started from #PID<0.7598.0> terminating
** (DBConnection.ConnectionError) [Elixir.Engram.Repo] connection not available and
   request was dropped from queue after 183ms
    lib/engram/repo.ex:84:                       Engram.Repo.run_with_tenant/2
    lib/engram/notes.ex:690:                     Engram.Notes.genesis_crdt_note/4
    lib/engram_web/channels/crdt_channel.ex:455: EngramWeb.CrdtChannel.prepare_create/3
    lib/task/supervised.ex:101:                  Task.Supervised.invoke_mfa/2
  Args: [... "path" => "Bulk/n0850.md"]
```

### Why

**1. Concurrency is sized off the wrong resource.** `max_concurrency: System.schedulers_online()` — but each entry opens its own `Repo.with_tenant` transaction and holds a pooled connection for its full duration, so N concurrent entries pin N connections. The e2e stack runs the `runtime.exs` default `POOL_SIZE=10` (nothing overrides it in the compose files). Phase 3 then runs *another* `schedulers_online` concurrent `seed_and_checkpoint` calls, which also write. The serial version never exposed this because it held exactly one connection at a time.

**2. One entry's failure kills the whole channel.** `Task.async_stream` links tasks to the caller, so the task's exit takes down the channel process (`#PID<0.7598.0>`). The socket then has no channel for that topic, and every subsequent client frame gets `{"reason":"unmatched topic"}` from Phoenix `socket.ex:871`. That is the cascade into test_81/test_85 — they were not touching the batch path at all, they were talking to a dead channel.

Note (2) is arguably pre-existing (a raise in the old serial `create_one` would also crash `handle_in`), but the serial shape never reached the pool timeout that triggers it. The batch contract explicitly promises per-entry partial failure, so a transient DB hiccup returning `create_failed` for one entry — instead of stranding the client on a dead topic — is the behaviour the docstring already claims.

### Fix (to apply after this rebases onto main)

- Derive `max_concurrency` from `pool_size`, not `schedulers_online`, with headroom left for the other channels, the checkpoint timers and the seq feed. Same for the phase-3 stream.
- Convert a per-entry task failure into that entry's `create_failed` result (logged, not swallowed) so the channel survives.
- Re-run the full e2e suite — the PR's own test plan already flagged this path as needing it.

Not merging until this is fixed and e2e is green.

Rasbandit:
### Code review

Found 4 issues (all fixed in 6998e1df):

1. `delete_folder/3` crashed with a MatchError on folders holding >500 attachments — it hard-matched `{:ok, _}` on `batch_delete/3`, which now returns `{:error, :batch_too_large}` above 500. The server-internal cascade now chunks under the request-boundary cap. TDD'd with a 501-attachment folder.

https://github.com/engram-app/Engram/blob/94f7edac82a2df15ceb2fbe4576d68f18ab6e78f/lib/engram/attachments.ex#L657-L664

2. The rewritten folder-rename stamped an unbounded cascade (all row classes + tombstones) with ONE shared timestamp — the exact legacy-feed wedge (`changes_server_time` bound, fixed for deletes in #1188) reintroduced at scale. Now every renamed row gets a per-row `bulk_stamp` from a single global counter across marker/v2/v1/tombstone classes (same-stamp runs ≤400 by construction; the shared per-op `seq` contract from #614 is untouched). TDD'd with a 501-note rename.

https://github.com/engram-app/Engram/blob/94f7edac82a2df15ceb2fbe4576d68f18ab6e78f/lib/engram/notes.ex#L3725-L3740

3. `batch_delete`'s `{:error, reason} -> {:ok, %{deleted: 0}}` recovery branch was unreachable dead code (the SQL path raises, never returns error tuples) — and it masked a real semantic change: DB failure now rolls back the whole batch instead of the old per-path independence. Branch removed; the all-or-nothing semantics are now stated honestly in a comment.

4. Batched `S3.delete_many/1` swallowed per-key partial failures (DeleteObjects returns 200 with an Errors list) that the old per-path deletes surfaced individually. Now counts `<Error>` entries, logs loudly (counts only, no key contents), and returns the actual deleted count. TDD'd via Bypass.

Also added (test-only): an order-pinning `crdt_create_batch` test with a conflicting entry mid-batch — the plugin matches results by array index (sync.ts), so index alignment is a silent-corruption class that the 3-phase `Task.async_stream` restructure must preserve (`ordered: true`, now pinned). One plugin follow-up flagged: the sync.ts comment citing "Enum.map_reduce" as the ordering mechanism is now stale (behavior unchanged).

⚠️ Reminder from the PR body stands: run the Obsidian e2e suite before merging — this touches the CRDT create path and sync feeds.

PR #1200, opened 2026-08-02
refactor(web): action_fallback rollout for shared error shapes

Rasbandit:
## Summary

Follow-up 1 of 4 from the backend review series: introduces `EngramWeb.FallbackController` + `action_fallback` and converts the 10 controllers whose error branches map 1:1 onto the repo's shared response shapes (notes, folders, attachments, vaults, connections, oauth_clients, users, auth, admin/invite, admin/user).

**Behavior-identical on the wire.** Seven shared shapes (422 changeset, 404 ×2, 409 ×3, 500 internal) each had 3-9 hand-rolled copies; every clause's status+body is pinned byte-for-byte by a new unit test, and bespoke branches stay inline by design — including deliberate oddities like notes' `"not found"`-with-a-space 404 dialect and attachments' `item_path` variants (guarded so they can't fall through to the wrong clause). 16 controllers with fully bespoke bodies are deliberately not converted. `attachments.rename` uses a guarded fallback so unknown errors crash loudly rather than silently becoming 500s.

Net ~−64 LOC in controller code; the moduledoc documents the never-invent-a-shape contract.

## Notes for reviewers

- Three wire-parity pins were added BEFORE converting; one revealed (and pinned as-is, unchanged) that vaults' POST 422 leaks internal field names (`name_ciphertext`/`name_hmac`/`name_nonce`) in changeset error keys — flagged as a possible future cleanup, deliberately not smuggled into this refactor.
- auth `create_api_key`'s 422 branch is unreachable over HTTP (changeset only validates server-set fields); pinned via the fallback unit test.

## Test plan

Full gauntlet: format, compile --warnings-as-errors, credo --strict, dialyzer, 745 controller tests + full suite 3991 tests, 0 failures.


Rasbandit:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance across five independent passes (guidance audit, wire-parity bug scan, git-history/client-contract trace, prior-PR concern mining incl. #1190 conflict check, code-comment invariant audit).

Specifically verified: both same-tuple-different-body traps stay inline (notes' space-variant `"not found"`, attachments' `item_path`); the frontend's hard-parsed `last_admin` contract and #725's bare-atom-vs-tagged-tuple `item_id` distinction survive byte-identically (pinned); idempotency-remember → broadcast → json ordering unchanged across every batch conversion; no #1190 BatchOps/format_errors decisions undone.

One informational note (improvement, not a defect): `admin/user#delete` and `admin/invite#delete` previously had no clause for rare error returns (changeset rollback / revoke errors) and would crash with a CaseClauseError; they now degrade gracefully through the fallback (422/500 respectively).

PR #1201, opened 2026-08-02
refactor(crypto): shared dual-DEK rewrap helper

Rasbandit:
## Summary

Follow-up 2 of 4 from the backend review series: the ~30-40-line "decrypt under OLD DEK → retry under NEW → double-failure log + row_failed telemetry + raise" block was copy-pasted SIX times inside `user_dek_rotation.ex` (the audit said five — `recrypt_blob` was a sixth). One `try_rewrap/7` helper now owns exactly the security-critical invariant (decrypt order, double-failure handling); everything site-specific stays at the call site:

- re-encryption (so tags' ETF re-encode, note HMAC transforms, and Qdrant base64 shapes are byte-identical),
- each site's exact log strings/metadata and telemetry table/phase,
- Qdrant's two deliberate divergences (same-AAD probes, error-tuple instead of raise) carried via explicit params — nothing silently harmonized.

## Test plan

- Two decryptability gaps filled BEFORE refactoring (note tags + attachment path metadata — previously only HMACs were asserted); both passed against pre-refactor code as a true baseline.
- Rotation tests re-run after EVERY individual site conversion (44/0 each), then crypto dir + mix-task + worker tests 244/0.
- Full gauntlet: format, compile --warnings-as-errors, credo --strict, dialyzer, full suite 3984 tests 0 failures.
- Honest note: one full-suite run under heavy parallel-agent machine load showed 3 CRDT-checkpoint-timer shutdown EXITs (area untouched here); clean on two subsequent full runs — watch on CI rather than consider diagnosed.


Rasbandit:
### Code review

No issues found. Checked via a maximum-suspicion parity pass over all six converted sites plus a git-history/comment-invariant pass.

Verified per site: decrypt order (old DEK first) preserved; AAD pairs exact (Qdrant's same-AAD-both-probes proven pre-existing from T3.7's original code, now correctly commented); `:already_rotated` no-op semantics identical; double-failure shapes carried explicitly (`{:raise, msg}` ×5, Qdrant's error tuple) — nothing silently harmonized; telemetry event/measurements/metadata and log strings byte-parity (keyword-order diff proven cosmetic); tags' ETF re-encode and blob's S3 PUT unchanged at call sites. History trace found no incident-driven per-site divergence beyond what the `on_both_failed` knob preserves. The two new tests assert production decrypt-path round-trips (a real pre-existing coverage gap), not helper internals.

PR #1202, opened 2026-08-02
refactor: tenant-scope composers + isolation lint

Rasbandit:
## Summary

Follow-up 3 of 4 from the backend review series. The tenant-isolation predicate (`user_id == ^ / vault_id == ^ [/ is_nil(deleted_at)]`) was hand-inlined 49 times across notes/vaults/attachments — in an encrypted multi-tenant product that predicate IS the isolation boundary, and nothing caught a dropped clause.

- **Composers**: `Notes.scoped/scoped_live`, `Attachments.scoped/scoped_live` (struct-or-id clauses), `Vaults.scoped` + `active/deleted` — mirroring the existing `notes_only/0` precedent. 46 sites converted with behavior-identical predicate sets (tombstone-including feeds gained NO deleted_at filter; extra clauses compose on top).
- **The real deliverable — a lint test** (`test/lint/tenant_scope_lint_test.exs`): scans the three context files for raw `.user_id == ^` predicates outside the helpers, attributes hits to their enclosing function, and fails naming file:line + function. Three allowlisted divergences, each with a WHY (cross-vault Qdrant resolution, bulk IN-list GROUP BY, account-wide billing usage). Negative-tested with an injected canary violation.

Context files shrank 57 lines; net +43 with the helpers + lint.

## Test plan
Full gauntlet: format, compile --warnings-as-errors (dev+test), credo --strict, dialyzer; per-context suites (791 tests) green; full suite 3983 tests — one unrelated flake on first run (isolate-rerun green), clean full rerun 0 failures.


Rasbandit:
### Code review

Found 2 issues, both in the lint itself (fixed in the follow-up commit):

1. The scan's `cond` was mutually exclusive — a one-line `defp ..., do: <query>` that both declares the head and inlines the predicate updated the head-tracker and was never predicate-checked, silently evading the lint. Fixed (head-tracking and predicate-check now run independently) and canary-verified both ways.

2. `field(x, :user_id) == ^...` (Ecto dynamic-field spelling) didn't match the pattern. Added to the regex; the remaining theoretical escapes (`dynamic/2` fragments, reversed operands) are documented at the pattern as known limitations — neither spelling exists in these files today.

The conversion itself is clean: all 46 sites verified predicate-identical against the exact bug class (no tombstone-feed gained a deleted_at filter — `find_folder_marker`, delete-wins probes, both seq feeds, and `clear_defaults` all specifically checked), helper nil-handling fails closed, all three allowlist entries confirmed legitimately divergent, and no invariant comments contradicted.

PR #1205, opened 2026-08-02
feat(otel): name the RLS tenant spans

Rasbandit:
Follow-on from the 2026-08-02 telemetry audit.

## The finding

**3,069 of 5,389 `repo.query` spans over 23h were anonymous** `engram.repo.query`. My first read of them was "background Oban noise." Wrong — there were **78** `oban_jobs` spans in that window. The anonymous ones are `Repo.with_tenant/2`'s own SQL, sitting on the hot request path.

Per `GET /api/sync/manifest`: 18 Ecto spans, **13 anonymous**, and those 13 are:

| | query time |
|---|---|
| 13 anonymous (RLS plumbing) | **7.9 ms** |
| 5 named (actual data) | **5.1 ms** |

Tenant setup costs more than the work it protects, and none of it was labelled — so it read as noise instead of as a number worth looking at.

## Why they were anonymous

`opentelemetry_ecto` builds the span name as:

```elixir
span_prefix <> if(source != nil, do: ":#{source}", else: "")
```

No per-query naming hook, and raw SQL has no schema source. So all four statements the RLS block issues — `begin`, set tenant+role, reset role, `commit` — render bare.

## The fix

Three keyword options. `source` for a raw query is plain `Keyword.get(opts, :source)` in `Ecto.Adapters.SQL.log/5`, and transaction opts reach the same call for `begin`/`commit` via `checkout_or_transaction/4`:

| span | statement |
|---|---|
| `engram.repo.query:tenant_txn` | begin / commit |
| `engram.repo.query:tenant_enter` | set `app.current_tenant` + role |
| `engram.repo.query:tenant_exit` | reset role |

Observability only — no SQL text, control flow, or timing change. `source` feeds telemetry metadata and constraint-error mapping; `set_config` can't raise a constraint error, so the latter is inert.

## Tests

Assert the block is *fully* named rather than pinning exact statements (sandbox runs it as a savepoint, so begin/commit may surface differently). One test covers `with_tenant`'s re-entrancy — that property is what would make collapsing adjacent call sites cheap later, and if it regresses every nested call silently resumes paying four round trips.

## Deliberately not included

Reducing the *number* of `with_tenant` blocks per request. That's the real ~5ms win, but the naive version — one outer transaction around the action — holds a DB connection across the Elixir-side decrypt loop. That's the exact shape behind the July pool-exhaustion incident against a 10-connection pool. Separate change, done carefully.

## Verification

- `mix format --check-formatted` clean
- `mix credo --strict` — no issues
- `mix test --warnings-as-errors` — **3,999 tests, 0 failures**
- `mix dialyzer` — passed

⚠️ **Two pre-existing flakes on `main`, surfaced not swept.** An earlier run of this branch showed 2 failures; I lost the detail by piping through `tail`, re-ran clean, then ran unmodified `main` to isolate. `main` fails the same two:

1. `EngramWeb.OpenApiEndpointTest` — `ExUnit.TimeoutError` at 60s, stuck in `:code_server.call/1` → `:code.ensure_loaded/1` inside OpenApiSpex schema resolution. Cold module-load contention.
2. `Engram.Cluster.ReadinessTest` — `assert elapsed_us < 2_000_000`, got **2,224,917**. A real `:inet_res` call against a 2s wall-clock bound, 11% over. #1177 already hardened this once; it's a recurrence.

Both load-sensitive, both reproduced on unmodified `main`, neither reachable from a diff that adds keyword options to telemetry metadata. Flagging rather than claiming a clean bill — #2 in particular looks like it wants a real fix, not another window widening.

https://claude.ai/code/session_01CvUvadaneVBQXax8NfeNJU

### DIFF
--- a/lib/engram/repo.ex
+++ b/lib/engram/repo.ex
@@ -81,31 +81,47 @@ defmodule Engram.Repo do
     Process.put(:engram_tenant, uuid)
 
     try do
-      transaction(fn ->
-        # `set_config(..., true)` == SET LOCAL, but as a regular SELECT it
-        # takes a bind parameter (no string interpolation) and applies the
-        # tenant + the engram_app role drop in a single round trip.
-        # Superusers bypass RLS even with FORCE — the role drop scopes
-        # enforcement to this transaction.
-        _ =
-          query!(
-            "SELECT set_config('app.current_tenant', $1, true), " <>
-              "set_config('role', 'engram_app', true)",
-            [uuid]
-          )
+      # `source:` is purely observability. Ecto reads it straight off the query
+      # opts (`Ecto.Adapters.SQL.log/5`) and opentelemetry_ecto appends it to
+      # the span name — there is no per-query naming hook, so without it these
+      # statements render as bare `engram.repo.query`. Transaction opts reach
+      # the same code path via `checkout_or_transaction/4`, so this also names
+      # the begin/commit pair.
+      #
+      # Worth the three keywords: a 2026-08-02 trace audit found 3,069 of 5,389
+      # repo.query spans anonymous over 23h, almost all of them this block on
+      # the hot path (13 per GET /api/sync/manifest). Anonymous spans got read
+      # as background-job noise; they are actually 7.9ms of tenant setup
+      # against 5.1ms of real data queries.
+      transaction(
+        fn ->
+          # `set_config(..., true)` == SET LOCAL, but as a regular SELECT it
+          # takes a bind parameter (no string interpolation) and applies the
+          # tenant + the engram_app role drop in a single round trip.
+          # Superusers bypass RLS even with FORCE — the role drop scopes
+          # enforcement to this transaction.
+          _ =
+            query!(
+              "SELECT set_config('app.current_tenant', $1, true), " <>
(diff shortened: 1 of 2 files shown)


Rasbandit:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

Five independent passes (instruction-file compliance, bug scan, git history, prior-PR comments, code-comment accuracy). Notes worth recording:

- Both prior fixes to this block survive: `0aa811cf` (role reset must be the last statement before commit/release, so it does not leak `engram_app` past `RELEASE SAVEPOINT`) and #549 `367485d6` (three utility round trips collapsed to two, plus the same-tenant re-entrancy short-circuit). Statement count and ordering are unchanged.
- `source:` is inert beyond telemetry. `to_constraints/2` is only reachable from `Ecto.Adapters.SQL.struct/10` (schema insert/update/delete); raw `query!/3` never hits constraint mapping. `:source` appears nowhere in `postgrex`, so statement caching is unaffected.
- Transaction opts do reach BEGIN/COMMIT logging via `checkout_or_transaction/4`, and are not stored in the process dictionary, so `source: "tenant_txn"` cannot leak into queries made inside `fun.()`.

PR #1206, opened 2026-08-02
fix: key vault-name validation errors on the public name field

Rasbandit:
## Summary
Two deferred cleanups from the 2026-08-01 review-remediation wave:

**Vaults 422 internal-field leak.** `POST /api/vaults` with a missing `name` returned a 422 whose error keys were the internal encrypted columns (`name_ciphertext`, `name_nonce`, `name_hmac`) — leaking schema internals into the public API. The trio is populated by `inject_name_phase_b/3` from the client-supplied `name`, so a missing trio means the caller omitted `name`. `Vault.changeset/2` now validates that as a single error on the virtual `:name` field (`validate_encrypted_name/1`), keeping the Phase B trio invariant enforced. TDD: flipped the two pins (`vaults_controller_test.exs` from #1200, `vaults_test.exs`) first, watched RED, then fixed.

**Cache test coverage.**
- The "plan limit caching" describe moves from `billing_test.exs` into a dedicated `test/engram/billing/plan_cache_test.exs` (unchanged assertions; doubles as direct coverage of the `Engram.Cache.PersistentTerm` macro). `billing_test.exs` stays `async: false` for its OverrideCache tests; header comment updated.
- New `test/engram/cache/node_local_ets_test.exs` pins the `Engram.Cache.NodeLocalEts` macro contract directly via throwaway cache modules: TTL round-trip/expiry, `cache_fetch/2` read-through, `clear_local/0`/`delete_local/1`, the `delete_local/1` override seam (user-keyed eviction), and rescue-guard degradation when the table is absent.

## Test plan
- `mix format`, `mix compile --warnings-as-errors`, `mix credo --strict`, `mix dialyzer` all clean
- Full `mix test --warnings-as-errors`: 4004 tests, 0 failures

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_015buSiNhFQTvQ8ELdkTuLpz

PR #1213, opened 2026-08-02
fix: attachments 422 keys on public fields + reject blank vault names

Rasbandit:
## Summary
Follow-through on the two API-hygiene decisions from the review-remediation wave:

**Attachments 422 internal-key leak** (same class as vaults #1206): the attachment changeset's `{:error, changeset}` reaches the FallbackController, so its error keys are public. `validate_encrypted/3` now surfaces the encrypted path trio as `path` and `content_nonce` as `content` instead of leaking column names. TDD: flipped the pinned context test first, watched RED.

**Blank vault names rejected**: `name: ""` (or whitespace) previously created a vault — the empty string encrypts to a non-empty trio and `slugify("")` falls back to slug `vault`. Now:
- create: blank is treated as absent in `inject_name_phase_b/3` → 422 `{"errors": {"name": ["can't be blank"]}}`
- update: explicit guard in `update_vault/3`. Without it a blank rename was a **half-update**: `"" && "" != name` is truthy, so `maybe_regenerate_slug` rewrote the slug from `""` while the stored name kept its old value.

Consumers checked: e2e greps show only the DB-level crypto probe touches these column names (never error bodies); frontend `firstValidationError` parses keys generically.

## Test plan
- 5 new tests written RED-first (attachments context ×2, vaults context ×2, controller ×1)
- `mix format` / `compile --warnings-as-errors` / `credo --strict` / `dialyzer` clean
- Full `mix test --warnings-as-errors`: 4011 tests, 1 failure = known readiness DNS wall-clock flake (readiness_test.exs, #1158/#1177 lineage; green in isolation, file untouched here)


Rasbandit:
### Code review

Found 1 issue (fixed in-PR):

1. The blank-name gate also changes `register_vault/3` (plugin OAuth vault registration) — a third, unadvertised call site of `inject_name_phase_b/3`. Desired behavior (the plugin sends the vault directory basename, never blank), but it was untested. Pinned with a context test in the follow-up commit.

Otherwise clean: the `do_update_vault` extraction preserves exact semantics (order, `clear_defaults` timing, `unwrap_transaction` shapes); all three `Attachment.changeset` call sites populate the trio server-side (upsert insert/update + tombstone), and rotation/rebind/backfill bypass the changeset via `update_all`; no e2e/plugin/frontend consumer parses the old internal error keys; required-field coverage is unchanged (nothing silently dropped).

PR #1215, opened 2026-08-02
refactor: retire the legacy timestamp change feed (410 Gone)

Rasbandit:
## Summary
Decision follow-through: prod Loki shows **zero authenticated requests** to `GET /api/notes/changes` / `GET /api/attachments/changes` in 14 days (only GPTBot probing docs-example URLs, 401). All live clients use the CRDT socket + `/sync/manifest`; the current plugin's only references to the feed are historical comments.

- Both endpoints now return **410 Gone** unconditionally (routes stay, so 410 not 404); OpenAPI ops marked `deprecated` with a 410-only response; `ChangesResponse`/`AttachmentChange(s)Response` schemas removed; `openapi.json` regenerated.
- Dead context code removed: `Notes.list_changes_page/4` + cursor codecs + trace sampling, timestamp `Attachments.list_changes/3`. Everything `*_by_seq` (the seq feed) stays.
- **Not** in this PR: `bulk_stamp`/`@stamp_chunk` removal (the constraint the feed imposed) — separate follow-up once this lands, per the batch-cap lesson (`docs/context/batch-write-caps-and-chunk-stamping.md` got a retirement banner).
- E2e harness (an API consumer!) ported: tests 19/32/72/74 observe via `get_manifest()`; test_76 (feed pagination contract) deleted; test_72 asserts the 410; `api.py get_changes` removed. Backend tests that used the endpoint as an arbitrary authenticated-GET probe (router pipeline, onboarding gate, account lifecycle, multi-tenant) moved to `/sync/manifest` — an always-410 endpoint can't prove an upstream plug didn't halt.
- TDD: controller 410 tests flipped first and watched RED (200 → 410).

## Test plan
- `mix format` / `compile --warnings-as-errors` / `credo --strict` / `dialyzer` clean
- Full `MIX_TEST_PARTITION=7 mix test --warnings-as-errors`: 3976 tests + 1 property, 0 failures
- `ruff check .` clean in e2e/
- e2e CI on this PR exercises the ported manifest-based tests

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_015buSiNhFQTvQ8ELdkTuLpz

PR #1216, opened 2026-08-02
refactor: remove chunk-stamping from bulk writes

Rasbandit:
## Summary
Follow-up to #1215. `bulk_stamp/2` + `@stamp_chunk 400` existed only to keep same-timestamp runs under the retired feed's 500-row page (the inclusive-boundary wedge). The seq feed orders by `(seq, id)`, so bulk deletes, batch upserts, and folder-rename cascades now share one `DateTime.utc_now/0`.

- Kept: `@max_batch_entries 500` upsert cap (compute-DoS guard) + its `batch_too_large` test, all seq-block and cascade-correctness assertions (>500-id delete acceptance, 1003-row rename cascade).
- Deleted: the distinct-`updated_at` pins — each observed FAILING against the new code before deletion, proving the stamping actually stopped.
- `lib/engram/attachments.ex` needed no changes (its #1194 batch delete already shares one timestamp; distinctness there is seq-only).
- `docs/context/batch-write-caps-and-chunk-stamping.md` flipped to HISTORICAL.

## Test plan
- Full gauntlet clean: format / compile `--warnings-as-errors` / credo `--strict` / dialyzer
- `MIX_TEST_PARTITION=6 mix test --warnings-as-errors`: 3981 tests + 1 property, 0 failures

### DIFF
--- a/docs/context/batch-write-caps-and-chunk-stamping.md
+++ b/docs/context/batch-write-caps-and-chunk-stamping.md
@@ -1,19 +1,25 @@
 # Batch-write caps + chunk-stamped tombstones — the fewer-than-500-rows-per-TIMESTAMP invariant
 
-_Last verified: 2026-08-01_
+_Last verified: 2026-08-02_
 
-**Status:** shipped on PR #1188 (review-remediation bundle); #1194 (batch
-write-path perf rewrite) builds on the same mechanism.
-**Symptom class:** a legacy (pre-pagination) client's change-feed pull loops
-forever on one page; or, the day-one version of the trap, e2e teardown gets
-422 `batch_too_large` on a legitimate 1100-id bulk delete.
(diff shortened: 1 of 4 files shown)


Rasbandit:
### Code review

Found 1 issue (fixed in-PR):

1. Both `@max_batch_entries` comments in `lib/engram/attachments.ex` still justified the cap by the retired feed's convergence bound — reworded to the cheap-guard rationale (plus a `(historical)` tag on the doc's fix section).

Verified clean: `bulk_rename_update!` VALUES shape and column lists unchanged with both renamed rows and tombstones still stamped; the single `update_all` in `batch_delete_notes` targets a provably identical id set (not-found guard); the only `order_by updated_at` consumers (embed-reconcile drain, attachments listing, frontend tree sort) tolerate ties; no Oban dedup, conflict resolution, or e2e assertion depends on distinct stamps; deleted tests were observed failing against the new code before deletion.

PR #1227, opened 2026-08-04
ci: version the CI-registry retention script

Rasbandit:
Closing a durability gap from #1225: the retention script was running on the FastRaid Unraid flash drive and existed in no repo. A cron'd `rm -rf` against the CI registry with no history, no review, and no copy if that flash drive is reset.

Placed in `ci/` beside `ci/fingerprint/` deliberately — fingerprint *writes* the `ci-*` pass markers this script expires, and the marker retention window has to stay shorter than the image window. If a marker outlives its image, CI skips `prebuild-ci-image` on the strength of the marker and then fails pulling an image nobody rebuilt. Keeping the two adjacent makes that coupling visible to whoever changes either one.

No behaviour change — this is the exact script currently installed, plus a header recording the install path and the `scp` to redeploy.

### DIFF
--- a/ci/prune-ci-registry.sh
+++ b/ci/prune-ci-registry.sh
@@ -1,0 +1,136 @@
+#!/bin/bash
+# Retention for the local CI registry (LocalCIRegistry, 10.0.20.214:5001).
+#
+# THIS FILE IS THE SOURCE OF TRUTH. It runs on the FastRaid Unraid host, not in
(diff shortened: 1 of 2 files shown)


Rasbandit:
Superseded — `ci/**` is not in verify.yml's push filter, so no CI ever ran and the required checks could not report. Reopened on a `chore/` branch.

Issue #1231, opened 2026-08-04
note_links: folder-rename/batch-move don't rebind path-qualified link targets


Rasbandit:
Bounded staleness (basenames unchanged in most moves; self-heals per source note on next edit); needs bulk fan-out of distinct basenames; from the final review of the note_links branch.

Rasbandit:
PR #1240 (phase 3) implements the bulk fan-out this issue asked for: `do_rename_folder` now enqueues a `RebindNoteLinks` job per distinct moved basename (case-insensitive hmac dedup), so path-qualified edges under a renamed folder re-resolve — proven end-to-end by the folder-rename round-trip test, including a pre-typed dangler at the new path getting bound.

One residual deliberately left as-is: a *dangling* `[[old-folder/Ghost]]` link whose target never existed keeps its stale text after the folder rename. It's unreachable by hmac-keyed fan-out (no moved note carries that basename), fixing it would require decrypting every edge in the vault (rejected), and Obsidian's own auto-update has the same behavior — it doesn't rewrite links to nonexistent files. Cosmetic only: the edge binds to nothing under either text, and self-heals if the ghost note is ever created at the new path.

Suggest closing this once #1240 merges, unless the residual should stay tracked separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Rasbandit:
Shipped in #1240 (merged): per-distinct-basename rebind fan-out on folder rename, proven by the folder-rename round-trip test. Known cosmetic residual (dangling links to never-existed notes keep stale text, matching Obsidian's own behavior) documented above — reopen if that ever warrants a decrypt-all sweep.

Issue #648, opened 2026-06-19
Wikilink integrity: update [[refs]] in all notes when a file/image is moved or renamed


Rasbandit:
## Problem

When a note or attachment is moved or renamed, existing `[[wikilinks]]` and `![[embeds]]` pointing at the old path become broken. Today the rename succeeds but referencing notes are not updated, so links silently rot across the vault.

## Desired behavior

Renaming/moving `old/path.md` → `new/path.md` (or an image) should atomically:
1. Rewrite all `[[old/path]]`, `[[old/path|alias]]`, `![[old/path]]`, `[[old/path#heading]]`, `[[old/path#^block-id]]` references across every note in the same vault to the new path.
2. Preserve aliases, header refs, block refs.
3. Preserve display short-form vs long-form (e.g. if the link was `[[path]]` and the new location is unambiguous, keep short form; if collision, qualify).
4. Touch `updated_at` on every modified referring note so plugin sync picks them up.
5. Run inside a single transaction (or compensating action on failure) so we never leave the vault in a half-rewritten state.

## Where the rename happens

- Web-editor rename → backend (`Engram.Notes` rename path — currently no link rewriter found via grep).
- Plugin rename → Obsidian's built-in "Automatically update internal links" handles vault-local rewrites; the resulting edits flow to backend as normal note updates, so plugin side is mostly fine. **But** any rename initiated server-side (web editor, MCP `move`, future bulk ops) needs a server-side rewriter.
- MCP tool: `lib/engram/mcp/tools.ex` + `handlers.ex` (already reference rename paths).

## Proposed approach

**New module:** `Engram.Notes.LinkRewriter`
- `rewrite_refs(vault_id, old_path, new_path, opts)` — scans notes in the vault for wikilinks matching `old_path` (any form), rewrites in place.
- Use existing markdown parser (remark/Phoenix has one already for indexing) — do NOT regex over markdown bodies; code blocks and inline code must be left alone.
- Return list of affected note IDs so the caller can emit sync events.

**Wire into:**
- `Engram.Notes.rename_note/3` (and attachment rename)
- MCP `move_note` handler
- Vault-rename / folder-rename paths

**Edge cases**
- Case-only rename on case-insensitive filesystems
- Rename to a path that already has incoming links (merge intent — fail or warn)
- Embeds vs links — both forms covered
- Attachment moves: `![[image.png]]` and Markdown `![](image.png)` both need rewrite (Obsidian-style users mix these)
[...]

PR #1229, opened 2026-08-04
feat: wikilink graph — note_links edges, resolution, backlinks, [[ autocomplete

Rasbandit:
Closes #591. Refs #592 (backlinks panel shipped here; visual graph view remains). Supersedes #1223 (its commits are included).

**Backend:** `note_links` edge table (RLS, per-row AAD-bound ciphertext, uuidv7) + `basename_hmac` columns on notes/attachments — Obsidian-style case-insensitive resolution as an indexed HMAC lookup. Extraction runs in the indexing pass (code-fence/frontmatter-aware, invalid-UTF-8-safe); lifecycle rebinding covers create/rename/delete/resurrect on BOTH REST and CRDT-native paths, plus attachments; DEK rotation rotates edge ciphertext and rebuilds all basename HMACs; chained 3-scope backfill worker (`mix engram.backfill_note_links` / release rpc `Engram.Links.Backfill.enqueue_all()`); `GET /api/notes/by-id/:id` gains `links`, new `GET /api/notes/by-id/:id/backlinks`.

**Frontend:** wikilinks render direct `/:slug/:uuid` hrefs from the payload map (unresolved targets fall back to the `/:slug/wiki/*` resolver, so fresh links work before indexing catches up); backlinks panel in the right rail; `[[` autocomplete fed by the manifest; SPA navigation everywhere (no full-page reloads).

**Migrations:** 3, expand-only (`create_note_links`, `add_basename_hmac`, concurrent indexes).
**Deploy follow-up:** run the backfill once post-deploy: `rpc 'Engram.Links.Backfill.enqueue_all()'`.
**Spec:** Engram vault → `50 Engineering/_Superpowers Specs/2026-08-03-note-links-graph-design.md`. Implementation ran subagent-driven with per-task + final adversarial reviews (7 review-caught bugs fixed pre-PR, incl. a rotation-seam resolution break, a CRDT-path rebind gap, and an Oban unique-option backfill stall).
**Note:** full backend suite green; two unrelated tests (NotesTest log-capture, CrdtHeadPropertyTest) showed one-off contention flakes on an earlier run — see PR checks.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01RqgVPyDtDbXmTqm4PkTioK

PR #1240, opened 2026-08-05
feat: server-side wikilink rewrite on rename (#648) — REST/MCP, web CRDT, and folder origins

Rasbandit:
Phase 1 of rename/move link propagation (Refs #648). When a note or attachment is renamed/moved via REST or MCP, every referring note's `[[wikilinks]]`/`![[embeds]]` are rewritten server-side to the new name — authored as real Y-updates through the client-update persistence path, so live editors and syncing plugins converge like on any ordinary edit.

## Design

Spec: Engram vault `50 Engineering/_Superpowers Specs/2026-08-04-rename-link-rewrite-design.md`; plan: `docs/superpowers/plans/2026-08-04-rename-link-rewrite-phase1.md`.

**One-rewriter invariant:** exactly one party rewrites — Obsidian for plugin-origin renames (its "automatically update internal links"), the server for everything else. Phase 1 wires only the unambiguously server-side origins (`Notes.rename_note/4`, `Attachments.move_attachment/4`; MCP routes through both — pinned by test). CRDT-origin (web) renames + `client_type` socket tagging are Phase 2; folder renames (#1231) are Phase 3.

## What's here

- `Engram.Links.Parser` — occurrences now carry `target_start`/`target_len` byte offsets (extraction behavior unchanged)
- `Engram.Links.live_basename_count/3` + `pre_rename_winner?/6` — form-rule ambiguity + pre-rename resolution reconstruction, reusing the existing private resolution machinery (rules live in ONE place)
- `Engram.Links.Rewriter` — planning core (occurrence selection, bare/qualified/casing form rule, byte splice) + write half: one Y-transaction of positional UTF-16 edits per source note, persisted via `SharedDoc.update_doc/2` (live room) or `CrdtPersistence.update_v1/4` + `CheckpointNote` (roomless, bounded head-advance retry); legacy pre-CRDT rows via `Notes.upsert_note/4`; never content snapshots
- `Engram.Workers.RewriteNoteLinks` — chunked cursor-chain Oban worker (NO `unique:` per the BackfillNoteLinks trap), args carry ids + base64 HMACs only (old path recovered from the rename tombstone at run time), per-source error isolation + `[:engram, :links, :rewrite, :failed]` telemetry → PromEx counter
- Source-set keyed on `target_basename_hmac`, not `target_note_id` — the same rename enqueues `RebindNoteLinks` jobs that flip edge ids on the same queue; hmac keying is rename-order-proof (both interleavings verified convergent)

## Verification

- TDD throughout; 7 tasks each through spec + quality review, fix rounds on Tasks 4/5 + one final-review round (see below)
- Round-trip property (extract → rename → rewrite → re-extract = same edge set at new target) and both-orders CRDT convergence, including an overlapping-span race that goes red if the head-advance retry is removed (proven by deletion)
- Astral-plane UTF-16 offset test (emoji before link) pins `utf16_len/1`
- No-`unique:` regression proven via `Oban.drain_queue` through real executing state (red/green re-validated after the wiring landed)
- Full gauntlet green: format, credo --strict, dialyzer (2 documented ignores, see `.dialyzer_ignore.exs` comments), `mix test --warnings-as-errors` — 4,101 tests, 0 failures

## Notable review catches (all fixed in-branch)

- Room-exit swallow in the live-room persist path could report success for an unpersisted delta and permanently orphan the note from future walks — now falls through to the roomless path (structurally double-apply-proof: SharedDoc replies before its persistence self-message)
- Task 6's wiring enqueue silently broke/vacated two Task 5 worker tests (cross-task seam) — fixture now isolates the wiring job

Deferred with rulings (final review): `Rewriter` sync-walk entry points have no production callers (worker path has full isolation); `with_tenant` MatchError unreachable from current funs.


Rasbandit:
**Phase 2 pushed** (`3b9f6dfa..dc427d5c`): CRDT-origin (web) renames now trigger the rewrite, gated by a join-time `client_type` tag.

- `client_type` join param → normalized socket assign (≤32-byte binary bound; join-time only, no per-message override) → threaded through `genesis_crdt_note/5` → `genesis_relocate_live/6` → gated enqueue via `Notes.crdt_rename_rewrites?/1`
- Gate semantics: `"obsidian"` → never; `"web"`/unknown-present → rewrite; **untagged → treated as obsidian for now** (deliberate skew-safety compromise; flips to rewrite via one-line `@untagged_crdt_client_type` change one release cycle after plugin PR engram-app/Engram-obsidian#389 circulates — both the flag comment and two test assertions carry pointers)
- CRDT relocates create no rename tombstone, so the worker gains an old-path fallback: AAD-bound ciphertext in job args (`aad_for_row("oban_rewrite_note_links", "old_path", target_id)`), tombstone-first recovery unchanged for REST. Replay/splice + garbage-ciphertext discard tests included; all failure modes fold to one opaque discard atom (no ciphertext-validity oracle)
- Enqueue rides the relocate's tenant transaction (atomic with the rename; `Enqueue.enqueue/2` is non-raising so a failed insert can't roll the rename back)
- Frontend declares `client_type: "web"` at its single CRDT join (exact-match pinning test)

Final phase-2 review: invariant holds under hostile review (reconnect, spoofing, batch path), flip-readiness verified (post-flip untagged jobs carry the ciphertext → real rewrites, not discards). Gauntlet green: format, credo --strict, dialyzer, 4,117 tests 0 failures with warnings-as-errors.

Phase 3 (folder renames, #1231) is being planned and will land on this branch next.

Rasbandit:
**Phase 3 pushed** (`dc427d5c..7db84c8c`): folder renames/moves now propagate. Refs #1231.

Research collapsed this phase to one production seam: a folder rename IS N note renames with unchanged basenames, and `do_rename_folder`'s cascade already writes per-note tombstones — so the phase-1 worker handles it with hmac-only args (no ciphertext needed), and the phase-1 form rule already does the right thing per occurrence (bare `[[Note]]` → no edit since the basename still resolves; `[[old-folder/Note]]` → new prefix; collision created by the move qualifies bare links).

- `do_rename_folder`'s existing side-effect loop enqueues one `RewriteNoteLinks` per moved note + one `RebindNoteLinks` per distinct moved basename (case-insensitive hmac dedup) — the #1231 bulk fan-out. Covers REST, MCP, and batch-folder-move in one seam (all funnel through `do_rename_folder`); attachment leg already fans out via phase 1's `move_attachment` (test-pinned).
- No origin gate here: the plugin never calls `POST /folders/rename` (verified — Obsidian folder renames arrive as per-file CRDT relocates, already gated by phase 2), so the spec's `X-Engram-Origin` header guarded a caller that doesn't exist and was dropped.
- End-to-end round trip proven through `Folders.rename` + real job execution: qualified text rewritten, bare links untouched, edges re-resolved, and a pre-typed dangler at the NEW path gets bound by the rebind fan-out.
- Known residual (matches Obsidian's own behavior): a dangling `[[old-folder/Ghost]]` link to a note that never existed keeps its stale text — unreachable by hmac-keyed fan-out without a decrypt-all-edges sweep, which we rejected. Cosmetic only; it binds to nothing under either text.

Gauntlet green: format, credo --strict, dialyzer, 4,127 tests 0 failures with warnings-as-errors.

Rasbandit:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

Five independent review passes (CLAUDE.md/AGENTS.md compliance, shallow bug scan, git-history invariants, prior-PR review carryover, code-comment adherence) surfaced 10 candidate findings; none survived confidence scoring at the ≥80 bar. Strongest sub-threshold notes, for awareness: `Links.pre_rename_winner?/6` refetches+decrypts both candidate tables per occurrence with no memoization across a walk (score 75, perf-only — the class links.ex:198 documents eliminating for `bind_danglers_for_hmac`); the `RewriteNoteLinks` moduledoc still says "REST/MCP origins only" though phases 2-3 added CRDT and folder enqueue sites (75, doc-accuracy); `Parser.target_span/2` re-derives `parse_inner/1`'s tokenization with test-enforced rather than structural agreement (72).

Rasbandit:
**Review fix wave pushed** (`7db84c8c..130e632f`, 8 commits): all six actionable sub-threshold findings from the code review are fixed.

- `Links.pre_rename_candidates/5` — candidate sets now prefetched once per job batch on `Rewriter.build_target/5`'s target map and threaded to `pre_rename_winner?/4` per occurrence (was: fetch+decrypt of BOTH tables per occurrence). Regression-pinned with an Ecto-query telemetry counter that goes red under the old per-occurrence code. Resolution rules untouched — same shared `route_resolution`/`resolve_from_candidates`, `pre_rename_winner?/6` kept as a fetch-then-delegate convenience.
- `Parser.parse_inner/1` now computes the target byte offsets during its own splitting; `target_span/2` deleted — single source of truth, no test-enforced twin invariant.
- Worker moduledoc names all four enqueue origins; rewriter moduledoc carries the genesis_resurrect deferred-gap caveat.
- The `[:engram, :links, :rewrite, :failed]` `:reason` label is now a genuinely closed set (known pipeline atoms pass through; any exception → `:exception`, unknown shapes → `:other`), enumerated in the spec and pinned by tests.
- `enqueue_crdt_rename_rewrite` no longer hard-matches `Crypto.get_dek/1` inside the relocate transaction — a crypto error now logs (ids only) and skips the enqueue instead of rolling back the rename.

Skipped with rationale: folder-loop `insert_all` batching (pattern-consistent with the adjacent pre-existing enqueues; batching loses per-job failure telemetry), the sync-walk/worker loop duplication (documented spec/test entry point over shared primitives), `client_type` trust posture (deliberate compromise with documented flip plan), tenant-lint file list (pre-existing #1229 design decision).

Scoped re-review verified all six ADDRESSED with no new defects; gauntlet green (format, credo --strict, dialyzer, 4,131 tests 0 failures, warnings-as-errors).

Issue #1208, opened 2026-08-02
OAuth consent screen shows no verification signal


Rasbandit:
Split out of #1207 (fix for #1204).

`/oauth/authorize`'s consent page renders the raw self-asserted `client_name` and nothing else — no verified badge, no logo, no "unverified client" chip. That is what bounded #1204's severity to *aiding survival of review* rather than enabling the phish outright: an impersonating client looks identical to a real one at the moment the user is deciding.

At consent time we know more than the connections list does, not less:

- the **requested** `redirect_uri`, already matched against the registered list
- `cimd_url`, if the client published a metadata document

Both are exactly the inputs `Engram.Connections.LogoAllowlist.resolve/4` takes since #1207, and `resolve/4` is already grant-shaped (one URI, not a list). So this is mostly wiring, not new logic.

Showing the chip there moves the defence from post-hoc audit to the decision point. The unverified case is the one that matters: a client claiming to be Claude while redirecting to loopback should say so before the user clicks Approve.

Worth checking against the OAuth consent-UI guidance in the MCP spec while doing it.

Rasbandit:
The OAuth spec audit in #1241 turned up the concrete requirement this issue asks about
("worth checking against the OAuth consent-UI guidance in the MCP spec").

**For a CIMD client, the consent screen must identify the relying party by the HOST of the
`client_id` URL — not by `client_name`.** Anthropic's connector docs state it directly:

> Because the document is self-asserted, the consent screen must display the **host of the
> `client_id` URL** (not the `client_name` field) as the relying party.

That inverts what we render today. `oauth-authorize-page.tsx:232-240` does:

```tsx
const clientName = clientQuery.data?.client_name ?? "this app";
...
Authorize <span className="text-primary">{clientName}</span>
```

`client_name` is the one field in a CIMD document with **zero** proof behind it — anyone
can publish a document calling themselves anything. The `client_id` host is the only part
that is un-spoofable, because serving the document at all requires controlling that host.
So we are currently leading with the unproven string and hiding the proven one.

Two further requirements from the MCP authorization spec worth folding into the same pass:

1. Display the **redirect URI hostname** clearly at consent.
2. Warn explicitly when the client's registered redirect URIs are **loopback-only** — MCPJam,
   Claude Code, Cline, Cursor and Windsurf are all in this class, and CIMD cannot prove
   anything about a loopback client because any local process can bind the port.

Note this is now cheap: since #1207 the grant records the redirect actually used, and
[...]

PR #1241, opened 2026-08-05
fix(oauth): negotiate CIMD metadata instead of policing it

Rasbandit:
Every Claude connect has failed with `invalid_client` since 2026-08-01.

Nothing in the DCR path broke — **Claude stopped using DCR.** #1167 advertised
`client_id_metadata_document_supported: true`, and Anthropic's clients select CIMD the
moment an AS advertises that plus `none` in `token_endpoint_auth_methods_supported`. A
one-word change to a `/.well-known/` document repointed every client at a path that had
never carried real traffic.

## Root cause

`Cimd.store/3` validated a fetched vendor document with `Client.registration_changeset/2`
— DCR **registration policy**. Two different questions sharing one answer:

| | DCR (`POST /oauth/register`) | CIMD (fetched document) |
|---|---|---|
| Who is speaking | an anonymous stranger asking permission | a vendor describing itself to every AS on the internet |
| Unknown `grant_types` | reject | **intersect** — they weren't asking us |
| Bad `logo_uri` | reject | **drop the field** |
| Over-long `client_name` | reject | **truncate** |
| Unsafe `redirect_uri` | reject | **reject** — code-leak vector, not a capability |

RFC 7591 §3.2.1 has the AS record what *it* supports and report back what it granted.
`negotiate/1` now intersects capability metadata before the changeset; safety rules stay
shared and still refuse.

## Proof

MCPJam's **real published** document declares `urn:ietf:params:oauth:grant-type:device_code`
and lists 14 redirect URIs. Against pre-fix code it fails twice over:

```
grant_types:   "contains an unsupported grant_type"
redirect_uris: "should have at most 10 item(s)"
```

Both are now fixtures. And the MCPJam OAuth conformance CLI, run against staging (still on
pre-fix `main`), reproduces the outage end-to-end:

```
==> dcr  reached the consent gate cleanly (10 steps passed)
==> cimd REGRESSION: received_authorization_code [HTTP 400]
```

DCR green, CIMD red — independent confirmation that DCR was never the broken part.

## Also in here

- **Diagnosability.** `:invalid_document` collapsed a changeset refusal and a lost insert
  race and discarded the errors. Now `{:invalid_document, errors}` vs `:store_conflict`,
  logged as `fields=[...]` — names only, never messages (several interpolate the
  attacker-supplied URI).
- **Transient vs terminal.** Every CIMD failure reported `invalid_client`, which is
  terminal. Worse, `:rate_limited` is deliberately never logged, so after the first store
  failed every retry re-entered discovery and burned the per-host bucket silently: one log
  line, many failures. Transient reasons now yield `temporarily_unavailable` + 503.
- **Per-path redirect caps** — 10 for DCR, 50 for CIMD.
- `scripts/mcp-conformance.sh` + a daily `mcp-conformance` job in `cron.yml`.
- `docs/context/cimd-vs-dcr-validation-policy.md`.

## Why the suite didn't catch it

The CIMD fixture was three fields long and authored alongside the validator it tested, so
it could only assert that our code accepts what our code expects. Every strict rule sat on
fields no test populated. Both new fixtures are copied verbatim from the wire.

## Reviewer notes

- The Claude Code fixture is a **canary, not a repro** — it passes pre-fix too.
- Full-flow conformance stops at consent (needs a Clerk session). Everything this protects
  happens before that, so it is still a real signal.
- `mcp-conformance` lives in `cron.yml`, not `verify.yml`: it grades a deployment, and on a
  PR the deployment still runs `main`.

Refs #1167

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -24,6 +24,7 @@ on:
     - cron: '17 13 * * 1'   # frontend-major-updates   (Mon 13:17 UTC)
     - cron: '0 14 * * 1'    # plan-triage              (Mon 14:00 UTC)
     - cron: '0 3 * * 6'     # dependency-submission    (Sat 03:00 UTC)
+    - cron: '40 5 * * *'    # mcp-conformance          (daily 05:40 UTC)
   workflow_dispatch:
     inputs:
       task:
@@ -38,10 +39,15 @@ on:
           - frontend-major-updates
           - runner-diagnostics
           - dependency-submission
+          - mcp-conformance
       force:
         description: '(obsidian-update only) Force re-download even if version matches'
         type: boolean
         default: false
+      target-url:
+        description: '(mcp-conformance only) MCP endpoint to test'
+        type: string
+        default: 'https://staging.engram.page/api/mcp'
 
 # No workflow-level permissions — each job grants what it actually needs.
 permissions: {}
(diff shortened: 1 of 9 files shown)


Rasbandit:
Updated to `fa775bbe` after a review pass. Three changes on top of the original push:

**1. `client_name` is now enforced.** MCP 2025-11-25 lists `client_id`, `client_name` and
`redirect_uris` as the properties a metadata document MUST carry — we were enforcing two of
three. New `:missing_client_name` reason covering `nil`, `""`, whitespace-only and
non-strings.

This one sits with the *refusals* rather than in `negotiate/1`, which is a deliberate line
given what this PR is about: an unknown `grant_type` costs nothing to ignore, but a nameless
client cannot be rendered on the consent screen at all, and "Authorize this app" is a worse
outcome for the user than a legible refusal. Both real-world fixtures supply it.

**2. `scripts/mcp-conformance.sh` no longer consults the CLI's exit code.** It cannot be
the verdict — the CLI exits non-zero unless the *whole* flow completes, which is impossible
against a Clerk-gated consent screen. The earlier structure had a `then` branch that could
never fire and, in its first revision, reported a clean pass against a server we knew was
broken. Now `|| true`, and the status-based classifier is the sole verdict, running every
time. There is no longer a path where the verdict can be skipped.

**3. Preflight, and three-valued exit codes.** The job runs on an isolated runner and
reaches both npmjs.org and (for staging) a Cloudflare hairpin back into the same host.
Neither leg is testable until this merges, since `workflow_dispatch` inputs only exist on
the default branch. So instead of pretending otherwise, the failure modes are now legible:

```
0  clean
1  conformance regression   -> the authorization server is wrong
2  environment              -> npm or the target unreachable; NOT a result
```

Verified live, both directions:

```
==> preflight
    npm and https://staging.engram.page/api/mcp reachable
==> dcr   reached the consent gate cleanly (10 steps passed)
==> cimd  REGRESSION — received_authorization_code [HTTP 400]     EXIT: 1

==> preflight
    ENVIRONMENT: cannot reach https://nonexistent-....invalid/api/mcp   EXIT: 2
```

Gauntlet on `fa775bbe`: **4073 tests / 0 failures** (`--warnings-as-errors`), credo strict
clean, dialyzer passed.

Audit follow-ups filed: #1242 (RFC 8707 audience binding), #1243 (same-origin CIMD
redirects, with the reasoning for deferring it here), plus a comment on #1208 with the MCP
consent-UI requirement.

**Post-merge, before promoting to prod:** `gh workflow run cron.yml -f task=mcp-conformance`
— CIMD should flip red → green. That is the other half of the proof and the first real
execution of the job.

PR #1256, opened 2026-08-05
fix(deps): ceiling req at 0.6.x — untested against the ex_aws S3 adapter

Rasbandit:
> **Note:** an earlier version of this description claimed req 0.7.1 *caused* six attachment-test failures. That was a misdiagnosis — see [this comment](https://github.com/engram-app/Engram/pull/1256#issuecomment-5195399071). The cluster was CI storage (MinIO) failing, and it reproduces on main at req 0.6.3. This PR now stands on the semver argument alone, which is independent and sufficient.

## The gap

`config/config.exs:249` routes every S3/KMS call through req:

```elixir
config :ex_aws, :http_client, ExAws.Request.Req
```

Our constraint was `{:req, "~> 0.5"}`, which **admits 0.7.x**. req 0.7.0 ships three documented `(BREAKING CHANGE)` entries against exactly the step API that adapter is built on — it removed `Req.Request.current_request_steps` and replaced the `run_finch` and `put_plug`/`run_plug` steps with adapter modules.

Nothing upstream closes that gap either: ex_aws 2.7.0 declares `{:req, "~> 0.5.10 or ~> 0.6 or ~> 1.0"}`, and a **two-component** `~> 0.6` means `>= 0.6.0 and < 1.0.0` — so ex_aws's own constraint admits 0.7.x too. Hex resolves it happily.

Dependabot duly proposed `req 0.6.3 → 0.7.1` as a **minor** (correct semver; 0.x minors are minors) and it resolved.

## The fix, and why three components

`{:req, "~> 0.6.0"}` — `>= 0.6.0 and < 0.7.0`.

I got this wrong on the first attempt and the check caught it: `~> 0.6` still let `mix deps.update req` jump to **0.7.2**; `~> 0.6.0` holds at **0.6.3**. Both runs are recorded in the commit rather than asserted from memory.

## Scope

`mix.lock` **untouched** — main already resolves req 0.6.3. No compiled change, no lock churn.

The ceiling lifts when ex_aws ships a release tested against req 0.7; the inline comment says so explicitly.

## Companion

engram-app/engram-infra#909 stops auto-merge arming any PR carrying a 0.x **minor**, since a 0.x minor is breaking by convention but a "minor" to semver. **13 deps in `mix.exs` sit on `~> 0.x`.**


Rasbandit:
## ⚠️ Correction: the causal claim in this PR's description is **wrong**

I attributed the 6 attachment-test failures on #1252 to `req 0.7.1` breaking the ex_aws S3 adapter. **That was a misdiagnosis.**

The same failure cluster subsequently failed on **main at req 0.6.3** (run `31024858500`), alongside:

```
test_40_storage_endpoint::test_storage_endpoint   FAILED
POST /api/attachments → 502          ← the upload fails too
CI_MINIO_CONTAINER: engram-ci-<run>-obsidian-minio-1
```

So it's the CI stack's **MinIO being down**, not req. My "asymmetric damage — PutObject fine, GetObject broken" reading came from a single test's trace in a single run; `POST` 502s as well.

The check I skipped: *before blaming a branch's diff for a whole-category failure, confirm that category passes on main in the same window.* One green main run an hour earlier isn't that check.

## What this PR still stands on

The semver argument alone, which is sufficient and independent:

- `~> 0.5` genuinely admitted 0.7.x, and the **three-component** `~> 0.6.0` is what actually holds the line (`~> 0.6` means `< 1.0.0` — verified, it let `deps.update` jump to 0.7.2).
- req 0.7.0 genuinely ships three `(BREAKING CHANGE)` entries against the step API `ExAws.Request.Req` is built on.

So: **precaution against an untested-together pairing, not a fix for a diagnosed incident.** Merging is still right; the justification is narrower than I wrote.

I'll amend the PR body to match. Consequence for #1252: it was probably innocent of this, and my disarming its auto-merge rested on a wrong reason — though the 0.x guard (engram-infra#909) holds it anyway for `tidewave 0.6.1 → 0.8.0`, which is a separate and still-valid reason.

The real, still-open issue this was masking: **CI storage is failing on main.**

PR #1279, opened 2026-08-05
fix(ci): CI_MINIO_ACCESS_KEY is a plain value, not a secret

Rasbandit:
## main is red

```
error while interpolating services.minio-init.environment.CI_MINIO_ACCESS_KEY:
required variable CI_MINIO_ACCESS_KEY is missing a value
```

`prebuild-ci-image` runs `docker compose -f ci/compose.yml build engram`. Compose interpolates the **entire** file before building anything, so the `:?` required-guards #1267 added to the `minio-init`/`engram` runtime env abort a build that never reads those values.

## Why no secret fixes it

`${{ secrets.CI_MINIO_ACCESS_KEY }}` resolved to empty in **all four** places `verify.yml` referenced it — because that secret does not exist, and must not.

engram-infra's SOPS+Terraform commit records the reasoning, learned the hard way:

> Only the SECRET key is managed here. The access key is the username `engram-ci` and stays a plain value in Engram's verify.yml: its value is also the CI image repository name (`${LOCAL_REGISTRY}/engram-ci:<hash>`), and GitHub redacts secret values from job OUTPUTS by substring — **as a secret it blanked `ENGRAM_CI_IMAGE` for every downstream e2e job and failed them with `invalid reference format`**.

So the access key is deliberately plain. Only `CI_MINIO_SECRET_KEY` is a secret, and it already exists (set 2026-08-05 20:37).

## Fix

- Set `CI_MINIO_ACCESS_KEY: "engram-ci"` once at workflow level, with the redaction rationale recorded inline so nobody "fixes" it back into a secret
- Drop the four per-job `${{ secrets.CI_MINIO_ACCESS_KEY }}` overrides that all resolved to empty
- Give `prebuild-ci-image` the `CI_MINIO_SECRET_KEY` half it was never given when #1267 landed

## Scope is wider than the prebuild job

`e2e-obsidian`, `e2e-crdt` and `headless-protocol` were passing an **empty** access key into their MinIO bucket create/reap steps. Same root cause, one layer further in — they would have failed on MinIO auth even after the prebuild job was unblocked.

## Why it did not break at the #1267 merge

`prebuild-ci-image` short-circuits when the image is already in the registry for that tree hash; the #1267 merge hit that path. The first run that actually had to rebuild is the one that surfaced it — `8b539e76` — and from there it is every PR touching `lib/`, `config/`, `frontend/`, `mix.exs` or the Dockerfile.

```
2026-08-05T22:05  8b539e76  failure   <- first rebuild after #1267
2026-08-05T21:52  636c4f9e  success
2026-08-05T21:14  28d50890  success   <- the #1267 merge
```

## Deliberately not done

**Not** defaulting the compose guards away. #1267 added them so a forgotten `CI_MINIO_BUCKET` cannot write into a shared bucket; softening a runtime protection to fix a build-time parse error is the wrong trade.

**Known wart left alone:** the retry loop reports this as `build attempt N failed (daemon reclaimed?)` and burns all three attempts in under a second, because a compose interpolation error is deterministic. That misattribution is what makes this read as an infra flake. Teaching it to separate config errors from daemon errors is a separate change.

## Verification

- `yaml.safe_load` parses; workflow-level `CI_MINIO_ACCESS_KEY` resolves to `engram-ci`
- This PR exercises the fixed path on its own run — same-repo PRs use the head branch's workflow, so a green `prebuild-ci-image` here *is* the test

Refs #1267

https://claude.ai/code/session_01X6KUbXWuDMGhHbz8uDPujU

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -98,6 +98,15 @@ env:
   # pull hex from the LAN instead of WAN. Signatures still verified client-side.
   HEX_MIRROR: "http://10.0.20.214:8090"
 
+  # MinIO username for the central FastRaid instance, scoped by policy to
+  # `ci-*` buckets (engram-infra: SOPS + main/github/secrets.tf). A PLAIN value,
+  # deliberately not a secret: it is also the CI image repository name
+  # (`${LOCAL_REGISTRY}/engram-ci:<hash>`), and GitHub redacts secret values out
+  # of job OUTPUTS by substring — as a secret it blanked ENGRAM_CI_IMAGE for
+  # every downstream e2e job, which then failed with `invalid reference format`.
+  # Only the paired CI_MINIO_SECRET_KEY is a secret.
+  CI_MINIO_ACCESS_KEY: "engram-ci"
+
   # CI-only encryption master key. Ephemeral — CI databases are recreated per run,
   # so the key does not protect persistent data. Production deploys must override
   # with a secret via `fly secrets set ENCRYPTION_MASTER_KEY=...`.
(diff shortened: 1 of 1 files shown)


Rasbandit:
Superseded by #1272, which landed on main while this was in review and fixes the same root cause more completely.

#1272 lifts **both** `CI_MINIO_ACCESS_KEY: engram-ci` and `CI_MINIO_SECRET_KEY` to workflow level (verify.yml:124-125), so `prebuild-ci-image` inherits them without needing the per-step `env:` this PR added — and it collapses the per-run buckets to one permanent `ci-e2e`, removing the reap steps entirely.

Same conclusion on the substantive point, independently reached: the access key is a plain value, never a secret, because GitHub's substring redaction blanks `engram-ci` inside `ENGRAM_CI_IMAGE`.

Verified main is genuinely fixed rather than assuming it:
```
$ git show origin/main:.github/workflows/verify.yml | grep -nE '^\s+CI_MINIO'
124:  CI_MINIO_ACCESS_KEY: engram-ci
125:  CI_MINIO_SECRET_KEY: ${{ secrets.CI_MINIO_SECRET_KEY }}
128:  CI_MINIO_BUCKET: ci-e2e
```

Closing rather than rebasing — nothing here adds anything main does not already have. This branch's run went fully green (including e2e-clerk), which is at least independent confirmation the approach is right.

PR #1294, opened 2026-08-06
ci: cover prod in the nightly conformance, and stop reporting config errors as flakes

Rasbandit:
Two CI-truthfulness fixes, one commit each so they're separable.

## 1. Prod has no recurring conformance coverage

`mcp_rewrite_host?/1` reads `:host_rewrite`, which **only prod sets**. So the bare-root `resource`, its derived `resource_metadata_url`, and the MCP-host `/oauth` passthrough are `false` on staging, on the CI stack, and in the unit suite. Their only verification to date was one manual run during the 0.14.0 deploy.

That is precisely the shape of the bug 0.14.0 shipped to fix: **prod-only code, no recurring coverage, green everywhere anyone could look.**

Adds prod as a second matrix leg:

| leg | target | stages |
|---|---|---|
| staging | `https://staging.engram.page/api/mcp` | `spec,oauth` |
| prod | `https://mcp.engram.page` (bare) | `spec` |

**Prod is `spec`-only deliberately.** The OAuth matrix performs real DCR and CIMD registrations — that would write throwaway client rows into the production database every night. `spec` is unauthenticated curl and writes nothing. It dials the **bare** host because that's what the docs tell users to paste, and what the prod-only branch actually produces.

`fail-fast: false` — independent deployments; one being red says nothing about the other.

### Alerting

The tracking issue stays the durable record (and now gets a distinct title per host, since it's derived from the target). The Discord post is the push notification.

Worth having because **a prod OAuth regression is silent from outside**: existing clients keep working on their issued tokens, and only *new* connections fail. Nothing else surfaces it.

The alert reports deploy drift inline, because a stale deployment failing these assertions is the most common false alarm and otherwise sends the reader hunting through OAuth code.

Guarded on `env.WEBHOOK != ''` — the idiom `deploy-prod.yml` already uses — so it **no-ops until the webhook consolidation lands** rather than blocking on it. Secret name `DISCORD_ALERT_WEBHOOK_URL`, matching the intended `announce`/`alert` scheme.

### One matrix gotcha handled

Artifact names gain the matrix label. `upload-artifact@v4+` **errors** on a duplicate name instead of merging, so a bare `mcp-conformance-results` would have failed whichever leg finished second — losing the evidence for the leg that may be the one that failed.

## 2. The retry loop blamed the daemon for config errors

```
::warning::build attempt 1 failed (daemon reclaimed?) — retrying in 15s
```

A compose interpolation error is deterministic — re-running re-parses the same file and fails identically — so all three attempts burned in **under a second** while the warning pointed at infrastructure.

That misattribution has a measured cost: it's why the missing `CI_MINIO_ACCESS_KEY` read as "the CI registry is down" and took **three separate PRs** to actually fix. The failure text said daemon, so nobody read the parse error two lines above it.

Now it classifies before retrying. Config error → fail immediately and say it's deterministic. Transient → retry exactly as before.

Reads the build's real status out of `PIPESTATUS` rather than setting `pipefail` mid-block: GitHub's default shell is `bash -e` **without** pipefail, so `build | tee` would otherwise report `tee`'s status and every failure would look like a pass.

### Verified both directions against real captured output

```
error while interpolating ... required variable CI_MINIO_ACCESS_KEY is missing a value
  -> CONFIG (fail fast)

#12 ERROR: failed to solve: rpc error: code = Unavailable ... graceful_stop
  -> transient (retry)
```

## Follow-up this sets up

Webhook consolidation onto `announce` + `alert`. Current sprawl is small: `DISCORD_RELEASES_WEBHOOK_URL` (4 refs in engram, 1 in Engram-obsidian) and `DISCORD_DEPLOY_WEBHOOK_URL` (**orphaned — referenced by no workflow in either repo**), all per-repo, none org-level.

https://claude.ai/code/session_01X6KUbXWuDMGhHbz8uDPujU

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -565,13 +565,41 @@ jobs:
     if: |
       github.event.schedule == '40 5 * * *'
       || github.event.inputs.task == 'mcp-conformance'
-    name: mcp-conformance
+    name: mcp-conformance (${{ matrix.label }})
     runs-on: [self-hosted, linux, x64, isolated, light]
     # 8 CLI runs (2 strategies x 4 protocol versions) plus the spec assertions.
     timeout-minutes: 25
     permissions:
       contents: read
       issues: write   # regressions open/update a tracking issue (see last step)
+    strategy:
+      # Independent deployments — one being red says nothing about the other,
+      # and finding out about both in one run is the point.
+      fail-fast: false
+      matrix:
+        include:
+          # Staging carries the full matrix: the OAuth stage performs real DCR
+          # and CIMD registrations, which write client rows, so it belongs on a
+          # deployment we are happy to dirty.
+          - label: staging
+            target: https://staging.engram.page/api/mcp
+            stages: spec,oauth
+
+          # Prod runs the spec stage ONLY — it is plain unauthenticated curl, so
+          # it writes nothing. Not thoroughness-limited by accident: the OAuth
+          # matrix would register throwaway clients in the production database
+          # every night.
+          #
+          # Prod is not redundant with staging, it is the only coverage that
+          # exists for a whole branch of the code. `ENGRAM_HOST_REWRITE_ENABLED`
+          # is prod-only, so `mcp_rewrite_host?/1` is FALSE everywhere else and
+          # the bare-root `resource`, its `resource_metadata_url`, and the
+          # MCP-host `/oauth` passthrough are unexercised by staging, by the CI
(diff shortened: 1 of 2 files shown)


Rasbandit:
Superseded — this was opened on a `ci/**` branch, which is not in `verify.yml`'s push-trigger allowlist, so it ran zero jobs and `gh pr checks` reported a vacuous green. Reopened on `chore/**` with the same two commits.

Issue #1303, opened 2026-08-07
flaky: crdt_head_property_test fails ~1-in-N under full-suite load, passes in isolation


Rasbandit:
`test/engram/notes/crdt_head_property_test.exs` ("served head reflects every acked write across edit/checkpoint/terminate/recreate") failed once during a local full `mix test` run (4202 tests, 1 failure). It then passed via `mix test --failed`, 3/3 in isolation, and the next full unpiped run was 4202/0.

**Not characterized — the failure text was lost.** The run was piped through `tail -30`, which truncated the diagnostic and (per our own rule) also returned tail's exit code, so the run reported success. Filing so the observation isn't lost rather than pretending it didn't happen.

## What's known
- Only reproduces under full-suite parallelism (`max_cases: 20`), never alone.
- The same run logged `Postgrex.Protocol (#PID<0.549.0>) disconnected: ** (DBConnection.ConnectionError) client #PID<0.85726.0> exited` alongside `CrdtCheckpointTimer could not fetch doc ... reason={:exit, {:normal, {GenServer, :call, [#PID, :get_doc, 5000]}}}` — consistent with the sandbox-connection-ownership class in `docs/context/channel-parallelism-db-pool.md`, but unconfirmed.
- The property is `async: false` and randomized (`max_runs: 200` locally, 50 in CI) with no pinned seed, so a failure is not reproducible from the summary alone.

## Next step when someone picks this up
Run the full suite in a loop with output captured to a file (**not** piped) until it reproduces, then keep the shrunk command sequence and the seed. Consider having the property print its seed on failure.

Rasbandit:
**Characterized.** Reproduced during the full-suite run for #1301 (PR #1305), with output captured to a file rather than piped, so the diagnostic survived this time.

## It is a timeout, not an assertion failure

```
3) property served head reflects every acked write across edit/checkpoint/terminate/recreate
   ** (ExUnit.TimeoutError) property timed out after 60000ms
   stacktrace:
     Engram.Repo.run_with_tenant/2
     Engram.Notes.maybe_broadcast_vault_populated/2
     Engram.Notes.upsert_note/4
     test/engram/notes/crdt_head_property_test.exs:59
```

[...]

Issue #1302, opened 2026-08-07
Links parser is wikilink-only: [text](path.md) links are never graphed or rewritten


Rasbandit:
`Engram.Links.Parser` (`lib/engram/links/parser.ex:14`) matches only `!?[[target]]`. Obsidian users with **Settings → Files & Links → Use [[Wikilinks]] = off** write markdown links (`[text](Note.md)`, `![alt](img/pic.png)`) instead. For those vaults:

1. **No graph edges** — `note_links` rows are never created, so backlinks / link graph / dangling-bind are empty. (Pre-existing, independent of rename.)
2. **No rename rewrite** — a web/REST/MCP-origin rename leaves every markdown link pointing at the old path. Obsidian rewrites these natively, so plugin-origin renames are fine; only server-origin renames rot.

## Scope
- `Parser.extract/1` — add a markdown-link arm. Must skip external targets (`http:`, `https:`, `mailto:`, protocol-relative), bare-anchor targets (`#heading`), and respect the existing code-fence/inline-code/frontmatter exclusion ranges. Percent-decode the target for basename keying.
- `NoteLink.link_type` — new values (e.g. `md_link` / `md_embed`); string column, no migration.
- `Rewriter.replacement_target/2` — emit markdown form for markdown occurrences (percent-encode spaces; keep `!` and the `[text]` label verbatim). The existing form rule — bare stays bare, qualified stays qualified — still applies to the path segment.
- Backfill existing notes (`Engram.Workers.BackfillNoteLinks`) so graphs fill in for current vaults.

## Blast radius
`Parser.extract/1` also feeds `Engram.Indexing`, `BackfillNoteLinks`, and `ExtractNoteLinks` — this widens the edge set everywhere, which is the point, but it is not a rename-only change.

Refs #648

Issue #1301, opened 2026-08-07
Flip @untagged_crdt_client_type to "web" (one-rewriter default)


Rasbandit:
Follow-up to #648 Phase 2.

`Engram.Notes.@untagged_crdt_client_type` is pinned to `"obsidian"`, so a CRDT client that joins **without** a `client_type` param gets **no** server-side link rewrite on rename. That was the right compromise while version-skewed plugins (pre-`client_type`, i.e. < 1.20.0) were the majority — flipping early would have double-rewritten for them (server + Obsidian both), which is the exact failure the one-rewriter invariant exists to prevent.

Now it's the wrong default: any *non*-Obsidian client that omits the tag (an old web build, a future mobile client) silently rots links with no error surface.

**Do this once plugin 1.20.x has circulated ~1 release cycle** (1.20.0 shipped 2026-08-05).

## Change
- `lib/engram/notes.ex` — `@untagged_crdt_client_type "obsidian"` → `"web"` (one line).
- `test/engram/notes_crdt_origin_gate_test.exs:27-28` — both asserts flip together (comment in the test already says so).
- `test/engram/links/rewrite_wiring_test.exs` — "untagged relocate enqueues NOTHING" flips to `assert [_] = all_enqueued(...)`.

## Verify before flipping
Confirm plugin < 1.20.0 traffic has drained — untagged CRDT joins in Loki.

Refs #648

Issue #970, opened 2026-07-08
sync broadcasts carry no origin device_id — clients cannot filter self-echo


Rasbandit:
Every `note_changed` broadcast is `mode=fanout` to all vault subscribers **including the device that caused the change**. Clients have no origin field to filter on, so each self-echo class needs its own bespoke client-side suppression (pushing/recentlyPushed sets, and now the wipeRemote path-marking in Engram-obsidian#215 after the 2026-07-08 replace-remote vault wipe).

Owed since the delivery-oracle wave (#909, device_id attribution). Proposal: the sync channel join already authenticates a device context — stamp `device_id` (or conn_id) into `Broadcast.emit` payloads so clients can drop events they originated, and the e2e log oracle can attribute deliveries.

Refs: Engram-obsidian#215, #969 (test_86), `docs/context/e2e-delivery-flake-playbook.md`

Rasbandit:
Delete-side attribution shipped in #969 + Engram-obsidian#215: `EngramWeb.OriginDevice` reads the plugin's X-Device-Id header, notes delete broadcasts carry `device_id`, and the plugin drops self-originated deletes. Remaining for this issue: upsert/rename attribution (needs care — upsert echoes drive client id-relocation), channel-push attribution beyond `broadcast_from`, and e2e log-oracle delivery attribution.

PR #1304, opened 2026-08-07
fix(links): rewrite links on resurrect-rename, not just live relocate

Rasbandit:
## Problem

A CRDT rename reaches `move_note` down two different roads:

- `genesis_relocate_live` — the row is **live** when the `crdt_create` for its id lands at a new path
- `genesis_resurrect` — the row is a **tombstone** when it lands

Same user action (a rename), same DB primitive, two call sites — and only the live one enqueued `RewriteNoteLinks`. So a web-origin rename of a note the server had tombstoned left every `[[Old]]` reference in every other note pointing at the dead path, with no error surface. The `Rewriter` moduledoc already carried this as a known deferred gap.

## Fix

Thread `vault` + `origin` through `genesis_resurrect`/`genesis_resurrect_decrypted` and enqueue behind the same `Notes.crdt_rename_rewrites?/1` gate, so the **one-rewriter invariant** holds on both legs:

- Obsidian-origin renames stay server-silent — the plugin's own "Automatically update internal links" owns those
- Everything else (web / REST / MCP) rewrites server-side exactly once

The old path rides as AAD-bound ciphertext in the job args, same as the live leg: `move_note` repoints the row, so there's no old-path tombstone left for the worker to read.

Moduledoc updated — the gap note becomes a statement that both legs are wired.

## Tests

Three added to the existing Phase 2 `describe` block in `rewrite_wiring_test.exs`:

- web-origin resurrect-rename enqueues one job with the same encrypted old-path args as a relocate (**red before this change**, green after)
- obsidian-origin resurrect-rename enqueues nothing
- same-path resurrect (not a rename) enqueues nothing even for web origin

## Verification

```
mix format          ok
mix credo --strict  4343 mods/funs, found no issues
mix dialyzer        Total errors: 5, Skipped: 5, Unnecessary Skips: 0 — passed
mix test --warnings-as-errors
                    1 property, 4202 tests, 0 failures (8 excluded)
```

One earlier full run showed a single failure in `crdt_head_property_test`. It has zero references to `genesis_crdt_note`, `genesis_resurrect`, or `delete_note`, passed 3/3 in isolation, and the full rerun above is clean — pre-existing load-dependent flake, filed as #1303 rather than dropped.

## Context

Found while auditing how [No-Instructions/Relay](https://github.com/No-Instructions/Relay) handles rename link propagation. Relay has no rewriter at all: it delegates to Obsidian on both ends, applying remote renames with `fileManager.renameFile` (the link-*updating* API), so every client rewrites independently and diff3 collapses the duplicates. That only holds while all clients' link-format settings agree. Our model — one rewriter chosen by origin, with the plugin applying remote renames via `vault.rename` (which never touches links) — is the stronger position; this PR closes the one leg where "exactly one" was actually "zero".

Related follow-ups filed, not in this PR: #1301 (flip the untagged `client_type` default once 1.20.x has circulated), #1302 (parser is wikilink-only, `[text](path.md)` never graphed or rewritten).

Refs #648

### DIFF
--- a/lib/engram/links/rewriter.ex
+++ b/lib/engram/links/rewriter.ex
@@ -7,10 +7,10 @@ defmodule Engram.Links.Rewriter do
   `[[wikilink]]`/`![[embed]]` targets are rewritten to the new name. The
   server is the SINGLE rewriter for every origin EXCEPT plugin-origin
   renames (Obsidian's "Automatically update internal links" owns those) —
-  exactly one party rewrites, never both. Known deliberate gap:
-  `genesis_resurrect`'s rename-restore leg (web-origin resurrect-rename)
-  currently enqueues no rewriter — deferred, tracked for the
-  untagged-default flip follow-up.
+  exactly one party rewrites, never both. Both CRDT rename legs are wired:
+  `genesis_relocate_live` (live row) and `genesis_resurrect`'s
+  rename-restore leg (tombstoned row), each behind
+  `Notes.crdt_rename_rewrites?/1`.
 
   POLICY: the server authors only mechanical, semantics-preserving
   transforms — the rewritten target resolves to the same row the old text
--- a/lib/engram/notes.ex
+++ b/lib/engram/notes.ex
@@ -754,7 +754,7 @@ defmodule Engram.Notes do
                  end
 
                {:tombstone, %Note{} = prior} ->
-                 genesis_resurrect(prior, user, sanitized_path, folder)
+                 genesis_resurrect(prior, user, vault, sanitized_path, folder, origin)
 
                :none ->
                  genesis_adopt_or_insert(user, vault, canonical_id, sanitized_path, folder)
@@ -1056,10 +1056,10 @@ defmodule Engram.Notes do
     end
   end
 
-  defp genesis_resurrect(prior, user, sanitized_path, folder) do
+  defp genesis_resurrect(prior, user, vault, sanitized_path, folder, origin) do
     case Crypto.maybe_decrypt_note_fields(prior, user) do
       {:ok, prior} ->
-        genesis_resurrect_decrypted(prior, user, sanitized_path, folder)
+        genesis_resurrect_decrypted(prior, user, vault, sanitized_path, folder, origin)
 
       {:error, reason} ->
         log_resurrect_decrypt_failure(reason, user, prior)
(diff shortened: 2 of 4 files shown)


Rasbandit:
### Code review

Found 3 issues. All three are fixed in 90df54b7b48a4850452432db9aae5a52d6100766 on this branch.

1. `RewriteNoteLinks` moduledoc enumerated "Four enqueue origins" and named only `genesis_relocate_live` as the CRDT leg. This PR adds a fifth call site (`genesis_resurrect_decrypted`), so the enumeration went stale — at the same spot where a #1240 reviewer already caught this once ("the `RewriteNoteLinks` moduledoc still says 'REST/MCP origins only' though phases 2-3 added CRDT and folder enqueue sites").

https://github.com/engram-app/Engram/blob/90df54b7b48a4850452432db9aae5a52d6100766/lib/engram/workers/rewrite_note_links.ex#L4-L14

2. `crdt_rename_rewrites?/1`'s `@doc` scoped the gate to "relocates reached via `genesis_crdt_note/5`". After this PR the same gate also guards the resurrect-rename leg, which is a different function — the doc described a narrower scope than the code.

https://github.com/engram-app/Engram/blob/90df54b7b48a4850452432db9aae5a52d6100766/lib/engram/notes.ex#L919-L927

3. The test "same-path resurrect (not a rename) enqueues nothing even for web origin" asserted nothing. The delete-wins guard (`recent_same_path_tombstone?`, FIX 1 / #970) returns `{:error, :recently_deleted}` *before* `renamed?` is ever computed, so `all_enqueued(...) == []` passed without the gate being reached — it proved the guard, not the `renamed? == false` half it claimed to cover. Now backdates past the window so the resurrect actually succeeds (asserted via `{:ok, resurrected}`), with the in-window refusal split into its own test.

https://github.com/engram-app/Engram/blob/90df54b7b48a4850452432db9aae5a52d6100766/test/engram/links/rewrite_wiring_test.exs#L168-L192

Checked and found clean: `decrypted.id` / `prior.path` correctness, transaction placement and the never-raise `get_dek` discipline (inherited by reusing the shared `enqueue_crdt_rename_rewrite/4` rather than reimplementing it), the FIX 1 / FIX 7 interaction, the tombstone-vs-ciphertext old-path recovery split, and AGENTS.md compliance.

Note: #1305 flipped `@untagged_crdt_client_type` to `"web"` while this branch was open; it came in via the branch sync and both gate tests merged cleanly.

PR #1308, opened 2026-08-07
refactor(web): remove the onboarding tour and its demo vault

Rasbandit:
## Summary

- Removes the guided onboarding tour (react-joyride) from the web SPA, along with the synthetic "demo vault" it rendered through the real query hooks. Deletes `src/onboarding/tour/`, `public/demo-vault.json`, the `react-joyride` dependency, every `data-tour` anchor, and the `engram:vault-switched` / `engram:note-opened` DOM events that existed only to advance gated tour steps.
- Un-forks the API layer: `useFolders`, `useFolderNotes`, `useAttachments`, and `useVaults` each carried an `enabled: !demo?.active` + `if (demo?.active) return {...query, data: fake}` shim so the tour could serve fixtures through production hooks. They are plain `useQuery` again. `api/active-vault.ts` loses its demo-id special cases, including `resetActiveVaultToStored`, which had no caller outside the tour.
- Drops the "Take the tour" checklist row (the rest of the checklist, the first-vault modal, and `OnboardingShell` all stay) and the `dismissed:tour` seeds in the e2e helpers.

Net: 44 files, −1117 lines.

### On the localStorage heal

`active-vault.ts` carried a guard that dropped a persisted `demo-vault-*` id on read — an early tour build poisoned storage, and those users 404'd every vault-scoped request with no self-recovery. Removing the guard does **not** re-open that hole: without `isDemoVaultId`, a stale demo id is simply an id the account does not own, so `reconcileActiveVault` (already called from the bootstrap seed and the `/vaults` fetch) re-points it at a real vault. Same recovery, one fewer special case. The tour-specific test in `active-vault.test.ts` is replaced with one asserting exactly that path.

### Deliberately out of scope

The backend onboarding actions `tour_offered_taken`, `tour_offered_skipped`, and `tour_completed` are left in `lib/engram/onboarding/action.ex`. Nothing sends them now, but existing rows stay valid and there is no migration risk. `docs/context/demo-vault-activevaultid-poisoning.md` is marked OBSOLETE rather than deleted — the fix it documents is gone, but the bug class (client-only sentinel ids reaching persisted state) is worth keeping.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `biome check --error-on-warnings .` clean
- [x] `bun run test` — 1304 tests across 174 files pass (baseline was 1324/176: 2 deleted files, 20 removed tour tests)
- [x] `bun run build:selfhost` succeeds
- [x] `mix format --check-formatted` on the one Elixir comment touched; pre-push hook (compile + credo + sobelow) green
- [ ] **CI**: the Playwright specs in `e2e/` were edited but cannot be exercised locally (they need the Clerk stack) — `e2e-clerk` / `e2e-browser` are the proof. Reviewer: confirm those jobs are green before merge.
- [ ] Manual smoke: sign in, confirm the dashboard renders with no tour prompt, the checklist shows only vault/plugin/tool rows, and the upload affordance in the files panel still works.

https://claude.ai/code/session_01JyirNUaZY19uVvzwBL8yX4


Rasbandit:
## Self-review pass

Reviewed the diff adversarially rather than confirmationally. Findings below; two fixed in 69fbb9cb, one accepted-and-documented.

### Verified: the heal claim actually holds

The load-bearing claim in the description is that dropping `isDemoVaultId` doesn't strand users whose localStorage still holds a `demo-vault-*` id, because `reconcileActiveVault` re-points it. That only holds if the request that *triggers* the heal isn't itself vault-scoped — otherwise the malformed header 404s the very call that would fix it, which is the original bug.

Checked the router: `/api/bootstrap` (`router.ex:313`) and `/api/vaults` (`router.ex:262`) both sit in scopes without `VaultPlug`; the vault-scoped scope begins at `router.ex:363`. So the stale header cannot block them. `reconcileActiveVault` runs from both the bootstrap seed and the `/vaults` queryFn, re-points, and `setActiveVaultId` persists the corrected id. Confirmed.

### Fixed — `record()` was left dead

`useOnboardingActions` returned both `record` (`mutate`) and `recordAsync` (`mutateAsync`). The tour's exit handler was `record`'s only caller. `recordAsync` still backs the checklist dismiss. Removed `record` from the hook and the two test mocks that mirrored its shape — `noUnusedLocals` does not catch unused object members, so this would have rotted silently.

### Fixed — index pointed at a doc describing deleted guards

`AGENTS.md:370` routed `demo-vault-activevaultid-poisoning.md` as "(fixed)". The fix it describes (`isDemoVaultId`, `resetActiveVaultToStored`, the `readStored()` heal) no longer exists, so the pointer sent readers to guards they would not find. Re-routed by bug class instead.

### Accepted risk — brief 404 window for a dormant population

Previously `readStored()` cleared a poisoned id **synchronously at module init**, so the first request already carried no bad header. Now the heal lands one round-trip later, so vault-scoped queries mounting concurrently with bootstrap can 404 once before recovering.

Not worth a guard: recovery is automatic (`useFolders`/`useFolderNotes`/`useAttachments` key on `["...", vaultId]`, so reconcile changes the key and refetches), and the affected population is users who ran the tour **before** 2026-07-05 **and** have not opened the app since — any load after that date already healed and cleared the key.

### Gaps in the local verification

- `tsconfig.json` is `"include": ["src"]`, so `tsc --noEmit` does **not** cover `e2e/`. My six Playwright edits are type-unchecked locally — biome lints them, but that catches syntax and unused bindings, not types. CI is the only proof.
- Re-read the e2e diff by hand as compensation: all six are surgical block deletions with no orphaned references (`auth`, `token`, `jwt` all still consumed by later calls in each helper).
- Removing the `dismissed:tour` seeds does not change what renders in e2e. Those users have `uses_obsidian: true, tools: ["claude"]`, so the checklist still mounts with its obsidian + claude rows exactly as before — the tour row was merely hidden, never the reason the widget was on screen.

### User-visible behaviour change worth naming

A user with `uses_obsidian: false`, no tools selected, and a vault already created now has **zero** checklist rows, so the widget unmounts entirely. Previously the standing tour row kept it alive. Intended, but it is a real change to what that cohort sees.

### Re-verified after the follow-up commit

`tsc --noEmit` clean · `biome check --error-on-warnings` clean · 174 files / 1304 tests pass · `build:selfhost` green.

PR #1309, opened 2026-08-07
feat(links): parse and rewrite markdown-syntax links

### LINKED ISSUE #1302: Links parser is wikilink-only: [text](path.md) links are never graphed or rewritten

Rasbandit:
## Problem

Obsidian writes links two ways, and the toggle is **Settings -> Files & Links -> "Use [[Wikilinks]]"**:

```
on  (default)   [[My Note]]
off             [My Note](My%20Note.md)
```

Wikilinks are an Obsidian invention; `[text](file.md)` is actual markdown. People turn wikilinks off for portability — so their notes render on GitHub, in other editors, in static site generators.

`Engram.Links.Parser` was a single regex looking for `[[ ]]`, so for those vaults every link read as ordinary prose. Two consequences, both silent:

1. **Backlinks were empty.** No `note_links` rows were ever written, and that table is user-facing — `GET /notes/by-id/:id/backlinks` and the `links` field on note detail both read it.
2. **Server-origin renames broke their links.** The rewriter walks occurrences the parser found, which was none, so `[My Note](My%20Note.md)` kept pointing at the dead path.

Plugin-origin renames were always fine — Obsidian rewrites markdown links natively, same as wikilinks — so the damage was confined to renames starting on our side (web / REST / MCP). No error, no failed request. Backlinks just look like you have none.

## Approach

**Parse both, rewrite each in its own syntax, author neither differently.**

`Parser.extract/1` gains a second arm for `[label](target)` / `alt`, handling angle-bracket destinations (`[x](<My Note.md>)`), trailing titles (`[x](Note.md "Title")`), `#anchors`, and percent-decoding — and skipping anything that isn't a vault path (`https:`, `mailto:`, protocol-relative `//cdn`, bare `#heading`, empty).

Each occurrence now carries two new fields:

- `form` — `:wiki | :markdown`
- `target_raw` — the literal bytes at `target_start`/`target_len`

The load-bearing decision: **`target` stays the DECODED path** (`My%20Note.md` -> `My Note.md`) so `basename_key/1` and `resolve_target/4` behave identically for both syntaxes, **while the span stays RAW**. That is why the rewriter needed exactly one behavioral change — percent-encode the replacement when the occurrence was markdown. Everything else falls out: `[x](My%20Old.md)` splices correctly because the edit is "replace these 11 bytes", and 11 is the encoded length even though resolution used the decoded string. 8 of the 9 new rewriter tests passed before I touched `Rewriter`.

A markdown link comes back as a markdown link. The server never converts one syntax to the other — rewriting someone's file into a syntax they deliberately turned off would be its own bug.

## What this does NOT do

**Authoring stays wikilink-only.** Our web app hardcodes `[[...]]` (`note-page.tsx:335`, `folder-tree.tsx:529`, `wiki-completion.ts`). Matching a user's preferred syntax when we insert a link would mean knowing their Obsidian setting, which lives in Obsidian's own config — not something we sync or should. Mixed syntax in a vault is cosmetically untidy but harmless: both kinds resolve, both get rewritten on rename.

## No schema change, no migration, no new backfill

- `link_type` still means "is this an embed" — `resolve_target/4` takes `_link_type` and ignores it (`links.ex:178`), so syntax lives in `form` and the `note_links` column plus every existing consumer are untouched.
- Backfill reuses what exists: the `links` scope walks every live note with **no** edge-existence filter and `Links.replace_links/4` is delete+insert, so re-running the existing job repopulates.

**Post-deploy runbook step:**

```
docker exec engram-saas /app/bin/engram rpc 'Engram.Links.Backfill.enqueue_all()'
```

Without it, existing vaults only pick up markdown edges as notes are next written.

## Known limitation

A bare `)` inside a path (`[x](My (file).md)`) is not matched — the destination group rejects `)`. Markdown requires angle brackets or an escape there anyway, and `<...>` **is** handled. Marked with a `ponytail:` comment naming the upgrade path (balanced-paren scan) rather than pre-building it.

## Defensive addition

`Enum.uniq_by(& &1.position)` on the merged occurrence list. `note_links` has `unique_index([:source_note_id, :position])`, so two occurrences claiming the same start byte would be a 500 on ordinary user text. I could not construct a collision — a wikilink needs `]]`, which terminates the markdown label — so it is identity today. Kept because the input is arbitrary user content and that reasoning shouldn't be load-bearing.

## Tests

13 new parser tests, 9 new rewriter tests. Written before the implementation; the parser set went 13 red -> green, the rewriter set caught the one real gap (encoding a new path containing a space).

Covers: percent-encoded round trips, angle brackets, titles, anchors, externals, code-fence/inline-code/frontmatter exclusion, mixed-syntax notes keeping each occurrence's own form, multibyte offsets, malformed `%ZZ` escapes, attachment embeds, and idempotence.

## Verification

```
mix format          ok
mix credo --strict  4352 mods/funs, found no issues
mix dialyzer        Total errors: 5, Skipped: 5, Unnecessary Skips: 0 — passed
mix test --warnings-as-errors
                    1 property, 4225 tests, 0 failures (8 excluded)
```

Closes #1302


Rasbandit:
### Code review

Eight-agent review. Three findings were real defects, all fixed in 02f2d120.

1. **ReDoS in both new regexes, on the synchronous write path.** `@md_title_re` backtracked quadratically on a whitespace run — measured **67s** for `[x](Note.md` + 100KB of spaces. `@md_link_re`'s label class `[^\]]*` allowed `[`, removing the early exit the wikilink regex has: **14.2s** on a 100KB bracket run vs 37ms for `@link_re` on the same input. Both reachable by accident on a multi-MB note, not just by malice.

Fixed by deleting the title regex entirely — CommonMark ends an unbracketed destination at the first whitespace, so one `:binary.match` is linear *and* more spec-correct — and by excluding `[` from the label class.

https://github.com/engram-app/Engram/blob/02f2d120e7b7bdb0f1c34b76ae5d7b0e0a0e1d9c/lib/engram/links/parser.ex#L36-L52

2. **`[x](My(file).md)` wrote a garbage edge.** The destination truncated at the first `)`, producing target `My(file` — which passed the empty/external guards and became a real dangling `note_links` row. My own comment claimed this input was "not matched"; it was matched and silently corrupted. Destination now admits one level of balanced parens.

3. **A `%FF` escape permanently 500s a note.** `URI.decode("%FF")` yields invalid UTF-8, which was stored unscrubbed, then decrypted straight into a JSON response where `Jason.encode!` raises — so `[bad](%FF.md)` writes fine and then breaks every subsequent read of that note's backlinks. Now scrubbed at the source, the same gate `clean/1` applies to `alias`/`anchor`. The `rescue ArgumentError` around `URI.decode/1` went with it — that function re-emits a literal `%` for malformed escapes and cannot raise.

Also fixed: a trailing-space target (`[x](Note.md #anchor)` → `"Note.md "`, silently dangling), a vacuous test that passed under a deliberate double-extraction mutation, four stale moduledocs, and five OpenAPI field descriptions still describing wikilink-only behavior.

**Stated, not fixed:** a markdown replacement is re-encoded across the whole target span, so `[x](Old%2CFolder/Note.md)` renamed to `Fresh.md` yields `Old,Folder/Fresh.md` — an untouched folder loses its `%2C`. Semantics-preserving and idempotent, but a byte-level change the user didn't ask for. Recorded as an explicit exception in the Rewriter POLICY paragraph rather than silently accepted; preserving per-segment encoding would mean segment-wise diffing in the one place a mistake corrupts a note.

**New coverage:** an end-to-end roundtrip (extract → persist edges → resolve → rename → rewrite → drain `CheckpointNote` → assert body). Nothing previously proved a markdown link becomes a resolved edge and survives a rename.

**Checked clean:** offset arithmetic (hand trace + 20k-case property fuzz, zero failures), encode/decode round trips across emoji/CJK/combining accents/`+`/quotes and a file literally named `a%20b.md`, every `Parser.extract/1` consumer, and AGENTS.md compliance.

**Follow-ups, not in this PR:** markdown links aren't clickable in the SPA viewer (`note-view.tsx` only routes hrefs starting with `/`), and `docs/api-contract.md` in the workspace repo still says "wikilink extraction".

PR #1316, opened 2026-08-08
perf(web): one bulk read for the file tree

Rasbandit:
## The problem

Loading the web file tree fired one `GET /api/folders/list?folder=X` **per folder**. Measured on a real 33-folder vault via Resource Timing:

| Metric | Measured |
|---|---|
| Folder-list calls per load | 20–33 |
| Total bytes those calls delivered | 39.5 KB |
| Max **server** time for one call | 467 ms |
| Max **queue** time before hitting the wire | **5,626 ms** |
| Protocol | HTTP/1.1 (all 44 calls) |

The server was never slow. Browsers allow ~6 concurrent connections per origin on HTTP/1.1, so the tail of the fan-out waited 5.6 seconds just to be *sent*. A note opened during that window inherited ~1s of queueing despite ~100 ms of server time.

The fan-out was deliberate — `folder-tree.tsx` carried a loop commented "a small burst of parallel requests at startup". A small burst was 33 requests taking ~6s to drain.

**Root cause is structural.** `/api/folders/list?folder=` was designed for the plugin, which walks one folder at a time. The web tree needs the whole vault at once and was built on a per-folder API.

## The change

`GET /api/vault/tree` — one read returning every folder, note and attachment the tree renders.

Payload is deliberately minimal: `noteToTreeItem` derives title and extension from `path`, so `title`, `tags`, `version`, `content_hash` and `crdt_head` are all omitted. Reusing `/api/sync/manifest` was considered and rejected — it carries ~128 bytes/note of hashes the tree never reads (~1.3 MB at 10k notes), and it would couple the web tree to a sync protocol the plugin depends on.

The client fetches it once and seeds the **existing** cache keys, so every consumer (rename, drag-move, the loader) is untouched.

## Measured result

| | Before | After |
|---|---|---|
| Per-folder `folders/list?folder=<name>` calls | **20–33** | **0** |
| Max queue time | **5,626 ms** | **3 ms** |
| Tree payload | 39.5 KB across 20–33 requests | 89 KB in **1** request |

A cold load now issues **4** requests, not 1: `/vault/tree`, `/folders`, `/attachments`, and `/folders/list?folder=` (the root-notes list — `useFolderNotesById(ROOT_FOLDER_ID)` is an active observer that fetches on mount). Those four go out in parallel in a single connection wave. Collapsing them to 1 by pointing those hooks at the tree cache is a deliberate follow-up, not this PR — the pathology was 33 requests *serialized*, and that is gone.

## Notable review findings, fixed in-branch

- **Tenant isolation** — nothing covered the `vault_id` clause: deleting it left every test green while the endpoint would dump *all* of the user's other vaults into the tree. Now covered by a same-user/second-vault test.
- **Folder markers** — marker rows have NULL `path_ciphertext`, so a regression in the `kind == "note"` filter doesn't return junk, it raises inside path decryption and 500s the whole tree. `SyncController` had this test; it wasn't carried over. Now it is.
- **Cache stomp** — seeding blanket-overwrote `folder-notes-by-id`, which the CRDT channel also writes. Seeding is now bootstrap-only (`seedIfAbsent`), so live writes and in-flight optimistic rows survive.
- **Attachment `mtime`** — seeded rows hardcoded `0`, which feeds a real sort comparator in `loader.ts`. `mtime` and `updated_at` now come from the endpoint, so seeded and fetched rows are interchangeable.
- **Dead `since_seq`** — the short-circuit was implemented, tested and documented but never called, and returns no payload, so it is unusable without client-side tree persistence that doesn't exist. Deleted.
- `Engram.Notes.folders_payload/2` is now shared by this endpoint and `/api/folders` instead of duplicated (the copies had already drifted on root's parent).

## Gates

`mix format` · `mix credo --strict` (782 files, no issues) · `mix dialyzer` exit 0 · `mix test` 4213 tests, 0 failures · `SPEC IN SYNC` · `tsc --noEmit` · `biome ci` (461 files) · `vitest` 1312/1312.

Verified in a real browser against a real vault, not just in tests.

## Follow-ups (not this PR)

- `channel.ts` invalidates the per-folder caches but not `vault-tree`, so live updates refill through the old per-folder endpoints. Bounded to root plus folders with an active observer; unchanged from previous behavior.
- Folders with zero notes get no seeded key, so expanding one costs a single lazy fetch.
- The ~500 ms per-note CRDT open (`openDoc`: session wait + per-doc IndexedDB + STEP1 handshake) is untouched and is the dominant remaining per-file latency.

### DIFF
--- a/docs/superpowers/plans/2026-08-07-vault-tree-bulk-endpoint.md
+++ b/docs/superpowers/plans/2026-08-07-vault-tree-bulk-endpoint.md
@@ -1,0 +1,720 @@
+# Vault Tree Bulk Endpoint Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Replace the web tree's per-folder request fan-out (20–33 calls, ~5.6s of HTTP/1.1 queueing) with a single `GET /api/vault/tree` call.
+
+**Architecture:** A new read-only Phoenix controller returns every folder, note, and attachment the tree renders, in one payload, with a `since_seq` short-circuit. The React client calls it once on vault load and seeds the *existing* TanStack cache keys (`folders`, `attachments`, `folder-notes-by-id`) via `setQueryData`, so every existing consumer keeps working untouched and the per-folder prefetch loop is deleted.
+
+**Tech Stack:** Elixir/Phoenix 1.8, Ecto, OpenApiSpex, ExUnit; React 19, TanStack Query v5, Vitest.
+
+## Global Constraints
+
+- **Additive only.** `/api/folders`, `/api/folders/list`, `/api/folders/by-id/:id/notes` stay exactly as they are — the plugin and public API depend on them.
+- **OpenAPI is CI-gated.** `.github/workflows/verify.yml` regenerates the spec and diffs it against the committed `openapi.json`. A new route without an `operation(...)` block, a response schema, and a regenerated `openapi.json` **fails CI**.
+- **Paths are encrypted at rest.** Every path read decrypts `path_ciphertext`/`path_nonce` with the user DEK and AAD binding. Never bypass `path_aad/3`.
+- **Tenant isolation.** All queries run inside `Repo.with_tenant(user.id, fn -> ... end)`.
+- **No version bumps** in feature work — release-please owns `mix.exs`.
+- Backend gate before push, run **sequentially**: `mix format`, `mix credo`, `mix dialyzer`, then full `mix test --warnings-as-errors`.
+- Frontend gate: `bunx tsc --noEmit`, `./node_modules/.bin/biome ci .`, `bunx vitest run`.
+
+## Deviations from the approved spec
+
+Two changes I'd make, both discovered while reading the code. **Confirm before Task 1.**
+
+1. **Folder objects use `name`, not `path`.** The spec wrote `{id, path, parent_id, count}`. The existing `GET /api/folders` and the frontend `Folder` interface (`frontend/src/api/queries.ts:303`) both use `name` — which already *is* the full path. Matching the existing shape means the client can seed the `folders` cache directly with zero mapping.
+
+2. **The client seeds `folders` and `attachments` caches too, not just notes.** The spec flagged "attachment double-source" as a risk. Seeding all three caches from the one response resolves it: `useFolders`/`useAttachments` become cache reads instead of separate fetches, so there is exactly one source of truth and no consumer changes. This is what makes it a true single call (35 → 1) rather than 35 → 3.
+
+## File Structure
+
+| File | Responsibility |
+|---|---|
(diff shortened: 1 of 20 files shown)


Rasbandit:
## Code review findings (high effort, 10 confirmed) + CI regression

`e2e-browser` failed on the first push — **a real regression this branch introduced**, not the known flake. `tree-ops-sync.spec.ts:280` "move note propagates to a second tab": after a move, tab B showed the note under **both** the old and new folder. Failed on the initial run and the retry.

**Root cause:** `seedIfAbsent` writes with `qc.setQueryData`, which creates a cache entry with **no `queryFn`**. The bulk-prefetch loop it replaced used `qc.prefetchQuery`, which registers one. `channel.ts` invalidates `folder-notes-by-id` with `refetchType: "all"` *specifically* because the tree loader reads via `getQueryData` and would otherwise leave entries "stale-but-unfetched forever" — but that can only refetch a query that has a `queryFn` or an observer. Seeded non-root folders have neither, so the invalidation silently no-ops (verified in `@tanstack/query-core` 5.101: `ensureQueryFn` returns a rejecting promise, swallowed by `promise.catch(noop)`, `state.data` preserved).

Fixed in `9df09d8c` by registering a `queryFn` default for the key family. Regression test verified RED then GREEN.

### Remaining findings — not yet addressed

Five findings share one root: **seeding caches behind the owning hooks' backs.**

| # | Finding |
|---|---|
| 2 | An invalidation firing while the tree request is in flight hits a key that doesn't exist yet, so it's discarded; the stale snapshot then lands marked fresh. Cross-device create/delete/rename in that window is lost until reload. |
| 6 | `useFolders`/`useAttachments` fetch unconditionally and win the race, so those payload halves are decrypted, serialized and discarded every load. |
| 8 | Note-less folders get no seeded key, so they still fan out per-folder. |
| 9 | `change_seq` is shipped but never read; nothing can detect the staleness in 2. |

**Recommended fix for all four:** point `useFolders`/`useAttachments`/`useFolderNotesById` at the tree query (`queryFn: () => qc.ensureQueryData(["vault-tree", vaultId])`) instead of seeding sideways. Every entry then has a real fetcher and one owner. This deletes `seedTreeCaches`/`seedIfAbsent` entirely and is a smaller diff than what's here now.

Independent of that:

| # | Finding |
|---|---|
| 3 | `PathCrypto.decrypt!` raises where `Attachments.list_attachments/2` skips-and-logs — one undecryptable row 500s the whole tree read. `useVaultTree`'s error is never read, so it silently reverts to the fan-out; only signal is a Sentry 500. |
| 4 | `Crypto.get_dek/1` can return `{:error, :unrecognised_blob}`; the two-clause `case` raises `CaseClauseError`. Inherited from `SyncController`, but blast radius grew from one folder to the whole sidebar. |
| 5 | `useVaultTree` is ungated on `vaultId` — a deep-link with a stale `localStorage` vault pays two full-vault decrypt passes. |
| 7 | `folder-tree.test.tsx` pins `useVaultTree` to `{data: undefined}`, so the `FolderTree → seedTreeCaches` wiring is never exercised. It proves the old fan-out is gone, not that the new path works — which is how finding 1 reached CI. |
| 10 | `render_tree/5` duplicates `SyncController.render_manifest/5` wholesale and lives in a controller rather than a context module; the notes `where` re-spells `scoped_live/2` by hand, outside the tenant-scope lint's reach. |

Issue #1320, opened 2026-08-08
CRDT checkpoint seam: four content-loss/corruption findings tracked as one workstream (#847, #902, #958, #959)


Rasbandit:
## Why this exists

Four issues, filed separately over ~5 weeks, are the **same seam**: `Engram.Notes.CrdtCheckpoint`'s `union_with_row_state` → `maybe_flatten` → persist pipeline. Each arrived as a single retro-review finding on an already-merged PR (#953, #846), so each got triaged on its own severity and landed at p2. Nothing in the tracker represented them as one thing.

Grouped, they are the most dangerous work open in this repo: **three of the four silently destroy or duplicate user note content.** Hence p1 as a bundle.

Both checkpoint write paths run the same `with` chain — `crdt_checkpoint.ex:229` (unchanged-branch `update_all`) and `crdt_checkpoint.ex:289` (full-path `Repo.update!`) — which is why a fix in one place plausibly covers several of these, and why fixing them piecemeal risks four partial patches to one function.

## The four

| # | Failure mode | Severity |
|---|---|---|
| #847 | Neither write path has a version guard. A room checkpoint racing a committed REST write **LWW-reverts the REST content**. Widened by the materializing unbind — the REST op never enters the tail-log, so nothing detects it. | Data loss (Important, PR #846 final review) |
| #902 | Same class from the other side — the REST→live-room convergence gap. Audit established `seed_from_content` is *not* the problem (it works, it's just only reached by migrated/legacy rows); the divergence is in the checkpoint path. | Data loss |
| #958 | `maybe_flatten` deliberately breaks Yjs lineage (a reset, not a merge) and persists fresh-lineage state to the row, while the live room keeps running its **old** lineage. The Phase-0 union then merges disjoint lineages → **content doubles**. | Data corruption (Important, ~82 confidence, rare trigger) |
| #959 | The Phase-0 abort on unreadable `crdt_state` is correct (never overwrite state we cannot prove we contain) but has no bounded remediation for *permanently* unreadable state (lost DEK version, botched AAD rebind). Every debounce tick appends a tail row and aborts → `crdt_update_log` grows unbounded, note is **frozen**. | Availability + unbounded growth (Important, ~85 confidence) |

## Suggested approach

Treat as one workstream, not four PRs. Roughly:

[...]

Issue #847, opened 2026-07-02
crdt: fence the compaction + structural checkpoint writes (materialize path done in #907)


Rasbandit:
From PR #846 final review (Important). Both checkpoint write paths (unchanged-branch update_all and the full-path Repo.update!) have no version guard: a room checkpoint racing a committed REST write can last-write-wins revert the REST content. Widened by the materializing unbind (the REST op never enters the tail-log, so nothing replays it). Fix: optimistic version check (where: n.version == ^note.version), skip + log on mismatch — tail replay reconciles on next bind.

Rasbandit:
Re-prioritized **p2 → p1** by the 2026-08-08 triage sweep, and bound into the checkpoint-seam workstream: **#1320**.

Rationale: this is one of four findings that all land on the same `Engram.Notes.CrdtCheckpoint` pipeline (`union_with_row_state` → `maybe_flatten` → persist, both write paths at crdt_checkpoint.ex:229 and :289). Each was filed separately off a retro-review of an already-merged PR, so each was triaged alone and landed at p2. Read together, three of the four silently destroy or duplicate user note content.

Fix as part of #1320 rather than in isolation — four independent patches to one function is how you get a fifth finding.

Rasbandit:
**Rescoped by the 2026-08-08 audit** — the materialize path is already fenced; full findings on #1320.

Original body says "both checkpoint write paths ... have no version guard." Against \`origin/main\` @ 60859f5b there are **three**, and one is done:

| path | line | fenced |
|---|---|---|
| materialize (content change) | :405 | ✅ #907 |
| compaction (`prev == content_hash`) | :344 | ❌ |
| structural (`.canvas`) | :293 | ❌ |

**Remaining item 1 — the compaction/structural fence.** Reachable and real: the union folds `row_state` read at :141, a REST write commits, then :344 persists `union(old_row, live)` — dropping that write's ops from `crdt_state` while `notes.content` holds the new text. Same revert class as #902, through the compaction door. Fencing is safe: a skipped compaction loses nothing, because the tail is not pruned and the ops replay on next bind.

**Remaining item 2 — the unbind/worker fence, which needs a DECISION, not a patch.** `current_version/2`'s docstring (:465-468) states the unfenced path is deliberate:

> nil is indistinguishable from the unbind path's absent `captured_version` **(which must stay unfenced to persist on room exit)**

[...]

PR #1325, opened 2026-08-08
fix(crdt): re-union on a stale checkpoint fence instead of dropping the room's ops

Rasbandit:
Closes the unbind half of #847. Part of #1320.

## The gap

The #902 version CAS (shipped in #907) **aborts** the checkpoint when the row moved after the doc snapshot. For a live room that is correct — `deliver_out` re-merges and a later debounce tick re-persists.

**On unbind there is no later tick.** The room is exiting, so an abort discards the doc's ops permanently. And they are not recoverable from the tail: a REST/MCP write only reaches `crdt_update_log` once `deliver_out` pushes it into the live room and fires `update_v1`, so inside the commit→deliver window there is *no tail row to replay*.

`current_version/2` documents this tension directly:

> nil is indistinguishable from the unbind path's absent `captured_version` **(which must stay unfenced to persist on room exit)**

So the unbind path had to choose between reverting a committed write (unfenced) and losing the room's work (fenced). This PR removes the choice.

## The fix

On a CAS miss: re-read the row (READ COMMITTED shows the winner's commit), re-union the doc against it, re-fence on the version just read. Both sides survive.

```elixir
|> checkpoint_write(ctx.vault_id, note_id, ctx.prune, opts, %{...})
|> retry_if_stale(note, ctx, opts, budget)
```

The Yjs union is monotone, so each retry strictly converges. `@fence_retries` bounds it at 3. Exhausting the budget writes nothing and prunes nothing — loss-free, because the ops stay in the live doc and the tail, and the next tick (or the next bind's `replay_tail`) re-checkpoints them.

## Why not a lock

Considered and rejected on evidence, not taste. Every write path takes the **vaults** row lock via `Vaults.next_seq!` (an `UPDATE vaults ... RETURNING`) *before* touching the note. Checkpoint reads the note first and calls `next_seq!` later, so adding `SELECT ... FOR UPDATE` to the note read inverts the lock order against ~20 call sites — a textbook deadlock.

Making locks work would mean reordering acquisition everywhere, or a per-note advisory lock taken by every writer. Optimistic concurrency is the right primitive when lock ordering cannot be made consistent cheaply, and it is the strategy this module already committed to.

## Test

Written first and **verified failing for the right reason** — the REST assertion passed while the room assertion failed, which is the bug exactly:

```
1) test a stale captured_version re-unions and still persists ...
   stale fence dropped the room's ops instead of re-unioning (#847)
   code: assert fresh.content =~ "ROOM EDIT"
```

The existing #902 abort test still passes: a re-union write does not revert either.

## Refactor

The retry threads six invariants; credo flagged arity 9. They now travel in a `ctx` map, which also makes the retry call site read as what it is.

## Still open (deliberately not in this PR)

The **compaction** (`:344`) and **structural** (`:293`) branches still write `crdt_state` through an unfenced `update_all`. That hole is real but only reachable via a mid-transaction interleave, which is **not reproducible in the ExUnit sandbox** (single connection, no seam to pause a transaction at a precise point). I did not want to ship concurrency code whose trigger no test can exercise.

Tracked on #847 with that constraint named. Verifying it properly needs either a deliberate test seam or a two-connection harness — worth doing as its own scoped piece rather than smuggled in here.

## Verification

| gate | result |
|---|---|
| `mix format --check-formatted` | clean |
| `mix compile --warnings-as-errors` | clean |
| `mix credo --strict` | **no issues** |
| `mix dialyzer` | passed (5 pre-existing, ignore-filed, 0 unnecessary skips) |
| `mix test --warnings-as-errors` | **4245 tests, 0 failures** |

Run sequentially in an isolated worktree off `origin/main` @ 60859f5b.

https://claude.ai/code/session_016D4V5B8Yixkk6TjG8Yc4B2

### DIFF
--- a/lib/engram/notes/crdt_checkpoint.ex
+++ b/lib/engram/notes/crdt_checkpoint.ex
@@ -25,7 +25,7 @@ defmodule Engram.Notes.CrdtCheckpoint do
 
   require Logger
 
-  # Bound on the version-CAS re-union retry (see `retry_if_stale/9`). Each miss
+  # Bound on the version-CAS re-union retry (see `retry_if_stale/5`). Each miss
   # means another writer committed first; the re-union folds their state in, so
   # a small budget is enough for any realistic contention on ONE note. Exhausting
   # it writes nothing and prunes nothing, which is loss-free.
@@ -71,6 +71,18 @@ defmodule Engram.Notes.CrdtCheckpoint do
         :ok
 
       user ->
+        # Default-fence. Callers that snapshot the doc themselves pass their own
+        # `:captured_version` (captured BEFORE the snapshot, which is strictly
+        # better). The unbind and overflow-worker callers pass none, and used to
+        # fall through to an UNFENCED write — so the fence, and the re-union
+        # retry behind it, could never engage on exactly the path with no
+        # follow-up tick to self-heal. Capturing here makes every caller fenced;
+        # nil (read failure) still degrades to the prior unfenced behaviour.
+        opts =
+          Keyword.put_new_lazy(opts, :captured_version, fn ->
+            current_version(user_id, note_id)
+          end)
+
         do_checkpoint(user, user_id, vault_id, note_id, doc, opts)
     end
   rescue
@@ -252,8 +264,13 @@ defmodule Engram.Notes.CrdtCheckpoint do
   # abort is fine for a live room — deliver_out re-merges and a later debounce
   # tick re-persists — but on unbind there IS no later tick, and inside the
   # commit→deliver window the winner's ops are not in crdt_update_log yet
-  # (update_v1 only fires once deliver_out pushes into the live room), so the
-  # dropped ops are unrecoverable. The re-union keeps both sides.
+  # (update_v1 only fires once deliver_out pushes into the live room), so an
+  # abort there drops the room's work for good. The re-union keeps both sides.
+  #
+  # That same asymmetry is why the exhaustion clause below settles for writing
+  # nothing: it is NOT a claim that the ops are recoverable (on unbind they may
+  # not be), only that not writing is strictly better than overwriting the
+  # winner. Retrying is the fix; giving up is the last resort.
   #
   # The union is monotone, so each retry strictly converges.
   # `ctx` carries the parts that never change across a retry (vault/note ids,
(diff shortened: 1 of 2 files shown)


Rasbandit:
**Converting to draft — the central claim does not hold.** High-effort review (10 findings, 6 confirmed) caught it, and I verified the top one by hand.

## The retry is dead code on the path it was written for

Neither unbind caller passes `:captured_version`:

```
crdt_persistence.ex:271  CrdtCheckpoint.checkpoint(user_id, vault_id, note_id, doc)
checkpoint_note.ex:129   CrdtCheckpoint.checkpoint(u, v, n, doc, prune_ids: ids)
```

So `captured` is nil, `fenced_query == base_query`, the UPDATE always matches one row, `{0, _}` is unreachable and `:stale` is never returned. I built the miss *handler* and never wired the unbind caller to produce a miss. The new test passes `captured_version:` explicitly, so it exercises a path production never takes — it was green for the wrong reason.

## And the concurrency model behind it is wrong

`Vaults.next_seq!/1` is called at :461 **inside** the checkpoint transaction, so the checkpoint already holds the vault-global seq row lock before the fenced update runs. Consequences:

- Retries re-run decrypt/union/encode/encrypt **while holding that lock**, serializing every write in the vault (up to 4 rounds).
- Because that lock blocks any competing writer, retries 2-3 and the give-up clause are unreachable — `@fence_retries 3` encodes contention the lock already prevents.
- Each attempt burns a `change_seq` that never gets stamped on a row, leaving permanent gaps in the vault sequence.

## This reopens the option I rejected

I rejected `SELECT ... FOR UPDATE` on the grounds that it would invert lock order (note → vaults) against the universal vaults → note order. That reasoning was incomplete: **checkpoint already takes the vaults lock**, just late. Acquiring it *before* the row read would match everyone else's order — no inversion, no deadlock — and make the whole read-modify-write atomic, removing the need for a retry at all.

The cost is holding the vault lock across the checkpoint's crypto work, which is the same throughput concern finding 2 raises about the retries. That trade is a real design decision, not a patch, so it should be made deliberately rather than bolted onto this branch.

Reworking. The problem statement on #847 stands; this implementation of it does not.

Rasbandit:
## Second review also came back with confirmed findings. Stopping here, not merging.

10 more findings, 8 confirmed. Two matter structurally:

### 1. `put_new_lazy` does not replace an explicit `nil`

`CrdtCheckpointTimer` and `CrdtChannel` compute their own `:captured_version` and pass it. When *their* `current_version/2` read blips and returns nil, `Keyword.put_new_lazy` sees the key **present** and skips the re-read. Those callers stay unfenced, and my new comment claiming "every caller is default-fenced" is false. One-line fix, but it means the last round's headline fix is still incomplete.

### 2. The retry cannot work inside the transaction — this is the real problem

`Vaults.next_seq!` takes the vault-global row lock **mid-transaction**, and a lock cannot be released without ending the transaction. My `stale_version?/2` pre-check is racy by construction (I said so in the comment), so a CAS miss can still land *after* the lock is taken. When it does, `retry_if_stale/5` recurses **inside that same transaction**, holding the vault lock across another decrypt/union/encode/encrypt cycle and calling `next_seq!` again with no rollback.

So the pre-check narrowed the window but did not fix the shape. My comment claiming it "leaves no gap in the vault's change sequence" is wrong.

## The correct design

Retry at the **outer** level, not the inner one. Each attempt gets its own transaction:

```
attempt:
  BEGIN
    read row (version v) -> union -> build -> next_seq! -> CAS on v
    miss? -> Repo.rollback(:stale)      # releases the vault lock
             # AND rolls back the next_seq! increment
  COMMIT
  on :stale and budget left -> new attempt, fresh transaction
```

That fixes the lock-holding, the seq burn (a rolled-back increment is genuinely undone), and makes the budget meaningful — all three from one structural change rather than three patches.

## Why I am not doing that now

This is the third structural revision of the most delicate file in the system, and each of the first two shipped with me confident and wrong about the transaction/lock structure. I do not think an unreviewed overnight redesign of the checkpoint write path, self-merged, is the right call — the failure mode here is silent note loss.

Staying **draft**. The problem statement on #847 stands, the design above is what I would build, and it wants human eyes on the rollback semantics before it goes near main.

Also still open from this round: the compaction fence cannot detect a competing compaction (compaction never bumps `version`), compaction can now give up without pruning (tail growth on the common path), `current_version/2`'s docstring still says unbind must stay unfenced, the stale-fence paths are invisible in Loki (`:info` on `:sync` does not ship), and the new retry branches are untested.

Rasbandit:
Closing in favour of #1330, having verified the supersession rather than taking the label at face value.

**The fix here is genuinely covered there.** I checked out #1330's branch, applied this PR's two tests to it **unmodified**, and ran them:

```
28 tests, 0 failures
```

including `stale fence dropped the room's ops instead of re-unioning (#847)`. So #1330 satisfies the room-side requirement this PR was opened for.

**And its design is better on three counts this PR could not reach:**

| | this PR | #1330 |
|---|---|---|
| retry | inside the txn, holding the vault-global lock | one txn per attempt, `Repo.rollback(:stale)` |
| `next_seq!` burn | permanent, 2-4 per write | rolled back with the txn |
| unfenced paths left | compaction + structural | none |
| reproduces the bug | no (sandbox can't) | yes, deterministically, 3/3 on unmodified main |

That last row is the deciding one. This PR reasons about a race no test can trigger; #1330 built a two-connection harness and observed the loss first.

**Nothing is lost by closing.** The two tests here assert the opposite direction of the race from #1330's interleave test (room's ops preserved, vs REST write not reverted), and #1330 did not carry that direction. They are now ported onto its branch in `cd843d97`, with the comments rewritten for the rollback-and-retry mechanism.

Issue #1317, opened 2026-08-08
perf(web): note open costs ~500ms with zero network — warm the CRDT document path


Rasbandit:
## Measured

Clicking between notes in the web app, dev-selfhost against a real vault, Resource Timing + DOM instrumentation:

| Note | REST fetch | Title swaps | Content appears |
|---|---|---|---|
| _AI Instructions | 237 ms | 328 ms | 663 ms |
| Abs | 240 ms | 640 ms | 756 ms |
| boobs | **cached (0)** | 456 ms | 519 ms |
| CLAUDE | **cached (0)** | 275 ms | 594 ms |

The decisive rows are the cached ones: **zero network, still ~550 ms to content.** The note fetch is not what's slow.

The remaining ~300–500 ms is the CRDT open path: `openDoc` waits on `waitForSessionStart()`, then `manager.getSharedText` awaits a **per-document IndexedDB database** (`y-indexeddb` creates one DB per note, `engram-crdt/<noteId>`) reaching `whenSynced`, then the STEP1 handshake, then CodeMirror view construction.

Then `closeDoc` calls `doc.destroy()` on every switch, so returning to a note you had open ten seconds ago pays the entire cost again. **Every note open is a cold start.**

## Why this is the root, not a symptom

Two shipped changes have already worked around the consequences rather than the cause:
- the note-pane flash fix held the outgoing document open to avoid a blank editor
- `/api/vault/tree` (#1316) removed startup request queueing, which was a *different* latency source

Neither makes opening a file faster. This is the one that would.

## Path, in order

**0. Instrument `openDoc` before designing.** The ~500 ms is confirmed and is not network, but the split between session-wait, IndexedDB open, STEP1 handshake, and CodeMirror construction is unknown. Every option below is sized differently depending on that split. A few `performance.mark` calls and one measurement session.

**1. Warm documents — an LRU instead of `closeDoc` on switch.** Almost certainly the largest win for the reported symptom (clicking between files). Back-and-forth navigation becomes instant because nothing was destroyed. Bounded memory, explicit eviction.

**2. One local store instead of one IndexedDB database per note.** A single IDB with a documents table. Per-doc DB open is a fixed cost paid on every first open, forever, and it scales badly with vault size.

**3. Prefetch on intent.** Hover or keyboard-focus a tree row → `openDoc` in the background. The tree already does exactly this for folder note lists.

[...]

PR #1328, opened 2026-08-08
perf(web): prefetch a note on hover

Rasbandit:
Prefetch a note's REST payload on sustained hover. That fetch is ~240 ms of the note-open path (#1317), and the gap between pointing at a row and clicking it is usually longer.

Deliberately **REST-only**. Warming the CRDT doc would also need `enroll`, which registers the note for server-side fan-out — paying that for every row a pointer crosses would put real load on the server for notes nobody opens.

Guards, all of which came out of review:

- **`fetchQuery`, not `prefetchQuery`.** query-core implements `prefetchQuery` as `fetchQuery().then(noop).catch(noop)`, so its promise never rejects and a `.catch` on it is dead code. The rejection matters: a failed speculative hover must not leave an errored entry cached under the key `useNote` reads, or a later real click renders the hard error pane instead of loading.
- **Optimistic rows are skipped via `pending`, not an id-shape check.** We mint those ids ourselves as real uuid7s, so no regex can spot one. `TreeItem` now carries the flag.
- **Attachment rows never fire it** — `GET /notes/by-id/<attachment-uuid>` is a guaranteed 404 that `vault-item-page` explicitly gates against elsewhere.
- **Cancels on pointer leave / blur**, so a pointer sweeping the list doesn't fetch the last row it crossed. Without that the debounce didn't actually require a sustained hover.
- **`onFocus` alongside `onPointerEnter`**, matching the folder branch, so keyboard tree navigation gets it too.

## What this PR no longer contains

It originally also added a "handshake seed" — the REST body rendered over the editor to cover the ~345 ms where the CRDT doc is bound but empty. **Three review rounds, and each set of fixes spawned new defects**, including one strictly worse than the blank editor it replaced: gating the reading-mode fallback on the seed's expiry meant a never-completing handshake *deleted the body the user was reading* off the screen after 2.5 s. Others: the latch re-armed on navigation and still resurrected just-deleted text; the seed sat inside the lazy-editor Suspense boundary so it never painted on the cold-chunk path (the longest blank window it was built for); the reading handover flashed empty for 300 ms against the debounced `liveContent`; `aria-hidden` left screen readers with nothing.

That is the signal to question the design, not to attempt fix #4. The seed is a second rendering surface that has to agree with the editor about content, mode, latch state, expiry, Suspense placement, a11y and click targets — every one of those produced a defect. It wants a deliberate design pass, not another patch. Filed as remaining #1317 work.

## Also from this work, no code attached

On the ~345 ms STEP1 round trip, two hypotheses are now ruled out with evidence:

- **Not `FanoutPacer`** — it paces only `note_yjs_update` fan-out on the `sync:` topic; a room's STEP2 reply to a client's STEP1 goes straight back through `SharedDoc`.
- **Not room cold-start** — a note whose room was demonstrably warm measured **341 ms**, against 357 ms and 399 ms for never-opened notes.

Consistent and server-side after routing; localising it further needs backend timing.

## Testing

`bun run test` → **177 files, 1359 tests, 0 failures**. `tsc` + `biome` clean.

Refs #1317


Rasbandit:
Closing rather than iterating further. This is the fourth review round on this PR and it is still producing confirmed correctness findings, which is the signal that the problem is the design rather than the individual bugs.

**Root issue:** the hover prefetch writes into the *same* `["note", vaultId, id]` cache entry NotePage renders from. That makes every speculative failure a user-visible failure:

- It prefetches the note that is **currently open** (its row is the one just clicked, so it is under the pointer). A 401/429 on that speculative request flips the shared query to error and NotePage replaces the note being read with "Failed to load note" — for a note that was rendering fine.
- Hover issues authenticated calls with **no user intent**. With a stale token, moving the pointer down the sidebar produces one 401 per row rested on, all silently swallowed — the exact shape that trips the `loki-auth-failure-burst` paging alert, fired by mouse movement.
- The error path's `removeQueries` can cancel a mounted observer's in-flight request and leave it wedged in `fetching` (this file bans `removeQueries` in comments twice).
- `onFocus` fires on headless-tree's *programmatic* focus after renames and folder expands, so fetches fire with no gesture at all.

Plus: `pending` threaded through only one of two note-item factories, one shared timer so mouse and keyboard cancel each other, no `vaultId` guard, `retry: 1` doubling every failure, and no test on the debounce/cancel logic.

All fixable — isolate the speculative read from the rendered entry, skip the open note, never evict, no retry, per-row timers, drop `onFocus` — but that is a redesign, and the payoff is hiding ~240 ms on hover. Not worth shipping in this shape.

Folded into #1317 alongside the handshake seed, with the constraints both attempts uncovered.

Branch `perf/note-open-rest` is left in place if we want to pick it back up.

PR #1330, opened 2026-08-09
fix(crdt): fence every checkpoint write, retry by rolling the txn back

### LINKED ISSUE #1326: test: two-connection harness for mid-transaction races (blocks honest coverage on the checkpoint seam)
Split out of #1320 / #847 after it blocked honest test coverage twice in one session.

## The gap

Several confirmed correctness issues in `CrdtCheckpoint` are only reachable when a write commits **between** the checkpoint's in-transaction row read and its `update_all`. At READ COMMITTED that window is real: the read sees v1, another transaction commits v2, and the unqualified UPDATE sees v2.

The ExUnit sandbox cannot produce it. Each test owns one connection inside one transaction, and there is no seam that lets a test pause a checkpoint mid-transaction while a second connection commits.

## What this has already cost

- **#1325** — the version fence and its re-union retry are correct and reviewed, but the race they guard is unproven by test. Two attempts at a test were green for the wrong reason (an unfenced write bumps `version` identically, so the assertion did not discriminate).
- **#847** — the compaction/structural fence had to ship without a test that exercises its trigger.
- **#958** (checkpoint union doubles content after a flatten) and **#959** (abort leaks tail-log) are the same shape and will hit the same wall.

The repo has concurrency tests (`note_upsert_race_test.exs`, `folder_marker_race_test.exs`, `invites_concurrency_test.exs`), but they race two *operations*, not two *transactions with a controlled interleave point*.

## What would close it

Some way to deterministically interleave. Options worth costing:

1. **Shared-mode sandbox + a second checked-out connection**, with the interleave point expressed as a test-controlled barrier (an agent the checkpoint calls into, injected only in test config).
2. **A `:telemetry` handler as the pause point** — checkpoint already emits around the write; a test handler could block there. No production seam, but couples tests to event names.
3. **Advisory-lock choreography** — the test takes a lock the checkpoint must pass through, releases it after committing the competing write.

(1) is the most honest and the most invasive; (2) is the cheapest and reuses machinery that already exists. Worth a spike before picking.

## Why p2 and not p3

It is not user-facing, but it gates verification on the highest-severity open cluster (#1320). Every fix in that cluster currently ships on reasoning plus review rather than on a failing-then-passing test, which is exactly the bar this codebase otherwise holds.

## Non-goal

Not asking for general concurrency testing infrastructure. One reusable helper that pins a specific interleave for the checkpoint write path is enough to unblock #958, #959 and the #847 remainder.

Rasbandit:
Closes #1326. Closes the #847 / #902 loss class. **Supersedes #1325.**

## The bug

`CrdtCheckpoint` reads the note row inside `Repo.with_tenant`, does crypto + Yjs work, then writes. At READ COMMITTED a REST/MCP write committing in that gap is visible to the UPDATE, so an unfenced checkpoint overwrites it — the user's committed edit is gone.

**Three prior attempts shipped without a test that could fail, and two passed for the wrong reason.** The ExUnit sandbox gives one connection inside one never-committed transaction, so no second writer can commit at all. `note_upsert_race_test.exs` admits this in its own header:

> ExUnit's shared sandbox serializes the tasks through one connection, so this does NOT trigger the SQL race in isolation.

## The harness (#1326)

`test/support/checkpoint_interleave.ex` — real (non-sandbox) connections so writes genuinely commit, plus a park point so a competing write lands **inside** the read→write gap. `interleave/1` in `CrdtCheckpoint` reads a hook from `Application` env; unset everywhere outside these tests.

It reproduced the data loss on unmodified `main`, **deterministically, 3 runs of 3**:

```
checkpoint reverted a REST write that committed mid-transaction.
got: "ROOM EDIT"        # "REST WINS" was silently destroyed
```

That is the first time this bug has been observed rather than reasoned about.

## The fix

**One attempt = one transaction.** The write is fenced on the row version read *inside* that transaction; a miss calls `Repo.rollback(:stale)` instead of retrying in place. The rollback is what makes it work:

- `Vaults.next_seq!/1` has already taken the **vault-global row lock**, and a lock cannot be released without ending the transaction. Retrying in place would hold it across another decrypt/union/encrypt cycle, blocking every write in the vault.
- The rollback **undoes that `next_seq!` increment**, so a lost race leaves no gap in the vault's change sequence.
- The next attempt re-reads, so it unions against the winner rather than the stale snapshot. The Yjs union is monotone, so attempts converge.

There is **no unfenced write path left** — compaction was previously unfenced and is not now (it rewrites `crdt_state`, and a lost race there drops ops not yet in the tail).

## Why this supersedes #1325

#1325 retried *inside* the transaction and needed an extra pre-transaction `current_version` read to source the fence. Two review rounds, 14 confirmed findings. This design drops three of them by construction:

| | #1325 | this |
|---|---|---|
| Retry | inside the txn, holding the vault lock | one txn per attempt, rollback |
| `next_seq!` burn | permanent, 2-4 per write | rolled back with the txn |
| Fence source | extra query + `put_new_lazy` (skipped an explicit `nil`) | the version already read in-txn |
| `stale_version?` pre-check | needed | deleted |
| Unfenced paths | compaction, structural | none |

## The harness's real cost, stated plainly

Real commits mean **every side effect must be swept, including asynchronous ones.** This took three rounds and the intermediate states were actively misleading:

1. **RLS silently ate the first cleanup.** A bare `DELETE` with no `app.current_tenant` matches **zero rows and reports success**, so children survived and the `users` delete failed on the FK. Deletes now run inside `with_tenant`.
2. **`catch _, _ -> :ok` hid that.** 8 users/vaults/notes accumulated across runs and surfaced as **27 failures in unrelated suites** that scan globally. Cleanup now raises.
3. **`on_exit` runs in `ExUnit.OnExitHandler`**, not the test process, so it owned no connection. It checks one out.
4. **`oban_jobs` has no FK to `users`**, so 13 enqueued jobs per run survived and broke **14 more suites** that assert on the queue.

Verified zero residue in `users` / `notes` / `vaults` / `oban_jobs` after a full run. Worth knowing before anyone adds a second test on this harness.

## Verification

| gate | result |
|---|---|
| `mix format --check-formatted` | clean |
| `mix compile --warnings-as-errors` | clean |
| `mix credo --strict` | no issues |
| `mix dialyzer` | passed (5 pre-existing, ignore-filed) |
| `mix test --warnings-as-errors` | **4253 tests, 0 failures** |
| DB residue after full run | **0** across all four tables |

Isolated worktree off `origin/main` @ 418a9428.

https://claude.ai/code/session_016D4V5B8Yixkk6TjG8Yc4B2


Rasbandit:
Deep review run (26 agents, high effort, adversarial verify). **10 findings: 5 fixed here in `64e967b6`, 5 routed.**

## Fixed

**The rename revert — the serious one.** `checkpoint_write` re-derived the path/folder envelopes from the row it read, and the envelope nonce is random, so those columns landed in `changeset.changes` on *every* checkpoint even when nothing moved. The rename writers bump `seq`/`updated_at` but **not** `version`, so the fence is blind to them.

The shape is worse than a plain revert: `path_hmac`/`basename_hmac` are deterministic over the path, so they matched the snapshot, stayed **out** of `changes`, and the row kept the rename's NEW hmac while `path_ciphertext` went back to the OLD path. **The row resolves by the new path and decrypts to the old one.**

Reproduced in your harness before fixing — without the fix:

```
row resolved at the new path but decrypted to "p.md" — path_hmac and path_ciphertext disagree
```

Fixed by dropping the seven location columns from the checkpoint's write set rather than widening the fence. Location is not this writer's column set, and a writer that does not touch a column cannot lose a race for it.

> A first attempt at that test lived in the sandbox suite and **passed without the fix** — the sandbox cannot commit between the read and the write, which is exactly the point of this PR's harness. Moved and verified failing.

Also fixed: a non-`:stale` rollback raised `CaseClauseError` that `do_checkpoint`'s rescue flattened into `:ok` (the room exited believing it persisted); the `caller_owns_txn?` branch degraded *silently* to a one-shot checkpoint that also gapped the vault seq, now logs at `:error` naming both; `interleave/1` is compiled to a no-op outside `:test` so the hook cannot be armed in prod; and a test for the design's headline "a lost race leaves no seq gap" claim, which nothing asserted.

## Routed, not fixed

| finding | where |
|---|---|
| compaction branch fences on `version` but never bumps it — same defect as `.canvas` | added to #1333 |
| REST/MCP write is PK-only, so the mirror-image clobber survives | filed as #1335 |
| give-up path drops room-exit edits | **your call** — see below |
| retry re-unions across a deliberate `flatten/1` lineage break | **your call** — see below |
| `:captured_version` / `current_version/2` now dead weight | follow-up |

Two need your judgement rather than mine:

- **The give-up path.** Three lost fences write nothing and announce nothing, so an unbind can drop the user's last edits with no follow-up tick. Your comment argues "not writing beats overwriting a committed write" and the tail stays unpruned, so they are recoverable on the next bind. That is a product call about which loss is worse — I did not want to overturn it in a review pass.
- **The flatten interaction.** The retry drops `:captured_version` and re-unions the stale `live_state` against the winner's state, which is only safe while they share Yjs lineage — and `flatten/1`'s own docstring says lineage is *intentionally* broken. Narrow (needs a competing checkpoint to cross both flatten thresholds inside the retry window) but real, and the right fix depends on what you want a retry to do when it observes a reset.

## Gates

format ✓ · credo --strict 0 issues ✓ · dialyzer ✓ · `mix test --warnings-as-errors` **4258 tests, 0 failures** ✓

Rasbandit:
**I reverted my two commits on this branch** (`64e967b6`, `9e7ae914`). `crdt_checkpoint.ex` is now byte-identical to the state your review left it in. Your fence + rollback + harness work is untouched.

## Why

Three review rounds found **six correctness bugs, all mine**, each round's fix introducing the next round's bug:

1. Dropping the location envelopes while still stamping `dek_version` left a legacy row **half-migrated** — permanently undecryptable, invisible to `AadRebind`, and **one such row 500s the entire vault tree**. Strictly worse than the rename revert it fixed.
2. Gating that drop on `dek_version == row_version_aad_bound()` is an equality test against `2`, but DEK rotation stamps `3, 4, …` — so every rotated user took the other branch and kept the original bug, while the same write stamped their row back **down** to 2.
3. The title drop reproduced the half-migration in a second place, and its filename heuristic means a title matching its basename can never be rewritten again.

Two of my tests were also vacuous: the sandboxed rename test passed with the fix reverted, and the seq-gap test's fence never actually went stale.

The common thread: I was patching an encryption-versioning invariant I did not understand, in a module where being wrong costs a whole vault's decryptability. That does not belong in a PR whose fence work was already reviewed and green.

## What the rounds did establish → #1336

Not lost. #1336 records:

- the rename revert, **reproduced in your harness**, with the exact corruption shape (`path_hmac` deterministic so it stays out of `changes`; `path_ciphertext` always written — row resolves at the new path, decrypts to the old)
- why a sandboxed test for it is **necessarily vacuous**
- a **pre-existing** half-migration in the compaction and structural branches that is independent of this PR and **reachable on `main` today**
- the four traps that each cost a round
- a suggested shape (allowlist, not denylist — `@location_keys` was a fourth hand-maintained copy of the same seven atoms)

## Where this leaves #1330

Back to your original scope, plus the #1325 room-side tests ported in `cd843d97` (verified passing against your implementation before porting).

The findings from the **first** round that were on your code — the give-up path dropping room-exit edits, the retry re-unioning across a `flatten/1` lineage break, the `CaseClauseError` on a non-`:stale` rollback, and `:captured_version` now being dead weight — are still open and still yours to judge.

Rasbandit:
**Closing as superseded.** The bug this fixes is already fixed on `main`, by a simpler mechanism that landed while this sat open.

## What overtook it

| PR | what it landed |
|---|---|
| **#1356** | the two-connection interleave harness — `test/support/checkpoint_interleave.ex`, the thing #1326 asked for and this PR built |
| **#1360** | `snapshot_fence/2` on **every** checkpoint write path (markdown, compaction, structural) + a mid-txn `checkpoint_interleave_test.exs` |
| **#1340** | the half-migration fix — i.e. the two commits reverted here in `25a21e5d` |
| **#1351** | the DEK-rotation gate |

So `main` already has: the harness, a fence with no unfenced path left, and a test that parks a competing write inside the read→write gap.

## The one real difference, and why main's side wins

This PR fences on `version` and, on a miss, **rolls the txn back and retries**, unioning against the winner. `main` fences on `seq` + the `crdt_state_ciphertext` pre-image and, on a miss, **skips without pruning the tail**.

Skipping is safe because `crdt_persistence.ex:135` appends every room update to `crdt_update_log` before any checkpoint runs. The tail is the durable copy, so a checkpoint that declines to write loses nothing — the next bind replays it. That invariant removes the need for the retry loop, the `Repo.rollback(:stale)` dance, and the `next_seq!`-burn reasoning in this PR's description entirely.

The consequence: this branch's `crdt_checkpoint_test.exs` case *"a stale captured_version re-unions and still persists"* asserts both `REST WINS` and `ROOM EDIT` survive a single stale-fence attempt. On `main` that write is skipped, so it would fail — **by design**, not by regression. Its sibling *"the unbind signature still persists on room exit"* is covered by the existing suite.

## Mechanical state

- CI green, but on a base that is now 20+ commits stale
- Trial-merged into current `main`: conflicts, in `lib/engram/notes/crdt_checkpoint.ex`
- Rebasing would mean re-litigating skip-vs-retry to land nothing new

## Salvaged

The review rounds here surfaced one thing that is still live on `main`: four comments still claim a nil `:captured_version` means an **unfenced** write, which stopped being true at #1360. Read literally they say the room-exit path writes unprotected — the exact premise behind three failed attempts on #1325. Fixed in **#1364**.

Everything else from the rounds is already recorded in #1336 (closed via #1340) and #1335 (closed via #1356/#1360).

PR #1373, opened 2026-08-13
fix(auth): quiet the device-flow happy path, attribute its callers

Rasbandit:
## Why

Prod 2026-08-13. A device login *looked* like an incident: `POST 428` every 5
seconds in the warn stream, from a caller we could not identify. I reported it
to the user as "a device is stuck, logged out."

It wasn't. Tracing it end to end showed a completely **normal, successful
login**:

| Time (UTC) | Event |
|---|---|
| 19:27:32 | `POST /api/auth/device` — code minted |
| 19:28:37 → 19:33:12 | poll every 5s, each `428 authorization_pending` |
| **19:33:17** | poll **succeeded** — that trace writes `device_refresh_tokens` |
| 19:48:17 | `POST /api/auth/token/refresh` — device healthy since |

Nothing was broken. The telemetry lied, for three separate reasons.

## 1. The happy path logged at `:warning`

`authorization_pending` means the code is alive and the human has not clicked
approve yet. `RequestLogger` maps every 4xx to `:warning`, so with a **300s
code TTL** and a **5s advertised poll interval**, one successful login emits up
to **60 warnings**. In the observed window that was **~82% of the entire warn
stream**, and it feeds the already-noisy `loki-auth-failure-burst` alert.

A log level is the claim *"a human should look at this."* Spending it on a
happy path is how a working system reads as an incident — as this one did, to
me, in this session.

Controllers can now mark a response `:expected_client_status` and
`RequestLogger` emits it at `:info`. **Opt-in per response, set server-side** —
never blanket-4xx and never keyed on route+status, so a genuine client error on
the same endpoint still surfaces at `:warning`. 5xx is never downgradable.

## 2. The endpoint was unattributable

`/api/auth/device/token` is public and pre-auth, so `user_id` is null by design
and the log line carried no IP and no user agent. There was no way to trace
polling to anyone — not to answer "who is calling this", and not to see
device-code guessing if it ever happened.

Device-auth request logs now carry `user_agent`, **scoped to that controller**
so every other log line in the system stays the same size.

**No client IP — deliberately, and this changed mid-PR.** My first pass logged
one. Proving the change against a real dev server surfaced that `config/prod.exs`
states a standing policy in its own comment: *request headers and client IPs
stay OUT of Loki.* That is the stated reason prod excludes
`Sentry.PlugContext`'s `:__sentry__` blob, and `Engram.Sentry.Scrubber` drops
`event.request` wholesale on the same grounds.

The codebase does record client IPs — as purpose-built DB columns with a
lifecycle (`onboarding/agreement.ex` `ip_address`, `oauth_register_controller`
`first_ip`) — not sprayed across a hosted, long-retention aggregator. So the IP
is gone and **a test now pins its absence** so a future "just add the IP" cannot
land silently. If per-source correlation is ever needed for device-code
guessing, the answer is a keyed digest via `Engram.Crypto.HMAC`.

## 3. `428` was never the spec status

RFC 8628 §3.5 returns `authorization_pending` as a **400** with the error in the
body. `428 Precondition Required` is not part of the device flow. Flipped.

**Nothing strands.** Every consumer of the old status was inventoried:

| Consumer | Before | After |
|---|---|---|
| `device_auth_controller.ex` | returns 428 | returns 400 |
| `device_auth_controller_test.exs` | asserts 428 | asserts 400 |
| `e2e/helpers/device_flow.py` | `== 428` | `in (400, 428)` |
| `e2e/tests/api_only/test_71_connections.py` | `== 428` | `in (400, 428)` |
| plugin `device-flow-modal.ts` | `=== 428` | `400 \|\| 428` (Engram-obsidian#424) |

Already-installed plugin builds are unaffected: an unrecognised status falls
through to "keep polling", so the loop continues and the 2xx branch still
completes the login. The e2e helpers accept **both** so they work against either
side of a paired backend/plugin rollout.

## Tests

TDD, three watched red first:
- `returns authorization_pending for pending code` — failed `got: 428`
- `a device-flow authorization_pending poll logs at :info` — failed `left: :warning`
- `device-auth requests carry the user agent` — failed `left: nil`

Two boundary guards **passed from the start** and must keep passing:
- `a 4xx with no expected-status marker still logs at :warning`
- `non-device-auth requests omit the user agent`

`mix format` + `mix credo --strict` (no issues) + `mix dialyzer` (clean) +
`mix test --warnings-as-errors`: **4321 tests, 0 failures**.

## Noticed, deliberately not fixed here

`onboarding_controller.ex:74` and `oauth_token_controller.ex:16` stamp
`RequestMeta.format_ip(conn.remote_ip)` onto persisted rows. Per `RemoteIp`'s
own moduledoc that is the **ALB's private IP** in prod, not the client's — so
ToS-agreement and OAuth-token provenance may be recording the load balancer.
Out of scope for this PR; worth its own issue.

## Verified against a real dev server

Not just unit tests — `make saas-dev BACKEND_DIR=<this worktree>`, with the
Phoenix pid's `/proc/<pid>/cwd` checked against the worktree so a port
answering 200 could not be mistaken for "serving my code":

| Case | HTTP | Log level |
|---|---|---|
| before (`main` @ 8b8ddadb), pending poll | `428` | `[warning]` |
| after, pending poll | **`400`** | **`[info]`** |
| after, bogus device code — real client error, same route | `410` | **`[warning]`** |

That last row is the one that matters: the downgrade is scoped to the marked
response, not to the route or to 4xx in general.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -194,6 +194,7 @@ config :logger, :default_formatter,
     :topic,
     :total_count,
     :trace_id,
+    :user_agent,
     :user_id,
     :vault_id,
     :worker
--- a/e2e/helpers/device_flow.py
+++ b/e2e/helpers/device_flow.py
@@ -39,7 +39,11 @@ def exchange_device_code(base_url: str, device_code: str) -> dict | None:
     """Try to exchange a device code for tokens.
 
     POST /auth/device/token with device_code.
-    Returns token dict on 200, None on 428 (pending), raises on other errors.
+    Returns token dict on 200, None while pending, raises on other errors.
+
+    Pending is 400 per RFC 8628 §3.5. 428 is the pre-2026-08 status this
+    endpoint used; still accepted so the helper works against either side of
+    a paired backend/plugin branch rollout.
     """
     resp = requests.post(
         f"{base_url}/auth/device/token",
@@ -48,7 +52,7 @@ def exchange_device_code(base_url: str, device_code: str) -> dict | None:
     )
     if resp.status_code == 200:
         return resp.json()
-    if resp.status_code == 428:
+    if resp.status_code in (400, 428):
         return None
     raise RuntimeError(
         f"Device code exchange failed: HTTP {resp.status_code}\n{resp.text[:500]}"
--- a/e2e/helpers/device_flow.py
+++ b/e2e/helpers/device_flow.py
@@ -35,6 +35,18 @@ def start_device_flow(base_url: str, client_id: str) -> dict:
     return data
 
 
+def _is_pending(resp) -> bool:
+    """True when a 4xx body is the device flow's `authorization_pending`.
+
+    RFC 8628 puts the discriminator in the body, not the status line. Anything
+    unparseable is NOT pending — fail loudly rather than poll forever.
+    """
+    try:
+        return resp.json().get("error") == "authorization_pending"
+    except ValueError:
+        return False
+
+
 def exchange_device_code(base_url: str, device_code: str) -> dict | None:
     """Try to exchange a device code for tokens.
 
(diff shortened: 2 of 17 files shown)


Rasbandit:
Plugin fixes merged as engram-app/Engram-obsidian#424 (7a909416). Re-running e2e: `test_89/90/91` were red-by-design against the pre-fix plugin main and should now go green.

PR #1376, opened 2026-08-14
test(e2e): a closed sync gate must hold the feed, not consume it

Rasbandit:
## What

One e2e test, `test_89_closed_gate_holds_feed.py`, for the 2026-08-13 first-sync failure: every folder created, zero notes, "success" recap at 100%.

## Why this layer

The plugin's own suite was fully green the entire time that bug was in production. Unit coverage landed in engram-app/Engram-obsidian#424 (22 tests, red-first), but the loop that actually broke crosses a real Obsidian, a real backend and a persisted cursor, so it needs pinning here too.

## The assertion that carries it

"Nothing on disk while the gate is closed" is equally true of the broken build, so on its own it proves nothing. What separates fixed from broken is the **catch-up cursor**: `walkOpLog` used to persist it per page regardless of the gate, so a blocked walk advanced past every row it had refused to write and the next catch-up resumed after them. The rows were skipped, not delayed.

So the test asserts, in order:
1. gate closed + 5 new server notes + full sync → nothing on disk
2. **cursor did not move**
3. gate open + full sync → all 5 materialise

## Red/green

This branch has no matching plugin branch, so e2e auto-links plugin `main` — the pre-fix build. **This run is expected to fail on step 2 or 3.** That failure is the proof the test catches the bug. It should go green once engram-app/Engram-obsidian#424 merges into plugin main; I'll re-run then and will not merge this on red.

Rerun-safe: run-unique paths (session-scoped vault survives runs) and a best-effort server-side delete in a `finally` that can't mask a real assertion.

Refs engram-app/Engram-obsidian#424

### DIFF
--- a/e2e/tests/test_89_closed_gate_holds_feed.py
+++ b/e2e/tests/test_89_closed_gate_holds_feed.py
@@ -1,0 +1,92 @@
+"""Test 89: a closed sync gate HOLDS the catch-up feed, it does not consume it.
+
+Prod, 2026-08-13. A first sync into an empty vault created every folder and
+zero notes, then reported success and 100% downloaded. Two defects stacked:
+
+  1. `flushFromCrdt` returned true while the gate was shut, having written
+     nothing. The replay counted those phantom writes as downloaded files, so
+     the progress bar filled, the recap said "success", and the
+     produced-nothing anomaly could not fire.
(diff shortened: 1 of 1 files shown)


Rasbandit:
Folded into #1373 as cc60121d — one backend PR for the whole first-sync initiative instead of two.

Issue #1154, opened 2026-07-31
index-crdt 5: per-folder index sharding — cost it, then decide


Rasbandit:
Decision issue arising from #1149, to be settled during #1150 rather than after.

## What changed

Sharding was originally noted as a possible memory optimisation. The spike showed it cuts **two**
costs, which makes it worth costing properly:

1. **Resident memory** — 39 MB for a 50k-note vault index, vs a fraction of that per folder shard.
2. **First-sync bandwidth** — a fresh client on a churned 10k-note vault pulls ~2 MB before it can
   trust its identity map. syncStep1/2 makes that a diff for a returning client; a first sync or a
   wiped client pays it in full.

## The cost side

More rooms (back toward the per-note room count that #896 worries about), and cross-shard renames
become a two-shard transaction — which is exactly where a CRDT index is supposed to be simple.

## Decide during #1150

The room-key shape is chosen there. Changing it afterwards is a migration.

Refs #1146, #1149, #896

Issue #1153, opened 2026-07-31
index-crdt 4: index compaction — design alongside the #958 checkpoint-union hazard


Rasbandit:
Fourth implementation step for #1146.

## Why it is not optional

#1149 measured churn growth: **1.26x after only 5 rename rounds on 20% of paths**, and it
accumulates rather than plateauing. That is a few weeks of normal use, not a worst case. Yjs keeps
tombstones for deleted map keys, so a vault that renames regularly grows its index forever.

## The hazard

This is the same shape as #958 — checkpoint union double-content after a flatten. Whatever compaction
the index gets must be designed **with** that, not bolted on afterwards, or we reproduce a bug we
already know how to cause.

## Scope

- A compaction/flatten policy for the index doc.
- Proof that a compacted index cannot double or lose entries when two devices hold divergent
  lineages across the flatten.
- Ideally a sim-tier test, since this is exactly the class the plugin's `tests/sim/` tier exists for.

Refs #1146, #1149, #958

Issue #1152, opened 2026-07-31
index-crdt 3: idle-exit timer + resident-room LRU (blocks everything downstream)


Rasbandit:
Third implementation step for #1146 — and **the riskiest one**, because nothing in the system works
this way today.

## Why it is mandatory

#1149 measured the index doc at **7.91 MB resident** for a 10k-note vault (~0.79 MB per 1,000
notes, linear, stable to 1%). Against the 1024 MB task with no per-container limit (#895), ~128
resident rooms consume the entire task. At a target of thousands of concurrent vaults, resident-
per-connection misses by two orders of magnitude.

## Scope

- **Idle-exit timer.** Note rooms use `auto_exit: true` = exit on last observer. An index room has
  an observer for the whole time any client is connected, so that is session-length. Needs a timer
  that exits **with observers still attached**. This is new behaviour, not a config flag.
- **Client re-spin** on the next mutation burst, and on connect.
- **Resident-room LRU** as a backstop, so a pathological vault mix cannot pin memory.

## The constraint that shapes it

The drift-free property only holds if the client reconciles against the **CRDT**, not a
projection. A client cannot read the manifest and skip the room — that is the current bug,
reintroduced. So this is one room spin per **connect**, not per mutation. Connect rate is the load
figure to design against, not concurrent-connection count.

## Do this early

Steps 2 and the whole client side depend on the lifetime model being right. Finding out it is
harder than expected is far cheaper now than after they are built on top of it.

## Done when

A vault's index room spins on connect, completes its handshake, checkpoints, exits, and re-spins
correctly on the next mutation — with resident room count bounded under a soak.

Refs #1146, #1149, #895

Rasbandit:
## Implemented in #1382 — but NOT as written here. Read this before following the scope above.

> **Idle-exit timer.** … Needs a timer that exits **with observers still attached**.

**Implemented literally, that ships silent data loss.** A client edit is a `sync_update` frame, which y_ex dispatches as a `GenServer.cast` (`deps/y_ex/lib/server/doc_server_worker.ex:26`) — and a cast to a dead pid returns `:ok` while dropping the frame. `crdt_channel.ex` already carries the warning at its room monitor:

> Without this, send_yjs_message casts to a dead pid return `:ok` and every subsequent edit is silently dropped.

Today that only happens on a **crash**. Exiting a room out from under an attached observer would make it happen **on a timer, on purpose** — inside the very epic (#167) whose purpose is eliminating silent drift.

[...]

Issue #1151, opened 2026-07-31
index-crdt 2: index checkpoint — encrypt state + project to notes rows


Rasbandit:
Second implementation step for #1146. Depends on step 1.

## Scope

Mirror `CrdtCheckpoint` for the index room:

- Encrypt the index Yjs state (same envelope as `notes.crdt_state`).
- **Project** entries into the existing `notes` path columns — `path_ciphertext` / `path_nonce` /
  `path_hmac` (`note.ex:56-57`).

The projection is the whole point: it is what keeps the REST + search + MCP façade working
against a client-owned CRDT, exactly as `CrdtCheckpoint`'s moduledoc describes for note content.
The server never reads CRDT state to answer a path lookup.

## Encryption

No new design. Same DEK, same envelope, same HMAC-for-lookup pattern already used for note paths.

## Sizing note from #1149

A churned 10k-note index encodes to ~2.0 MB. That is the `crdt_state` blob size, and it is ~1/4
of the doc's resident cost — do not size anything else from it.

## Done when

An index room checkpoint produces a decryptable `crdt_state` blob AND leaves the projected path
columns queryable by the existing manifest endpoint.

Refs #1146, #1149

Rasbandit:
Done-when is met on both halves:

- **Decryptable checkpoint blob** — #1387 persists the encrypted `filemeta_v0` snapshot, #1392 adds the per-update tail log so an ungraceful room death no longer loses claims.
- **Projected path columns queryable** — #1388 made the map authoritative and added `ProjectVaultIndex`, which derives `notes.path_*` from it, so the existing manifest/REST/search façade answers from the rows exactly as before.

Residency (the follow-up this issue kept pointing at) is now closed separately by #1152 via #1394. Closing.

PR #1383, opened 2026-08-15
feat(crdt): per-vault index room + channel plumbing (index-crdt 1)

Rasbandit:
## Summary

- **index-crdt 1 (#1150)**, under the engram-workspace#167 p0: `{:global, {:crdt_index, vault_id}}`, one `Y.Map` named `filemeta_v0` (`path -> %{note_id, type, hash}`), riding the existing per-vault `crdt:` channel. The substrate for making identity converge through the CRDT instead of a REST projection plus a client-side map.
- **Deliberately inert** — the room exists, syncs and can be observed. Nothing writes to it, nothing reads it back. Checkpoint/projection is #1151; client adoption is Engram-obsidian#362/#363.
- The `filemeta_v0` name matches Relay's `SyncStore.ts:20` on purpose: the shape is the part of their design that removes our drift class, so keeping the name makes the correspondence checkable rather than folkloric.

## The trap this surfaced: #1150 and #1152 combine badly

Draining a **note** room is lossless *only* because `terminate/2` → `CrdtPersistence.unbind/3` checkpoints on the way out. The index room has **no persistence until #1151** (`CrdtIndexPersistence.bind/3` is a no-op — `bind/3` is the only required callback).

So giving the index room `idle_exit_ms` would exit it and **evaporate the entire index**. The drain would be working perfectly; it just has nothing to save here. Nothing in #1152's own suite could catch it, because from the drain's side the behaviour is identical.

**The index room therefore runs no `CrdtCheckpointTimer` and sets no `idle_exit_ms`,** with a test pinning it. Residency stays bounded the old way (`auto_exit` on last observer), which is adequate precisely because the room holds nothing durable. Enabling the drain is a #1151 follow-up, not a knob to flip here.

## Wire

- `crdt_index_msg` on the existing channel — **no new transport**, per the issue.
- **No `doc_id`**: the vault is implicit in the topic and there is exactly one index room per connection. A doc_id-addressable index room would also let a client drive it through the note path and bypass `note_in_vault?` — pinned by a test asserting `crdt_msg` with the vault id replies `note_not_found`.
- **Handshake rate bucket, not edit.** Index sync is once-per-connect; billing it to the edit budget would let it starve real edits (the 2026-07-07 cross-file-overwrite incident shape).
- `handle_info({:yjs, …})` checks `index_room` **first**, so an index room can never be mistaken for a note room whose pid was reused; the `:DOWN` handler clears the cache, or the next frame casts into a corpse and returns `:ok`.
- `CrdtIndexRegistry` reuses `CrdtRegistry.observe_with_retry/3` rather than re-rolling the auto-exit race — that helper is public and injectable for exactly this reason.

## Test plan

- [x] `mix format` + `credo --strict` (0 issues) + `dialyzer` (passed) + full suite
- [x] Registry: cluster-wide singleton, per-vault isolation, `lookup/1` does not spin a room
- [x] Sync: two clients converge via syncStep1/2; the map survives observer churn (first observer leaves after a second joins, so the room genuinely outlives it)
- [x] Channel: step1 → step2 reply, cross-client convergence, index unreachable via `crdt_msg`, handshake-bucket billing
- [x] Inertness: no checkpoint timer linked to the room; `restart: :temporary`

**Two of these tests were vacuous when first written**, both caught by asking "would this go red if the implementation were wrong?" and then confirmed by mutating the implementation:

| test | was | mutation → red |
|---|---|---|
| no-drain | asserted over the `opts` the test itself passed in | wire a `CrdtCheckpointTimer` in |
| rate bucket | `assert_push`, but the first frame succeeds under either bucket | `check_rate(socket, :edit)` |

Both mutations reverted after confirming red.

## One unreproduced failure, disclosed rather than buried

The first full-suite run had **1 failure: `Engram.Vector.QdrantHybridTest`** — `Req.TransportError{reason: :timeout}`. It did **not** reproduce: green in isolation (3 seeds), green in isolation under heavy CPU load (5 seeds), and green on a full-suite re-run.

Why I do not believe it is this PR: these changes add no application children, no ETS, no boot-time work, and nothing that executes during a Qdrant test. The test is `async: true` using **Bypass** — a real local HTTP server with a `Req` timeout, one of 47 such tests, racing 4400 others. That is a contention-sensitive pattern.

A `main`-baseline run is in flight to confirm it is pre-existing; I will comment with the result either way rather than let a rerun-pass stand as the answer.

## Not in scope

Checkpoint/projection (#1151, also what unblocks the drain here), compaction (#1153, entangled with the #958 hazard), sharding (#1154), client adoption (Engram-obsidian#362/#363). **#167 does not close until #363** — everything before it is scaffolding, and the p0's original trigger (316 notes not materialising in a prod first-sync) is still unrooted.

Design + traps: `docs/context/crdt-index-room.md`.

Refs #1150, #1146, #1152, engram-app/engram-workspace#167

### DIFF
--- a/ci/compose.yml
+++ b/ci/compose.yml
@@ -163,6 +163,19 @@ services:
       # Same CI=true gating (runtime.exs) keeps prod limits un-weakenable.
       CRDT_MSG_RATE_LIMIT_OVERRIDE: "100000"
       CRDT_HS_RATE_LIMIT_OVERRIDE: "100000"
+      # Room drain (#1152) turned ON for e2e only, under the same CI=true gate.
+      # In production NOTHING sets idle_exit_ms, so the drain never fires and is
+      # therefore never exercised against a real client — every unit and channel
+      # test stands a test process in for the plugin. Enabling it here means the
+      # whole Obsidian suite runs with rooms being released out from under the
+      # client mid-session, which is the only way to prove the property the
+      # design rests on: the client re-spins on its next frame and loses nothing.
+      #
+      # 5s is short enough that rooms genuinely drain between test steps and long
+      # enough that it is not pathological. If this destabilises the suite, that
+      # is a REAL finding about the drain, not CI noise — raise it deliberately
+      # rather than deleting the line.
+      CRDT_IDLE_EXIT_MS: ${CRDT_IDLE_EXIT_MS:-5000}
       # Free-tier launch test_72_cancel_to_free_overlimit signs a synthetic
       # subscription.canceled webhook with this fixed secret so the e2e
       # exercises the real verify_signature → upsert_from_paddle_event path
--- a/ci/compose.yml
+++ b/ci/compose.yml
@@ -176,6 +176,17 @@ services:
       # is a REAL finding about the drain, not CI noise — raise it deliberately
       # rather than deleting the line.
       CRDT_IDLE_EXIT_MS: ${CRDT_IDLE_EXIT_MS:-5000}
+      # LRU backstop, also e2e-only. Idle-exit alone never evicts a room that
+      # stays busy, so without these the LRU only ever ran against spawn stubs
+      # in unit tests — it had never touched a real SharedDoc.
+      #
+      # A tiny cap plus a fast sweep is the point: e2e vaults hold few notes, so
+      # a production-shaped cap would never be reached and this would cover
+      # nothing. The sweep interval matters as much as the cap — it defaults to
+      # 30s, and most e2e tests finish inside that, so a low cap alone would arm
+      # the LRU and never fire it.
+      CRDT_MAX_RESIDENT_ROOMS: ${CRDT_MAX_RESIDENT_ROOMS:-3}
+      CRDT_LRU_SWEEP_MS: ${CRDT_LRU_SWEEP_MS:-3000}
       # Free-tier launch test_72_cancel_to_free_overlimit signs a synthetic
       # subscription.canceled webhook with this fixed secret so the e2e
       # exercises the real verify_signature → upsert_from_paddle_event path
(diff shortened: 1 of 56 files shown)


Rasbandit:
## Baseline result: the QdrantHybridTest failure is NOT this PR — and it corrected a claim in #1382

Ran the full suite on plain `main` (at 21f89023, the #1382 merge) with none of this PR's changes. It **also failed** — on a *different* timing-sensitive test, `FanoutPacerTest`.

So the suite carries load-sensitive tests independent of any diff here. Characterised:

| test | shape | observed |
|---|---|---|
| `FanoutPacerTest` | 200 ms pacing assertions | **4/6 under CPU load**, no cluster tests involved |
| `QdrantHybridTest` | `async: true` + Bypass HTTP + `Req` timeout (1 of 47 Bypass tests) | 1 full-suite run; green in isolation ×3, green in isolation under load ×5, green on re-run |

### This also retracts something #1382 shipped

#1382's context doc claimed a controlled comparison proved that adding `CLUSTER_TESTS=1` to the `unit-tests` job breaks `FanoutPacerTest` — without the flag 4368 tests / 0 failures, with it 4371 / 1 failure. The mechanism was plausible (ClusterCase starts a second `Phoenix.PubSub` under the same name on the peer and connects the nodes, so a 200 ms pacing assertion would not survive cross-node fan-out).

**It was one observation against an unknown baseline.** Now that `FanoutPacerTest` is measured at 4/6 under load with no cluster tests anywhere, that single failure cannot be attributed to distribution. The experiment proves nothing either way.

The retraction is in this PR (`docs(context): retract the cluster-CI causal claim`), along with the flake table above so the next person budgets for these instead of attributing a red run to their diff.

**Consequence for the open cluster-CI decision:** the in-job approach is *unresolved*, not ruled out. The separate `--only cluster` job remains safe by construction (green 5/5 in isolation), but it is no longer the only option on the table.
EOF

Issue #1146, opened 2026-07-31
[decision] Make the vault file index a CRDT — retire the REST manifest + seq-cursor drift class


Rasbandit:
From the Relay audit (`docs/context/relay-pattern-audit.md` in engram-workspace). Relay finding **A** — the
one structural difference that actually matters.

## The difference

Relay stores the path→file mapping **inside the folder's synced Y.Doc**
(`Relay/src/SyncStore.ts:20`):

```ts
this.legacyIds = this.ydoc.getMap("docs");        // legacy: path -> guid
this.meta      = this.ydoc.getMap("filemeta_v0"); // path -> { id, type, mimetype, hash }
```

Identity, folder structure, mimetype and content hash converge through **the same channel as content**.
There is no reconcile path because there is no second source of truth.

We have the opposite: identity is `NoteIdMap` (a plain `Map<path,id>` in `data.json`) reconciled against a
REST manifest, with a separate seq-cursor feed for catch-up.

## What that costs us today

Everything below exists *only* to keep the client's id map honest against server rows. All in
`plugin/src/sync.ts` unless noted:

- `reconcileNoteIdMapFromManifest`, `ensureNoteIdMapped`, `cacheManifestOwners`
- `validateFromManifest`
- `catchupViaSeqReplay`, `cursorAdvances`, `scheduleSeqHeal`, `getCatchupSeq`/`setCatchupSeq`/`getCatchupId`/`getManifestSeq`
- `socketConverge`, `fireCrdtReHandshake`, `commitCrdtConvergence`, `releaseHealRoom`, `healNoteOnOpen`
- `applyLiveOpWithSeq`, `discoverAnnouncedNote`, `materializeEmptyDiscovered`
- `drainStrandedFlushes` + `clearStrandHealAttempts` (`plugin/src/crdt/wiring.ts`)
- the `crdt_note_not_found` wire message — a protocol message that exists because the client's id map can
  disagree with the server's rows

Incidents traceable to this one root (*identity and content converge through different channels at
different rates*):

| Incident | Ref |
|---|---|
| Manifest-validator stall — cursor frozen, "N rows still behind" | plugin#327 |
| Seq-fence `<=` drops unseen content | plugin#296 |
| Direction-blind catch-up backfill reverting live content | #1058 / plugin#278 |
| Stale `NoteIdMap` breaks sync | context doc `noteidmap-stale-breaks-sync.md` |
| `note_not_found` cross-wire on create race | #955 / plugin#202 |
[...]

Issue #844, opened 2026-07-02
flaky e2e-crdt: test_concurrent_edits_both_survive — FROM_B not propagated within 30s


Rasbandit:
Main run 28576184072 (post-merge of frontend-only #842, no sync code touched):

```
FAILED tests/crdt/test_crdt_sync.py::test_concurrent_edits_both_survive - TimeoutError: File E2E/Crdt/Merge.md did not contain 'FROM_B' within 30s
1 failed, 5 passed, 1 rerun in 123.71s
```

Failed its in-run retry too. Same 08:2x UTC runner-contention window that produced postgres boot races (e2e-api/e2e-local) and the SPA CRDT interleave flake (#843) on PR #842's runs — likely load-dependent propagation latency, but 30s is generous, so a real B→A delivery stall under contention is worth ruling out. Related: #669 (sync hardening test pass), #843 (SPA-side CRDT e2e flake, assertion design).

Filed per the no-silent-rerun rule; rerun triggered on the same commit to gather a pass/fail data point.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Rasbandit:
Rerun of the identical commit passed (run 28576184072 now fully green incl. e2e-crdt) — consistent with the contention-window theory. Keeping open for the recurrence pattern / root-cause per the no-silent-rerun rule.

Rasbandit:
Another occurrence: run 28624031576 (PR #876, e2e-harness-only diff — Clerk sweep scoping + retry budget; no sync-path changes). Same signature: test_concurrent_edits_both_survive, FROM_B not in Merge.md within 30s.

Rasbandit:
### Root-cause datapoint: a *deterministic* repro of this symptom (not latency) — history-less first-delivery → keep-both

While reviewing the vault-channel CRDT fan-out work (backend #997 / plugin #233), this test failed **deterministically** on run `29178486814` (backend `917516c0`):

```
FAILED tests/crdt/test_crdt_sync.py::test_concurrent_edits_both_survive
  - TimeoutError: File E2E/Crdt/Merge.md did not contain 'FROM_B' within 30s
```

The client log for `E2E/Crdt/Merge.md` shows the actual failure mode is **not** delivery latency — it's a conflict resolution:

```
dev=8b288cab [client:conflict] history-less drift → keep-both | original=E2E/Crdt/Merge.md
```

[...]

PR #1399, opened 2026-08-17
fix(crdt): crdt_create is the web app's rename, so it must claim (+ the e2e that proves the client half)

Rasbandit:
The test the index-CRDT effort was missing. Every existing test has our client on both ends: the plugin's unit tests wire two client rooms to each other, and the backend's e2e drives a *synthetic* client through the real channel. Neither has ever asked whether the plugin's frames are something the server accepts — which is how a write-only client shipped through three green CI runs.

Two tests, both green against a real backend + real headless Obsidian:
- a create claims its path in `filemeta_v0`
- a rename moves the claim **and releases the old one**

### It had never been run
The first version was written but never executed, and it had three defects — all in the RPC helper, all of which raised before any assertion ran:

1. `vaults` is one of `Repo`'s `@tenant_tables`, so the Ecto query tripped the tenant guard.
2. The ordering column is `created_at`, not `inserted_at`.
3. "Newest vault" was the wrong handle anyway — vault names are encrypted, and a session may hold several.

It now takes the `sync_vault_id` fixture and resolves `user_id` with raw SQL, which `prepare_query/3` does not rewrite. `load_doc/2` establishes tenant context itself, so that lookup is the only thing needed ahead of it.

Reading through `load_doc/2` rather than the checkpoint row is deliberate: it's the same snapshot+tail recipe `bind/3` uses, so a claim that only reached the tail log still counts and the test doesn't pass/fail on checkpoint timing.

Refs engram-app/Engram#1146, engram-app/Engram-obsidian#362

### DIFF
--- a/docs/context/crdt-create-is-a-rename.md
+++ b/docs/context/crdt-create-is-a-rename.md
@@ -1,0 +1,88 @@
+# Context Doc: `crdt_create` is the web app's rename, so it must claim
+
+_Last verified: 2026-08-17 (Engram #1400, fixed on the #1399 branch)_
+
+## Status
+Fixed. `genesis_crdt_note/5` now claims before the row transaction.
+
+## What This Is
+
+`crdt_create` reads like a create. It is also **the web app's only rename and
+move path**, and that is not obvious from either side of the wire:
+
+- `frontend/src/api/queries.ts` — `useRenameNote` calls `crdtCreateNote(id, new_path)`,
+  commented *"Replaces POST /notes/rename"*.
+- `frontend/src/viewer/folder-tree.tsx:147` — *"A CRDT move = crdt_create per id
+  at its NEW path"*.
+
+Server side it lands in `Notes.genesis_crdt_note/5` → `genesis_relocate_live/6`
+→ `move_note/5` (Phase E2, "rename-as-move"). That chain moved the row and
+claimed nothing in `filemeta_v0`.
+
+## Why that was a latent data-loss bug
+
+`filemeta_v0` is the authority for paths; `ProjectVaultIndex` derives
+`notes.path_*` **from** it, and it *"walks the INDEX ENTRIES and corrects the row
+each one names"*. So:
+
+1. A client claims `note → A.md`.
+2. The user renames it to `B.md` in the web app. Row moves; the entry still says `A.md`.
+3. The index room checkpoints, which enqueues `ProjectVaultIndex`.
+4. Projection reads the entry and repaths the row **back to `A.md`**.
+
+The rename silently reverts — and because projection re-encrypts the path,
+rewrites `path_hmac`, repaths Qdrant and enqueues link rewrites, it is a full
+write, not a display artifact.
+
+`identity.ex` already named this exact failure: *"a rename that moves and a claim
+that does not gets REVERTED by the next projection run."* `claim_rename/5` was
+built to prevent it on the REST path. The CRDT leg simply never got the same
+treatment, and stayed invisible because **no shipped client wrote the index** —
+until Engram-obsidian#362.
+
+## The fix, and why it is deliberately narrow
+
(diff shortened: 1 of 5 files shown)


Rasbandit:
### The first e2e-crdt red here was a pairing miss, not a test defect

`Index currently claims: []` — the server index was **empty**, because the backend e2e ran against plugin `main`, which has no index-CRDT client at all. That code is on the unmerged plugin PR engram-app/Engram-obsidian#431.

Per `oauth-e2e-pairing-and-token-binding.md`, priority-3 auto-pairing is a **branch-name match**: this backend branch is `test/e2e-index-crdt-client` and the plugin PR is `feat/index-crdt-client`, so nothing paired and CI silently tested the plugin's old behaviour. Exactly the "confusing red that looks like a broken test but is really 'you didn't pair them'" the doc describes.

Pushed a plugin pairing branch `test/e2e-index-crdt-client` at the PR head (`5135793`). It is a **snapshot, not a living branch** — it needs re-pointing if #431 gains commits.

**Merge order: plugin #431 first**, so plugin `main` carries the client; this PR then goes green against `main` on its own and the pairing branch gets deleted.

Both tests pass locally against a real backend + real Obsidian with the paired plugin build.

Rasbandit:
### Paired run: both new tests PASS in CI

[Run 32002060146](https://github.com/engram-app/Engram/actions/runs/32002060146), with `PLUGIN_SHA: 5135793efb9c…` — the pairing took effect and CI built the plugin PR's code:

```
tests/crdt/test_index_crdt_client_writes.py::test_client_rename_reaches_the_server_index PASSED
tests/crdt/test_index_crdt_client_writes.py::test_a_created_note_is_claimed_in_the_server_index PASSED
1 failed, 19 passed in 210.30s
```

This is the first CI-verified proof that the plugin's `filemeta_v0` writes are something the server accepts — previously every test had our client on both ends.

The one remaining failure is `test_cold_send_over_fanout_opens_no_room`, the known flake tracked in #844. Worth stating plainly: **this PR's whole diff is one new test file**, so that failure has no code change to attribute it to. Evidence posted on #844.

Rasbandit:
### Scope grew: this now carries the #1400 fix as well

Asking "does the index CRDT need anything in the web app?" turned up a real latent bug, so this PR now has two halves.

**1. The e2e** (as before) — proves the plugin's `filemeta_v0` writes are something the server accepts. Verified green in CI against the paired plugin build; plugin #431 is now merged, so this should go green against plugin `main` on its own and the pairing branch has been deleted.

**2. The fix for #1400** — `crdt_create` is the web app's **only** rename/move path (`useRenameNote` → `crdtCreateNote`, "Replaces POST /notes/rename"), and it relocated the row without claiming. Since `ProjectVaultIndex` walks the entries and corrects the row each one names, a web rename of a claimed note was dragged back to its old path. `identity.ex` already named this failure; `claim_rename/5` prevents it on the REST path and the CRDT leg never got the same treatment. Latent only while no client wrote the index — #431 is that client, which is why this lands now.

The fix is deliberately narrow (only a live id → free path claims), because a claim cannot be rolled back and claiming for an operation that does not happen makes a path permanently unclaimable. Rationale in `docs/context/crdt-create-is-a-rename.md`.

**Gauntlet:** format, credo --strict, dialyzer, and the full suite — **4562 tests, 0 failures**, run sequentially.

**Known cost, flagged rather than buried:** one indexed PK read per `genesis_crdt_note/5`; `crdt_create_batch` shares the function, so a large first sync pays it per note. Unmeasured. Every cheaper variant I considered either claims inside the row transaction (no durability guarantee, per `Identity`) or claims after the row moves (reintroduces the bug). Happy to restructure if the cost matters.

**Not audited:** `crdt_delete` plausibly owes an `Identity.release` for the same reason — a tombstoned note whose entry survives keeps a path reserved that nothing can reuse.

PR #1422, opened 2026-08-19
perf(edge): preconnect to cross-origin hosts, make public JSON edge-cacheable

Rasbandit:
Origin-side half of an edge-performance pair. The Cloudflare half is **engram-app/engram-infra#1014** — neither half does much alone, see "Why two PRs" below.

### 1. Preconnect to the two cross-origin hosts

The SPA is on `app.engram.page` but cannot do anything without **two other origins**: `clerk.engram.page` (auth) and `api.engram.page` (every query + the WS). It discovers both **late** — `clerk-auth-provider` is a lazy `import()`, so the Clerk origin is unknown until the main bundle has downloaded *and* parsed.

Measured on prod before this change:

```
tcp=1.026s  tls=1.078s   ← cold
tcp=0.024s  tls=0.092s   ← warm
```

~95 ms per hop before byte one, ~1 s cold — all of it **serial dead time** on the sign-in path. A response `Link: rel=preconnect` header starts both handshakes while the bundle is still downloading.

Two deliberate choices:
- **`_headers`, not a `<link>` in `index.html`.** `index.html` is shared with the self-host build. A self-hoster who deliberately runs no Clerk should not open a connection to Clerk on every page load. Phoenix does not read `_headers`, so this is SaaS-only by construction.
- **`crossorigin` is required**, not decorative. Without it the browser opens an *unauthenticated* connection, then a second authenticated one for the real credentialed fetch — the hint buys nothing while costing a socket, and it looks correct in devtools.

### 2 & 3. Make the deploy-static public JSON cacheable

| Path | Size | Was |
|---|---|---|
| `/.well-known/oauth-authorization-server` | small | `DYNAMIC` |
| `/.well-known/oauth-protected-resource` | small | `DYNAMIC` |
| `/api/openapi` | **~72 KB** | `DYNAMIC` |

Every MCP client (Claude, Cursor, ChatGPT) probes the discovery documents on **every connect, before auth**. `/api/openapi` is re-rendered by OpenApiSpex on every request. All three are pure functions of (host, deploy) — I verified the discovery responses are byte-identical across requests.

Phoenix's default `max-age=0, private, must-revalidate` is what pinned them: **`private` forbids a shared cache from storing the response at all**, so Cloudflare passed straight through to Fargate every time.

## Correctness notes

- **The discovery bodies vary by `Host`** (`OAuthMetadata.base_url/1` resolves the dialed host against the `:cors_origin` allowlist). Cloudflare's cache key includes hostname, so `mcp.` and `app.` get separate entries and cannot be crossed. Called out in the code so nobody "optimises" it into a host-independent constant.
- **The CI OpenAPI drift gate is unaffected.** It generates the spec locally with `mix openapi.spec.json` and diffs the committed `openapi.json`; it never fetches this endpoint, so a cached copy cannot make it pass on stale content.
- **TTL is 5 minutes on all three, deliberately.** These carry `authorization_endpoint` / `registration_endpoint`; a client caching a stale endpoint after we move one fails auth in a way that looks like a client bug. Five minutes takes essentially all of the repeat-connect win while bounding staleness to something a deploy outruns.

## Why two PRs

The app header alone does nothing at the edge: Cloudflare's default Cache Level only treats known static **file extensions** as eligible, so extensionless JSON stays `DYNAMIC` regardless of what the origin sends. The Cache Rule in engram-infra#1014 grants eligibility; this PR decides the TTL. Merge order does not matter — neither half breaks anything on its own.

## Test plan
- [x] `mix format --check-formatted`
- [x] `mix compile --warnings-as-errors`
- [x] `mix credo --strict` — no issues
- [x] `mix sobelow --exit low` — clean
- [x] `mix test` — 1 property, **4635 tests, 0 failures**
- [x] New tests pin `cache-control` on all three endpoints
- [x] New test asserts **`/api/health` is NOT cached** — a cached health check reports healthy from the edge while the app is down, the classic broad-cache-rule outage

After both merge, expect `cf-cache-status: HIT` on a second request to the discovery docs and `/api/openapi`, and a `103` early hint on `app.engram.page`.

### DIFF
--- a/docs/context/edge-cache-request-varying-headers.md
+++ b/docs/context/edge-cache-request-varying-headers.md
@@ -1,0 +1,135 @@
+# Edge-caching a response whose HEADERS vary by request
+
+_Last verified: 2026-08-18 against prod (`api.engram.page`, `mcp.engram.page`,
+`app.engram.page`, `staging.engram.page`). Shipped in Engram#1422 +
+engram-infra#1014._
+
+## When you need this
+
+You are about to make something edge-cacheable — adding
+`Cache-Control: public` to a route, or adding a hostname or path to the
+Cloudflare Cache Rule in `engram-infra/main/cloudflare/cache.tf`.
+
+## The one idea
+
+"The body is identical for every caller" is **not** sufficient for a shared
+cache. The headers have to be identical too, and ours were not.
+
+`EngramWeb.Plugs.CORS` echoes the request `Origin` back whenever it is in the
+`:cors_origin` allowlist (`lib/engram_web/plugs/cors.ex`, `resolve_origin/1`).
+Verified against prod:
+
+```
+Origin: https://mcp.engram.page   -> access-control-allow-origin: https://mcp.engram.page
+Origin: https://api.engram.page   -> access-control-allow-origin: https://api.engram.page
+Origin: https://evil.example.com  -> access-control-allow-origin: https://app.engram.page
+```
+
+Cache that and the edge stores **one** entry with whichever origin happened to
+fill it. For the next 300s every other client gets that same header replayed,
+so a browser on `app.engram.page` receives a document stamped for
+`mcp.engram.page` and the fetch is CORS-blocked. Which client "wins" is a
+race, so the symptom is intermittent, region-dependent, and clears itself in
+five minutes — the worst possible bug report.
+
+**`Vary: Origin` does NOT fix this.** Cloudflare honours `Vary` only for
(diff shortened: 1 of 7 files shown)


Rasbandit:
Folded into #1407 so this lands as a single PR. The branch's commits are merged there, plus the review fixes (ACAO pinned to a constant on the cached documents, the duplicated cache-header plug collapsed into one `:public_cacheable` pipeline, and the crossorigin / Early Hints comments corrected).

PR #1425, opened 2026-08-19
docs(spec): decouple CRDT room creation from writes

Rasbandit:
## Summary

Spec only, no behaviour change. Proposes stopping the creation of a CRDT room as a side effect of receiving a write, so room count scales with the number of notes open in editors rather than with vault size.

`handle_in("crdt_msg", ...)` currently calls `ensure_room/2` for every content frame. The rule the system implements today is "a frame arrived, therefore spawn a stateful actor". A `SharedDoc` room is a collaboration session, and a bulk import has no collaboration in it.

> **Rev 2.** Rev 1 was reviewed adversarially and **did not survive**. The diagnosis was unaffected; all four defects were in the remedy. Rev 2 adds Phase 0 as a hard prerequisite and retracts rev 1's "do not add a lock" guidance, which was only valid with 0a in place. Review history is recorded at the end of the spec.

## Evidence

Measured on staging 2026-08-19 importing a 1,516-note vault, using the `[:engram, :crdt, :room_start]` counter added in #1424:

| | count | share |
|---|---|---|
| rooms created | 406 | 27% of notes |
| drained because they went idle | 166 | 41% |
| **force-evicted by the memory backstop** | **240** | **59%** |

Residency from the same run (`cap=64`) peaked at 316 and then decayed in exact `-16` steps, which is `@max_evictions_per_sweep` per 30s sweep. **The 316 peak is not demand, it is `CrdtRoomLru` unable to hold its own cap.** The LRU is load-bearing during normal operation, which means it conceals a creation bug rather than protecting against abuse.

Raising the eviction rate is explicitly rejected in the spec. It changes the graph and nothing else.

## Why the direction is right

`CrdtPersistence.update_v1/4` is append-only (`INSERT crdt_update_log`), Y-CRDT updates are commutative, and fan-out is already vault-channel based rather than room-observer based. `CrdtRegistry.terminate_room/1` confirms it independently: it brutally kills rooms and documents why that is safe, *"Nothing is lost: every room update is already in the durable tail-log."*

So a room is a **performance cache for hot notes, not a correctness requirement**.

## Phase 0 (hard prerequisites, found by review)

Commutativity makes append-versus-append safe. It says nothing about three mechanisms that silently assume the room is the **sole writer**:

- **0a (CRITICAL, silent data loss).** Live-room checkpoints pass no `:prune_ids`, so they take the watermark path, deleting every tail row `inserted_at <= wm` regardless of whether the room's doc folded it. Safe only under sole-writership. A row appended between a room's `bind` and its checkpoint is deleted unfolded and lost from both snapshot and tail. This is the #285 class through a different door. Fix: all checkpoints use exact `:prune_ids`, and the watermark branch is deleted so nothing can reintroduce the assumption.
- **0b (CRITICAL).** `head_marker` is compared by the client to *skip* convergence (`sync.ts:2107`). Concurrent detached writes compute markers from independent transient docs, producing sibling heads reflecting no real state, which lets a client skip convergence while genuinely behind (the deaf-note class). Fix: the detached path emits an explicit unknown. Never emit a claim of convergence you cannot prove.
- **0c.** Compaction was framed as tuning, but "self-limiting because hot notes have rooms" depends on the plugin's client-side `isLiveBound`, which the server does not control. Unenrolled repeat edits are O(n²) in tail depth with a full-snapshot decrypt per write. Fix: server-side mandatory trigger on folded row count.

## Phase 1: routing rule

| frame | resolved room | action |
|---|---|---|
| handshake | any | `ensure_room` (unchanged) |
| edit | live pid | relay to room (unchanged) |
| edit | none, or **dying pid** | **detached apply** (new) |

Rev 1 treated "room exists" as a boolean. `CrdtRegistry.lookup/1` returns raw `:global.whereis_name` and the module header warns it "can hand back a room that is mid-termination"; rooms use `auto_exit: true`, so termination is routine. The edit route needs `ensure_started`-style bounded-retry resolution returning a live pid or nothing, falling through to detached rather than relaying into a terminating room.

Fan-out is also promoted from background to a stated correctness invariant with a test: once writes bypass rooms, the vault-channel broadcast is the only delivery path for a client whose room bound before the append. #1424 already lost fan-out once while removing a room, and device B wrote a 0-byte file.

## Notes for review

- Adds an observability ask: a `source` label on `room_start`. The measurement above could not attribute which 27% of notes still demanded a room because the counter is unlabelled.
- Open item: the 0c threshold needs a number measured against real tail depths, not guessed.
- The spec belongs in the Engram vault under `50 Engineering/_Superpowers Specs/`. It is committed here temporarily because the engram MCP token was expired when it was written.

Refs #1146

### DIFF
--- a/docs/superpowers/specs/2026-08-19-detached-writes-room-decoupling-design.md
+++ b/docs/superpowers/specs/2026-08-19-detached-writes-room-decoupling-design.md
@@ -1,0 +1,227 @@
+# Decoupling CRDT room creation from writes
+
+**Date:** 2026-08-19
+**Repos:** `engram-app/engram` (backend), `engram-app/Engram-obsidian` (plugin, read-only impact)
+**Epic:** #1146 (identity/content converge through different channels)
+**Follows:** #1409 / #1424 (genesis seed detached), which piloted this pattern for note creation
+
+> **Placement note:** this belongs in the Engram vault at
+> `50 Engineering/_Superpowers Specs/`. It lives here temporarily because the
+> engram MCP token was expired when it was written. Mirror it to the vault and
+> delete this copy.
+
+---
+
+## Goal
+
+Stop creating a CRDT room as a side effect of receiving a write. Room count
+should scale with the number of notes open in editors, not with vault size.
+
+## The problem
+
+`handle_in("crdt_msg", ...)` calls `ensure_room/2` for every content frame. The
+rule the system implements today is:
+
+> a frame arrived, therefore spawn a stateful actor
+
+Room creation is triggered by **traffic**, not by **demand for collaboration**.
+A `SharedDoc` room is a live Y.Doc plus a checkpoint timer whose purpose is to
+let multiple clients converge on one note in real time. A bulk import has no
+collaboration in it: one writer, no concurrent peers. The correct number of
+rooms for an import is zero.
+
+### Evidence
+
+Measured on staging 2026-08-19, importing a 1,516-note vault, using the
+`[:engram, :crdt, :room_start]` counter added in #1424:
+
+| | count | share |
+|---|---|---|
+| rooms created | 406 | 27% of notes |
+| drained because they went idle | 166 | 41% |
+| **force-evicted by the memory backstop** | **240** | **59%** |
+
+Fifty-nine percent of the rooms created during a normal import were killed by
+`CrdtRoomLru` before they ever went idle. A room that gets force-evicted before
+it idles was created for work that never needed a session.
+
+Residency behaviour from the same run (`cap=64`):
+
+```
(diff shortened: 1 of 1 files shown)


Rasbandit:
Closing without merging: superseded by the canonical copy in the Engram vault at `50 Engineering/_Superpowers Specs/2026-08-19-detached-writes-room-decoupling-design.md`.

Project convention puts superpowers specs in the vault rather than `docs/superpowers/specs/`; this lived in the repo only because the engram MCP token was expired when it was written. The PR served its purpose as the review vehicle — the adversarial review that produced rev 2 (Phase 0) happened here.

Phase 0a has since shipped as #1428.

Issue #1409, opened 2026-08-18
Bulk vault upload creates a CRDT room per note (root cause of the 2026-08-18 prod incident)


Rasbandit:
## Problem

A first-sync / bulk upload sends a `crdt_msg` per note, and `ensure_room/2`
(`crdt_channel.ex:1364`) starts a room for every `doc_id` it sees. Uploading a
1.7k-file vault therefore materialises ~2000 CRDT rooms, none of which
correspond to a note anyone has open in an editor.

Rooms should exist for notes open in an editor. A bulk upload does not need
live collaborative state for a note the user is not looking at.

## Evidence (prod, 2026-08-18)

| signal | before | during |
|---|---|---|
| BEAM process count | 757 (flat for hours) | **2744**, held |
| process memory / task | 41 MB | **324 MB** (limit 820 MB) |
| reductions/sec | ~500k | **12.3M** |
| run queue | 0 | 9 (on 0.5 vCPU) |

Each resident room carries a checkpoint obligation. When the tasks were
replaced, ~2000 rooms tried to `unbind -> checkpoint` at once against
`POOL_SIZE=15`, which produced:

```
DBConnection.ConnectionError ... connection not available and request was
dropped from queue after 2001ms
  at=DBConnection.rollback_or_raise/1 <- Engram.Repo.run_with_tenant/2
  <- Engram.Notes.CrdtCheckpoint.do_checkpoint/6
  <- Engram.Notes.CrdtPersistence.unbind/3
```

That is the durability guarantee for ephemeral rooms *failing*: a room is only
safe to drop because it checkpoints to Postgres first.

## Why this is the root cause

Nearly every other symptom is downstream of it — the memory ceiling, the
checkpoint storm, the pool exhaustion, the task replacement, and the killed
client channel (see the linked issues). Fix this and resident rooms collapse to
"notes actually open", at which point the residency cap never binds, per-user
fairness stops mattering, and the checkpoint storm has nothing to storm with.

## Constraints / why this needs design, not a patch

Routing first sync around the CRDT path changes **which path mints note
identity**. That is precisely the machinery behind:

- #1318 (first-sync cross-vault id re-mint, data loss)
- the 39-deletion first-sync incident
- `crdt_create IS the web rename` semantics

So this needs a design pass, not a quick edit, and it spans plugin + backend.

## Acceptance

- A bulk/first sync of N notes creates O(open editors) rooms, not O(N)
- Note identity minting is unchanged or explicitly re-specified
[...]

PR #1449, opened 2026-08-21
feat(crdt): attribute why a genesis body was not seeded roomlessly

Rasbandit:
The #1409 investigation could measure the fallback rate — **70% of a real 319-note vault still opened a room** — but not explain it. `maybe_seed_detached/4` ended in a bare `_ -> false`, so every decline looked identical.

Adds `[:engram, :crdt, :genesis_seed]` with a reason per outcome: `seeded`, `write_declined`, `no_b64`, `b64_not_binary`, `not_markdown`, `room_already_exists`, `not_sync_update`, `frame_decode_failed`, `frame_unsafe`, `apply_failed`.

**This is what found the actual cause.** 318/318 notes reported `seeded` with zero declines, which ruled out the seed path entirely and pointed at the client re-sending the body afterwards — fixed in engram-app/Engram-obsidian#460 (content was doubling 2x per sync, 128x on a sampled note).

## Reasons are per step, and carry nothing else

Three underlying calls (decode, guard, apply) all answer `{:error, _}`, so one arm would lump a malformed frame, a crafted state vector and a NIF apply failure — a client bug, an attack and a server fault. Each step now returns a reason atom unique to that step. Tagged tuples would be the obvious mechanism; credo forbids them in `with` (`Refactor.WithClauses`), so the distinction lives in small step functions instead.

Metadata is the reason atom and **nothing else**. A first revision passed the raw error term as `detail` with a comment saying it must never become a tag — nothing enforced that, and any handler doing `inspect(meta)` would have leaked it. These are exactly the terms this module already warns about at `log_create_failed` ("error_kind/1, never inspect(reason)"): they can embed a wrapped DEK or a provider response body. A test pins `Map.keys(meta) == [:reason]`.

## e2e room probe

`helpers/room_probe.py` counts rooms **allocated** by call path; `test_77` asserts a bound on a 1,000-note bulk sync.

Residency is not usable here: `CRDT_IDLE_EXIT_MS` means rooms a regression opens have drained before the test ends, so a full room-per-note regression passes a `:global` sample taken afterwards — the burst comes and goes between two samples.

Slot count and the atom→slot mapping derive from `SOURCES`. An earlier revision repeated the number in three places; adding a fifth source would have silently under-counted, and `:counters.add` past the end raises *inside* the telemetry handler, which detaches it — the probe would then report zeros, and zeros read as "no rooms allocated".

## Known gap

`test_77` writes uniformly tiny notes (~40 bytes) and measured **0 rooms**, while a real 319-note vault measured **227** on the same code. The bound does not yet cover whatever the real vault trips; it needs a size-mixed fixture before it can be trusted as a regression fence.

## Verification

- 91 tests in the touched files, 0 failures (5 new)
- `mix format`, `mix credo --strict` clean, compiles with `--warnings-as-errors`
- Probe verified against a live node: all four buckets route correctly, including an untagged emit landing in `unknown`

Refs #1409

### DIFF
--- a/e2e/helpers/room_probe.py
+++ b/e2e/helpers/room_probe.py
@@ -42,20 +42,33 @@ from helpers.backend_rpc import backend_rpc
 # RoomStarts — all three are read positionally.
 SOURCES = ("handshake", "edit", "create_batch", "unknown")
 
+# SOURCES is the ONE definition of both the slot count and the atom->slot
+# mapping. An earlier revision hard-coded `:counters.new(4, [])` and `1..4`
+# alongside it — three places encoding the same number, so adding a fifth
+# source would have silently under-counted (and `:counters.add` past the end
+# raises INSIDE the telemetry handler, which detaches it: the probe would then
+# report zeros, and zeros read as "no rooms allocated"). Derived, not repeated.
+_N = len(SOURCES)
+_LAST = _N  # the catch-all slot; SOURCES[-1] must stay "unknown"
+
+_CASE_ARMS = "; ".join(
+    f":{name} -> {i}" for i, name in enumerate(SOURCES[:-1], start=1)
+)
+
 _ARM_EXPR = (
-    ":persistent_term.put(:e2e_room_starts, :counters.new(4, [])); "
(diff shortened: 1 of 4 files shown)


Rasbandit:

PR #1453, opened 2026-08-23
fix(links): chunk note_links inserts under the bind-parameter cap

Rasbandit:
## What

`Engram.Links.replace_links/4` built **one** `insert_all` for every edge in a note. A `NoteLink` row binds up to 17 parameters (13 base + the optional `alias`/`anchor` envelopes), so a note with more than ~3800 wikilinks overflows the wire protocol's 65535-parameter limit.

Postgrex answers that overflow by **killing the connection**, not by returning a query error:

```
[error] Postgrex.Protocol (#PID<0.503.0>) disconnected:
  ** (Postgrex.QueryError) postgresql protocol can not handle 145792 parameters,
     the maximum is 65535
```

So a single oversized note costs a pooled connection every time its extraction runs, and the job retries until it discards.

## How it was found

Watching a real vault sync into a dev backend. 145792 params = **8576 links in one note**, against `pool_size: 10`, with 41 `ExtractNoteLinks` jobs already in `discarded`.

## Fix

Chunk at 2000 rows — a deliberate half-margin under the 3855 the cap allows, so adding an optional column cannot silently re-arm this. The `rows != []` guard goes away because `chunk_every([], n)` is `[]`.

The delete+insert stays inside the existing advisory-locked transaction, so replacement is still atomic and last-writer-wins across chunks.

## Tests

- 5000 edges × 17 binds = 85000 params. Pre-fix this **kills the connection** rather than failing an assertion, so the regression signal is a connection error — called out in the test comment. Verified failing on the unfixed code:
  ```
  ** (Postgrex.QueryError) postgresql protocol can not handle 85000 parameters
  ```
- a narrower rewrite still clears every row from a wider one (chunking must not turn delete+insert into a partial update)

`mix test test/engram/links_test.exs` 36 pass / 0 fail. `format` + `credo --strict` + `dialyzer` clean.

## Note on `.dialyzer_ignore.exs`

Two pre-existing, intentional entries for this file are pinned **by line number**, and the added lines shifted them (dialyxir line-matches — that file's own comment warns about exactly this). Re-anchored 355→371 and 711→727. No new suppressions.

### DIFF
--- a/.dialyzer_ignore.exs
+++ b/.dialyzer_ignore.exs
@@ -23,7 +23,7 @@
   # rather than the literal `200` dialyzer infers from the current
   # `@backlinks_limit` value — the spec documents the contract callers (and
   # tests) can rely on, not today's specific cap. Same pattern as above.
-  {"lib/engram/links.ex", :contract_supertype, 711},
+  {"lib/engram/links.ex", :contract_supertype, 727},
 
   # `EmbedNote.backfill_priority/0` is intentionally specced as `pos_integer()`
   # rather than the literal `9` dialyzer infers from `@backfill_priority` — the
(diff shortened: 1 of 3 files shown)


Rasbandit:

PR #1454, opened 2026-08-23
test(e2e): fence first-sync lineage count against the double-write

Rasbandit:
**Stacked on #1449** (needs its `genesis_seed` telemetry for `seed_probe`). Do not merge before it.

Pairs with plugin PR engram-app/Engram-obsidian#464, which carries the fix these tests fence.

## The defect

A real vault synced into a fresh server vault and **every note was stored twice** — server bytes exactly 2x the file on disk. Disk was never wrong. 224 of 316 notes held two Yjs clients, each holding a full copy.

## Why it stayed invisible

Three things, and each one dictated how these tests are shaped:

- **Opening a note healed it.** `Visit Notes Log (Revere Health).md` went 67,630 → 33,833 bytes the moment it was viewed in the web app. Anything a human inspects has already converged.
- **The copies don't always concatenate.** Where the two insertions interleaved, the text was still 2x but not a clean `X+X` — a "first half == second half" check reported 17 of 60 doubled when the true answer was 224 of 316.
- **Every summary signal was correct.** Note count, room count and seed outcomes all matched expectations. Only the doc *structure* was wrong.

So: assert on **Yjs client count**, read from the **server**, never from a client that could have healed it.

## Tests

| | scenario | result |
|---|---|---|
| `test_97` | clean first sync, device has never seen the notes | passes before and after — the defect isn't here |
| `test_98` | re-sync one local vault into a FRESH server vault (the reporter's actual workflow) | passes before and after |
| `test_100` | frontmatter that does not round-trip byte-for-byte | **7/11 doubled + 7 edit rooms before the fix, 0 and 0 after** |

`test_100` is the reproduction. Verified red against the unfixed plugin, green with it.

## `seed_probe` (new)

Exists because of a specific trap. `test_97` measured **zero rooms** — which is the number reported both when the roomless path works perfectly *and* when the notes never touched the CRDT create path at all. Opposite conclusions from an identical reading; only a seed count separates them.

## Three details worth review attention

- **`vaults_cap` override.** Free allows ONE vault, so `test_98`'s second registration 402s with an empty body — reads as a broken call rather than a plan decision.
- **`test_98` restores `settings.vaultId` in a `finally`.** `obsidian_a` is session-scoped; a plugin left pointing at a throwaway vault silently redirects every later test in the worker. That surfaced here as `test_100` reporting "landed 0/11" with no hint why.
- **`CONVERGE_BOUND_S = 90`, deliberately under `pytest.ini`'s 180s.** Set equal to it, the loop can never reach its own assertion — pytest-timeout kills the test first and reports a generic timeout instead of "converged only N/M".

Refs #1409

### DIFF
--- a/docs/context/crdt-frontmatter-reseed-second-lineage.md
+++ b/docs/context/crdt-frontmatter-reseed-second-lineage.md
@@ -1,0 +1,123 @@
+# Context Doc: identical frontmatter re-seed mints a second lineage
+
+_Last verified: 2026-08-23_
+
+## Status
+Root-caused and fixed. Plugin PR engram-app/Engram-obsidian#464, e2e fence Engram#1454.
+
+## Symptom, as reported
+"We are duplicating the content of each note dozens of times." Then, after a
+clean re-sync: server bytes exactly 2x the file on disk, on 224 of 316 notes.
+Disk was never wrong.
+
+## Why it resists diagnosis
+
+Four traps, in the order they burned time:
+
+1. **Opening a note heals it.** `Visit Notes Log (Revere Health).md` read 67,630
+   bytes at 09:37 and 33,833 at 09:44, `version` 3 -> 6, because it was viewed in
+   the web app at 09:42. Anything a human inspects has already converged, so
+   "I don't see any duplication" and "the server holds it twice" are both true.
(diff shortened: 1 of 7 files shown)


Rasbandit:

PR #452, opened 2026-06-05
feat(billing): inline Paddle checkout for onboarding

Rasbandit:
## Summary

Onboarding flow felt clunky post-#440: Paddle modal → full-card white overlay → navigate. Paddle's recommended pattern for embedded payment in a wizard step is **inline checkout**. Switch onboarding to inline; keep settings on overlay (on-demand action wants a modal).

## Before / after

```
BEFORE                                    AFTER
┌──────────────────────────┐              ┌──────────────────────────┐
│  Choose Your Plan        │              │  Choose Your Plan        │
│  ┌──────┐ ┌──────┐       │   click      │  ← Choose a different    │
│  │Start.│ │ Pro  │       │  ───────►    │  ┌────────────────────┐  │
│  └──────┘ └──────┘       │              │  │ email              │  │
│                          │              │  │ card               │  │
│         ▼ click          │              │  │ [   Pay $7      ]  │  │
│  ┌─────PADDLE MODAL──    │              │  └────────────────────┘  │
│  │ ON TOP OF EVERYTHING  │              │   (Paddle renders        │
│  └───────────────────    │              │    in the SAME panel)    │
│                          │              │                          │
│  ▼ pay, Paddle closes    │              │  ▼ pay, Paddle shows     │
│  ┌──FULL-CARD OVERLAY──  │              │  │ Payment successful   │
│  │ STEP 1 done           │              │  │ (Paddle's own UI)    │
│  │ STEP 2 in progress... │              │  ▼ push event arrives   │
│  │ STEP 3 pending        │              │  navigates to next step │
│  └────WHITE TAKEOVER──── │              │                          │
└──────────────────────────┘              └──────────────────────────┘
```

## State machine

| State | Inline mode | Overlay mode (settings) |
|---|---|---|
| `idle` | Plan cards visible | Plan cards visible |
| `checkout` | `.paddle-checkout` mount div; Paddle drives content | Paddle modal open over the page |
| `slow` | `.paddle-checkout` replaced by small recovery banner | Banner shown |

Transitions:
- Start trial → `checkout`
- `CHECKOUT_PAYMENT_FAILED` / `ERROR` → `idle` (plan picker returns)
- `CHECKOUT_COMPLETED` → start 15s cooldown timer (Paddle's success screen stays visible)
- `subscription_activated` push → close Paddle, fire `onActivated`, navigate
- 15s no push → `slow` (close Paddle, show recovery banner)

## Why this fixes "white takeover"

[...]

PR #1455, opened 2026-08-23
fix(crdt): first-sync double-write — telemetry, e2e lineage fence, note_links cap

Rasbandit:
Supersedes #1449, #1453 and #1454 (all closed in favour of this one).

Pairs with plugin PR engram-app/Engram-obsidian#464, which carries the actual client fix. This PR is the backend half: the telemetry that made the bug attributable, the e2e fence that reproduces it, and one independent server bug found along the way.

## The defect

A real vault synced into a fresh server vault and **every note was stored twice** — server bytes exactly 2x the file on disk. Disk was never wrong. 224 of 316 notes held two Yjs clients, each with a full copy.

Root cause is client-side (#464): `applyFrontmatterInto` replaced the frontmatter ORDER array unconditionally, so re-seeding identical content still minted ops — which makes the writing device a *new client* on the doc, i.e. a second lineage. It fires on a first sync because the engine's own write-back reformats YAML, Obsidian reports that as a real edit, and `handleModify` deliberately skips its echo guard for CRDT notes.

## Why it stayed invisible

Three properties, each of which shaped the tests:

- **Opening a note healed it.** One file went 67,630 → 33,833 bytes the moment it was viewed in the web app. Anything a human inspects has already converged.
- **The copies don't always concatenate.** Where they interleaved, text was still 2x but not a clean `X+X` — a "first half == second half" check reported 17 of 60 doubled when the truth was 224 of 316.
- **Every summary signal read healthy.** Note count, room count and seed outcomes all matched. Only the doc *structure* was wrong.

So the assertion is structural — count Yjs clients, read from the server, never from a client that could have healed it.

## Contents

| commit | what |
|---|---|
| `f3714f1a` | test_77 bounds room allocation on a bulk first sync |
| `b659479a` | `genesis_seed` telemetry — why a body was not seeded roomlessly |
| `b285d0d8` | e2e lineage fence: `test_97`/`98`/`100` + `lineage_probe`/`seed_probe` |
| `fa1fa669` | context doc, including the failed approaches |
| `5ce1ce89` | **independent bug:** `note_links` insert exceeded Postgres' 65535 bind-parameter cap |

`test_100` is the reproduction: **7 of 11 notes doubled with 7 edit-class rooms** against the unfixed plugin, **0 and 0** with it.

### The independent one (`5ce1ce89`)

Unrelated to the doubling, found while reading logs during the investigation. `replace_links/4` built one `insert_all` for every edge; a note with ~8,576 wikilinks asked for 145,792 parameters. Postgrex answers that by **killing the connection**, not by returning an error — so one oversized note cost a pooled connection (of 10) on every extraction, and 41 jobs had already discarded. Now chunked at 2000 rows. Happy to split this back out if you'd rather review it alone.

## Two traps worth review attention

- **`seed_probe` exists because zero is ambiguous.** `test_97` measured zero rooms — the number reported both when the roomless path works perfectly *and* when the notes never touched the CRDT create path at all. Only a seed count separates them.
- **`test_98` restores `settings.vaultId` in a `finally`.** `obsidian_a` is session-scoped; a plugin left pointing at a throwaway vault silently redirects every later test in the worker. That surfaced as `test_100` reporting "landed 0/11" with no hint why.
- **`CONVERGE_BOUND_S = 90`, under `pytest.ini`'s 180s.** Equal to it, the loop can never reach its own assertion — pytest-timeout kills the test first and reports a generic timeout.

## Verification

`format` + `credo --strict` + `dialyzer` clean. `links_test` + `crdt_channel_test`: 122 pass. Full local CRDT e2e down to the 4 documented headless-Chromium `Locator.click` failures that fail on this box regardless of the change.

Refs #1409

### DIFF
--- a/.dialyzer_ignore.exs
+++ b/.dialyzer_ignore.exs
@@ -23,7 +23,7 @@
   # rather than the literal `200` dialyzer infers from the current
   # `@backlinks_limit` value — the spec documents the contract callers (and
   # tests) can rely on, not today's specific cap. Same pattern as above.
-  {"lib/engram/links.ex", :contract_supertype, 711},
+  {"lib/engram/links.ex", :contract_supertype, 727},
 
   # `EmbedNote.backfill_priority/0` is intentionally specced as `pos_integer()`
   # rather than the literal `9` dialyzer infers from `@backfill_priority` — the
@@ -37,5 +37,5 @@
   # `Repo.one/2` types as `term()`, so dialyzer widens the `+` to `number()`
   # and flags `float()` as missing from the `non_neg_integer()` spec. The
   # spec states the real contract.
-  {"lib/engram/links.ex", :missing_range, 355}
+  {"lib/engram/links.ex", :missing_range, 371}
 ]
--- a/docs/context/crdt-frontmatter-reseed-second-lineage.md
+++ b/docs/context/crdt-frontmatter-reseed-second-lineage.md
@@ -1,0 +1,123 @@
+# Context Doc: identical frontmatter re-seed mints a second lineage
+
+_Last verified: 2026-08-23_
+
+## Status
+Root-caused and fixed. Plugin PR engram-app/Engram-obsidian#464, e2e fence Engram#1454.
+
+## Symptom, as reported
+"We are duplicating the content of each note dozens of times." Then, after a
+clean re-sync: server bytes exactly 2x the file on disk, on 224 of 316 notes.
+Disk was never wrong.
+
+## Why it resists diagnosis
+
+Four traps, in the order they burned time:
+
+1. **Opening a note heals it.** `Visit Notes Log (Revere Health).md` read 67,630
+   bytes at 09:37 and 33,833 at 09:44, `version` 3 -> 6, because it was viewed in
+   the web app at 09:42. Anything a human inspects has already converged, so
+   "I don't see any duplication" and "the server holds it twice" are both true.
+2. **The two copies do not always concatenate.** Where the insertions
+   interleaved, text was still 2x but not a clean `X+X`. A
+   "first half == second half" detector reported 17 of 60 doubled when the real
+   answer was 224 of 316. Count Yjs clients, not bytes.
+3. **`plain == crdt` proves nothing.** The `content` column is derived FROM the
+   doc, so a server-side union produces that equality just as a client that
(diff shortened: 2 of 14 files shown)


Rasbandit:
## CI triage

Three findings from the red run, two of them mine and now fixed in `33482b74`.

### 1. I broke `test_32_vault_api_key_isolation` — fixed

The `vaults_cap` override went into `TEST_USER_OVERRIDES`, which every `grant_test_plan` caller applies. `test_32` asserts the Free plan **blocks** a second vault, so its expected 402 became a 201. Same leak class the comments in that dict already warn about for `attachments_enabled`. Now opt-in via `grant_vault_headroom/1`, called by test_98 alone.

### 2. `test_77`'s room bound asserted a fix that doesn't exist — fixed

My comment said "MEASURED 0 locally" and that was true. CI measured **544 handshake rooms** for the same 1,000 notes. Those come from *enrolment*, not cold `crdt_msg` sends — which is #1409's **open** half. A total bound of 8 claimed the whole issue was closed.

Split rather than loosened:
- `ROOM_ALLOC_BOUND` — `edit + create_batch + unknown <= 8`, the paths #1424 + plugin #452 actually fixed. Still a small constant, so a per-note regression fails by orders of magnitude.
- `HANDSHAKE_ROOM_RATCHET` — `handshake <= 700` against the 544 baseline. Fails only when enrolment gets *worse*; meant to be tightened as #1409 lands.

### 3. The lineage fences are red on purpose until the plugin fix merges

`test_97` reported `assert 19 == 0` — **19 notes doubled in CI**. That is the fence working: e2e runs plugin **main**, which still has the bug. `test_97`/`98`/`100` go green when engram-app/Engram-obsidian#464 lands.

Standard pairing deadlock. **Merge #464 first, then this.** Both are non-gating (`e2e-*` is report-only per #1076), so neither blocks.

Worth noting the fence is *more* sensitive in CI than locally — the same tests passed here against plugin main and caught 19 doubled notes there.

### Not mine

- `e2e-crdt` cold-send failures: passed on the re-run after rebase, and `test_cold_send_over_fanout_opens_no_room` fails on main too (run 32622644657). Reproducibly green locally against **both** plugin builds.
- `test_77_bulk_first_sync` timing: the documented load-sensitive assert in `e2e-clerk-failure-taxonomy.md`.

PR #1457, opened 2026-08-24
docs(context): correct stale "drain off by default in prod" claim

Rasbandit:
## Summary
- `docs/context/crdt-room-lifetime-and-drain.md` still claimed prod ran with `idle_exit_ms` unset (drain off by default) — that was true when the doc was written (2026-08-15) but PR #1413 (merged 2026-08-18, released in v0.19.0, live in prod since 2026-08-23 via engram-infra#1041) flipped the default to ON (300s) in every environment.
- Added a dated update note at the top pointing to the correction, and confirmed the fix is actually live: prod process count has held flat for the following day, and `room_drain_total` shows real drain activity (~73 idle-timeout + ~35 LRU-evicted per day against ~119/day room opens).

## Test plan
- [x] Docs-only change, `mix format` clean (pre-push fast path)


Rasbandit:

PR #1459, opened 2026-08-24
fix(e2e): scope the lineage assertion to each test's own notes

Rasbandit:
## Summary

`test_97`, `test_98`, and `test_100` have failed on **every CI run since they merged** (#1455) — never once green — and `main` has been red on `e2e-clerk` since.

They are not catching a live product regression. `read_lineages` sampled **every note in the vault**, but:

- `sync_vault_id` and `vault_a` are `scope="session"` — one vault is shared by the whole suite and its notes persist across tests (test_77's own cleanup docstring says so).
- **34 e2e tests drive a second device**, and a note legitimately edited by two devices holds **two Yjs clients**.

So the unscoped count has a permanent non-zero floor no product fix can reach. The evidence:

- Counts did not budge across two different plugin builds: 12/135 then 12/146.
- `test_98` contributed **11 fresh notes of its own and 0 new multi-client ones** — and was still red purely on accumulated notes from other tests.

## What changed

- `read_lineages(vault_id, path_prefix="")` filters to the caller's prefix. Filtering happens in Elixir **after decrypt** — `path` is encrypted at rest, so there is no column to `LIKE` against.
- All four call sites pass the prefix each test already computed.
- Kept the per-note path print, so a future failure names the offending notes instead of only a count.
- Added `notes >= expected` vacuity guards so a broken filter cannot make these pass on an empty sample.

This can only make the tests **more** sensitive to the fixtures they control, never less.

## Test plan
- [x] `ruff check` + `ruff format` clean on all four files
- [x] Probe expression renders valid Elixir for both scoped and unscoped cases
- [ ] `e2e-clerk` green on this branch — the actual proof

## Note
If `test_100` still fails once scoped, a real frontmatter-echo bug remains and this PR has simply removed the noise hiding it. The per-path output now in the run log will say which notes.

### DIFF
--- a/e2e/helpers/lineage_probe.py
+++ b/e2e/helpers/lineage_probe.py
@@ -35,7 +35,7 @@ from helpers.backend_rpc import backend_rpc
 # wrapper (same constraint room_probe.py documents).
 _PROBE = (
     '{{:ok, vb}} = Ecto.UUID.dump("{vault_id}"); '
-    '{{:ok, vr}} = Ecto.Adapters.SQL.query(Engram.Repo, '
+    "{{:ok, vr}} = Ecto.Adapters.SQL.query(Engram.Repo, "
     '"select user_id::text from vaults where id=$1", [vb]); '
     "u = Engram.Accounts.get_user!(hd(hd(vr.rows))); "
     "{{:ok, nr}} = Ecto.Adapters.SQL.query(Engram.Repo, "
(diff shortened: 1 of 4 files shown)


Rasbandit:
## Confirmed by instrumentation — the tests' own fixtures were always clean

I ran the probe unscoped but printing the **path** of every multi-client note. Every single one belongs to a *different* test:

```
5  E2E/RapidEdits-…              7  E2E/RapidEditsStale-…   ← the "worst: 7 clients"
2  E2E/Referrer88b-…             2  E2E/Outside88-…
2  E2E/ApiUpdateRT.md            2  E2E/ModifyTest.md
3  E2E/AppendSync37.md           3  E2E/CreateRace-…/Raced.md
3  E2E/Orchestra-…               3  E2E/MissedDelivery-…/Note.md
2  E2E/HashOnlyLive.md
```

**Zero** notes under `Roundtrip-*` (test_100), `Lineage-*` (test_97), or `Resync-*` (test_98). Every flagged note comes from a multi-device test — rapid edits, create race, append sync, cross-interface orchestration — where two or more Yjs clients is **correct CRDT behaviour**, not corruption.

So there was never a product regression here. These three tests asserted a vault-wide property while controlling only their own ~11-15 notes, in a `scope="session"` vault shared by ~110 tests.

## Result with the fix

```
110 passed, 1 skipped in 322.63s
  PASSED tests/test_100_frontmatter_echo_single_lineage.py
  PASSED tests/test_97_first_sync_single_lineage.py
  PASSED tests/test_98_resync_to_fresh_vault_single_lineage.py
```

`e2e-clerk` ✅ and `e2e-crdt` ✅ on this branch. This unblocks `main`, which has been red on `e2e-clerk` since #1455 merged.

PR #1461, opened 2026-08-24
docs(e2e): record the measured #1409 room range instead of tightening the ratchet

Rasbandit:
## Summary

`HANDSHAKE_ROOM_RATCHET` was a **700** placeholder set against a 544/1000 baseline — loose enough to hide a large enrolment regression, and the comment on it explicitly said to tighten it as #1409 work landed.

Measured the real number in CI (run `32774859300`, with plugin #466 merged):

```
test_77 rooms allocated for 1000 notes: 335 rooms
  (handshake=335, edit=0, create_batch=0, unknown=0)
```

**335/1000**, down from 544 — a 38% reduction from plugin Engram-obsidian#466, which gated the create-ack self-heal's `reset+enroll` on `isLiveBound`. The `crdt_msg` half stays fixed at zero.

Ratchet set to **400** = measurement + headroom for run-to-run variance.

## #1409 is not closed

335 is still roughly **one room per three imported notes** — O(N), when the acceptance criterion is O(open editors). The constant now records that explicitly, so the remaining work is visible rather than buried under a bound nothing can trip.

## Note on measuring

`test_77` prints its room split **only on failure**, so a passing run hides the number the ratchet is supposed to track. To re-measure, set the ratchet to `0` on a throwaway branch and read it out of the assertion message — that's how this number was obtained.

## Test plan
- [x] Measured in CI, number recorded in the constant's comment with the run id
- [x] `ruff check` + `ruff format` clean
- [ ] `e2e-clerk` green with the ratchet at 400 (335 < 400, so it should pass)

### DIFF
--- a/e2e/tests/test_77_bulk_first_sync.py
+++ b/e2e/tests/test_77_bulk_first_sync.py
@@ -50,7 +50,7 @@ ROOM_ALLOC_BOUND = 8
 # that makes enrolment worse still fails, while the known O(N) baseline does not
 # spend every run red. TIGHTEN THIS as #1409's enrolment work lands; a run that
 # comes in far under it means the bound is stale, not that nothing happened.
-HANDSHAKE_ROOM_RATCHET = 700
+HANDSHAKE_ROOM_RATCHET = 0  # MEASUREMENT ONLY — forces the assert to print the real count
 
 SET_BLOCKED = "app.plugins.plugins['engram-vault-sync'].syncEngine.setSyncBlocked({})"
 
--- a/e2e/tests/test_77_bulk_first_sync.py
+++ b/e2e/tests/test_77_bulk_first_sync.py
@@ -49,15 +49,27 @@ ROOM_ALLOC_BOUND = 8
 # target: it exists so a change that makes enrolment WORSE fails, while the
 # known-open O(N) baseline does not spend every run red.
 #
-# MEASURED 335/1000 on 2026-08-24 (CI, run 32774859300), down from the 544/1000
-# baseline of 2026-08-23 after plugin #466 gated the create-ack self-heal's
-# reset+enroll on isLiveBound. The bound below is that measurement plus headroom
-# for run-to-run variance.
+# THIS NUMBER IS HIGHLY VARIABLE — do not read it as a property of the code.
+# Measured on 2026-08-24 against the SAME commit, twice:
+#     run 32774859300 -> 335 rooms / 1000 notes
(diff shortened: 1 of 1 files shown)


Rasbandit:
## Correction: the number is not stable, and the PR body oversold it

A second CI run of the **same commit** measured **57** rooms/1000 notes where the first measured **335**:

| run | rooms / 1000 notes |
|---|---|
| 32774859300 | 335 |
| 32778220678 | **57** |

Same code. So this metric tracks run conditions (reconnects, load, socket churn) at least as much as it tracks enrolment logic, and my original framing — "335 + headroom" — implied a precision that does not exist.

**400 stays**, but the rationale changes: it is a loose fence set above the worst *observed* value, not a snug bound above a measured one. Setting it tightly against either sample buys a flaky job rather than a real fence, while 400 still fails loudly on a return to the O(N)-per-note shape #1409 is about.

The comment in the file now records both measurements and the attribution data (`fireCrdtReHandshake` 20, `applyChange` 6 = 26 `enroll()` calls for 57 rooms — so more than half the rooms never go through `enroll()`), so the next person tightening this has the variance in front of them.

Full analysis and suggested next steps posted to #1409.

Rasbandit:
## I was wrong — reverting my own tightening

This PR originally set the ratchet 700 → 400 off a **single** 335-room sample. The very next CI run measured **583** and failed the job on a change that had regressed nothing. That is exactly the "a bound near the middle of a noisy range buys a flaky job, not a fence" failure I had written into the comment one commit earlier, and then walked into anyway.

Observed across runs of essentially the same code (post plugin #466):

```
57, 81, 335, 583   rooms / 1000 notes      (baseline 544 pre-#466)
```

A **10x spread**. The count tracks reconnects and socket churn as much as enrolment logic, so no single sample supports a bound.

**700 stays.** It sits above every observed value while still failing loudly on a true return to one-room-per-note (~1000).

## What this PR is now worth merging for

Not a smaller number — the recorded evidence:

- the measured range, so nobody re-tightens off one sample (including me)
- attribution: these are ordinary `enroll()` calls dominated by `fireCrdtReHandshake`, which is a **delivery path for durably queued edits and must NOT be gated** to shrink this number (see #1409 and Engram-obsidian#469)
- the falsified reconnect re-advertise theory, so it does not get re-derived
- the measurement trap: `read_room_starts()` counts rooms server-wide across all three Obsidian instances, while a `cdp_a`-based client probe sees only device A — the two are not comparable without accounting for it

Rasbandit:

PR #1462, opened 2026-08-25
test(e2e): bound PEAK CONCURRENT rooms during a bulk import

Rasbandit:
## The gap this closes

#1409's incident was a **memory** event: a 1.7k-file import left ~2000 rooms **resident** and took prod from 757 to 2744 processes. `test_77` only bounded **allocation** — and since the idle drain shipped, allocation no longer implies memory.

Worse, allocation turned out to be unusable as a fence. Measured across runs of essentially the same code:

```
57, 81, 335, 583   rooms / 1000 notes
```

A 10x spread (it tracks reconnects and socket churn). I tried tightening the ratchet to 400 off one sample; the next run hit 583 and went red on a change that regressed nothing. See #1461.

## Residency is the number that matters, and it's stable

```
PEAK RESIDENT rooms during import: 4     (1000 notes)
```

Measured **twice** — once at CI's 5s idle drain, once with the drain set to prod's **300s**. It did not move.

That is not luck. A note room exits on **`auto_exit`, when its last observer leaves** — not on the idle timer. The timer is a backstop for the per-vault *index* room, which is observed for a whole session. Enrolment being gated to live-bound notes (#1424, plugin #466) is what lets observers leave promptly, and *that* — not the drain — is what actually resolved the 2026-08-18 shape.

I nearly reported the first `4` as proof the drain obsoleted #1409. It wasn't: CI drains 60x faster than prod, so that number had to be re-measured at the prod window before it meant anything.

## What's added

- `e2e/helpers/residency_probe.py` — counts live children of `CrdtDocSupervisor`. Sampled on **every pass** of the convergence loop, never once at the end: `room_probe`'s docstring is right that a single trailing sample cannot see a burst come and go.
- `PEAK_RESIDENT_ROOM_BOUND = 32` — ~8x the observed peak. A regression to one resident room per note misses by two orders of magnitude; normal churn stays far under.

The allocation ratchet stays at 700 and is unchanged here.

## Test plan
- [x] Measured 4/1000 at both the CI and prod drain windows
- [x] `ruff check` + `ruff format` clean
- [ ] `e2e-clerk` green with the new assertion

### DIFF
--- a/e2e/helpers/residency_probe.py
+++ b/e2e/helpers/residency_probe.py
@@ -1,0 +1,41 @@
+"""Sample CRDT room RESIDENCY — how many rooms are alive at one instant.
+
+The companion to `room_probe`, which counts rooms ALLOCATED. Those two answer
+different questions, and since the idle drain shipped they can differ by orders
+of magnitude:
+
+  * ALLOCATION is the #1409 acceptance claim ("an import must not open a room
+    per note"). It is cumulative and cannot be missed by sampling.
+  * RESIDENCY is what actually costs memory. `CrdtCheckpointTimer`'s idle drain
+    defaults ON (300s) in every environment as of 2026-08-18, and `ci/compose.yml`
+    shortens it further, so an allocated room is released long before a test ends.
+
+`room_probe`'s docstring rightly warns that a SINGLE residency sample taken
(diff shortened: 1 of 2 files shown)


Rasbandit:

PR #1464, opened 2026-08-25
Update README to remove redundant statement

Rasbandit:
Removed redundant line about notes being AI's memory.

## Test plan

Closes #


Rasbandit:
Superseded by #1474, merged as f8a20818 — identical two-line deletion.

Re-cut rather than merged because CI here is `push`-triggered on conventional branch prefixes only (`feat/**`, `fix/**`, `docs/**`, …) with no `pull_request` trigger, so `Rasbandit-patch-1` never ran `lint`, `ci`, or `unit-tests`. Those are the three contexts `protect-main` requires, which left this at `mergeStateStatus: BLOCKED` with nothing able to clear it. Renaming the branch would have closed this PR anyway, so the change moved to `docs/readme-remove-redundant-line`.

PR #1476, opened 2026-08-26
fix(embed): bulk enqueue must skip notes that already have a pending job

Rasbandit:
## The hole

`EmbedNote`'s `unique:` option is enforced by Oban's `insert/3` advisory lock. All three bulk callers use `Oban.insert_all/2`, and **the basic engine ignores `unique` there** — bulk unique is an Oban Pro (Smart Engine) feature, confirmed in the 2.23.1 docs.

On its own that's a duplicate. Combined with `ReconcileEmbeddings` it's a **ratchet**. That worker selects notes on `embed_hash != content_hash` plus a cooldown stamp, and never asks whether a job is already pending — its own comment said:

```elixir
# Per-vault fairness isn't needed: EmbedNote is uniq-deduped
```

which is exactly the guarantee its own insert path disables. A note whose job was stuck therefore collected one more job every backoff window, forever.

Measured in dev on 2026-08-25: **61,536 pending jobs for 4,266 notes**, one note queued **184 times across five days**. Five days ÷ 30-minute backoff ≈ 240 windows, which matches. That backlog is what kept the queue from draining and drove RSS to 1 GB.

## The fix

`EmbedNote.reject_already_queued/1`, wired into all three bulk sites: batch upsert, `ReconcileEmbeddings`, `ReindexKeyword`.

## Two design decisions, both pinned by tests

**`executing` does not suppress.** A running job may already have read the previous content, and its `embed_hash` stamp is a no-op under the existing optimistic lock — so suppressing on `executing` would strand an edit made mid-embed until the next reconcile sweep (~45 min). The pile-up still caps at 2: the follow-up job is `available`, which suppresses every later tick.

**Lookups chunk at 50 ids.** Postgres estimates `args->>'note_id' = ANY($1)` badly for large arrays. Measured on a 40k-row `oban_jobs`:

| ids per query | plan | time |
|---|---|---|
| 25 | Bitmap Heap Scan on `oban_jobs_embed_note_note_id_index` | 0.4 ms |
| 50 | Bitmap Heap Scan (same index) | 2.8 ms |
| 100 | **Seq Scan** | 23.6 ms |
| 250 | **Seq Scan** | 31.5 ms |
| 500 | **Seq Scan** | 27.8 ms |

The seq scan is O(backlog) — it degrades precisely when this guard is load-bearing. Ten small index probes beat one scan that gets worse over time. Still far cheaper than the per-note `existing_burst_start/1` SELECT that `clamp: false` exists to avoid: 500 notes cost 10 queries, not 500.

## Verification

10 new tests, all mutation-checked. Chunk-size mutations to 500, 25 and 1 each fail the query-count assertion — it's exact rather than an upper bound, so both the per-id regression and an over-large chunk fail. State-list and pass-through mutations fail the behavioural tests.

- `822 tests, 0 failures` across `test/engram/workers/`, `test/engram/notes/`, `notes_test.exs`
- format / `credo --strict` / `compile --warnings-as-errors` / dialyzer / sobelow all clean

Two housekeeping notes, neither a new suppression:
- The dialyzer ignore entry for `embed_note.ex` moved 411 → 486; it is line-anchored and the new function shifted it.
- Three `.sobelow-skips` entries for `notes.ex` were re-anchored (4647/4982/6632 → 4660/4995/6645). Skips encode `file:LINE,FINGERPRINT`, so the 13-line shift broke them. Regenerating added exactly 3 entries and removed none, confirming no new findings; the stale trio was then deleted, leaving the count at 21.

Refs #1409

### DIFF
--- a/.dialyzer_ignore.exs
+++ b/.dialyzer_ignore.exs
@@ -30,7 +30,7 @@
   # spec documents the contract ("some priority that loses to interactive"),
   # not today's value, so callers and tests don't bake the constant in a second
   # place. Same pattern as `Links.backlinks_limit/0` above.
-  {"lib/engram/workers/embed_note.ex", :contract_supertype, 486},
+  {"lib/engram/workers/embed_note.ex", :contract_supertype, 519},
 
   # `Links.live_basename_count/3` sums two `Repo.one(select: count(...))`
   # results. SQL `count()` is always a non-negative integer at runtime, but
--- a/.dialyzer_ignore.exs
+++ b/.dialyzer_ignore.exs
@@ -30,7 +30,7 @@
   # spec documents the contract ("some priority that loses to interactive"),
   # not today's value, so callers and tests don't bake the constant in a second
   # place. Same pattern as `Links.backlinks_limit/0` above.
-  {"lib/engram/workers/embed_note.ex", :contract_supertype, 411},
+  {"lib/engram/workers/embed_note.ex", :contract_supertype, 486},
 
   # `Links.live_basename_count/3` sums two `Repo.one(select: count(...))`
   # results. SQL `count()` is always a non-negative integer at runtime, but
--- a/.sobelow-skips
+++ b/.sobelow-skips
@@ -6,10 +6,10 @@ Config.CSWH: Cross-Site Websocket Hijacking,lib/engram_web/endpoint.ex:45,6EA19F
 Config.HTTPS: HTTPS Not Enabled,config/prod.exs:0,2B5C077
 Misc.BinToTerm: Unsafe `binary_to_term`,lib/engram/crypto.ex:581,7EA51D6
 Misc.BinToTerm: Unsafe `binary_to_term`,lib/engram/crypto/user_dek_rotation.ex:1411,772DFE4
-Misc.BinToTerm: Unsafe `binary_to_term`,lib/engram/notes.ex:4660,3A896E5
-Misc.BinToTerm: Unsafe `binary_to_term`,lib/engram/notes.ex:4995,482C111
+Misc.BinToTerm: Unsafe `binary_to_term`,lib/engram/notes.ex:4665,1171D7E
+Misc.BinToTerm: Unsafe `binary_to_term`,lib/engram/notes.ex:5000,8A8E13
 SQL.Query: SQL injection,lib/engram/attachments.ex:1049,2991CD5
-SQL.Query: SQL injection,lib/engram/notes.ex:6645,4F747BE
+SQL.Query: SQL injection,lib/engram/notes.ex:6650,199FE8B
 Traversal.FileModule: Directory Traversal in `File.read!`,lib/engram/legal/seeder.ex:84,12C6307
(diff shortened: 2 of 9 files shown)


Rasbandit:
## Self-review pass — one real defect found and fixed

Reviewed the first commit adversarially. Two findings, plus one pre-existing bug worth a separate issue.

### 1. Priority inversion (serious, introduced by this PR) — fixed in 6eee614

`priority_for/1` exists so a live edit jumps an import backlog: a never-embedded note enqueues at priority 9, a re-embed at 0. Its own doc says a 50k-note import would otherwise park every other user's edit behind it.

The guard defeated exactly that. During an import every note holds a pending **priority-9** job; a user edit takes the batch-upsert path at **priority 0**, and the guard suppressed the enqueue — so the edit inherited the pending job's position behind the whole backlog. The behaviour this PR replaced (a duplicate priority-0 job) did *not* have that problem, so this was a genuine regression, not just a missing improvement.

`reject_already_queued/2` now takes the incoming priority and **promotes** a pending job ranked worse than it, in place. One pending job per note is preserved and the edit keeps its position. Oban fetches `priority ASC`, so lowering the number is the promotion.

The UPDATE only fires when something is actually outranked, so the common path is still one query per chunk.

### 2. Reconcile logged a count it no longer enqueues — fixed

`total_count: length(note_ids)` was what the eligibility query matched, not what was queued. Post-guard those diverge, so during a backlog incident — the exact situation this guard exists for — the log would claim a full batch while suppressing all of it. Now reports `total_count` / `eligible_count` / `already_queued_count`.

### A mutation caught a hole in my own tests

Deleting the `priority > ^new_priority` clause from the promotion UPDATE left **every test green**. It isn't redundant: promotion selects by `note_id`, and a note *can* hold two pending jobs via the insert race documented in the moduledoc — without that clause the better-ranked sibling gets demoted. Added a test for the two-pending-jobs case; the mutation now fails.

### Pre-existing, not addressed here

`ReindexKeyword` looks inert for its stated purpose. It re-enqueues a vault's notes to recompute BM25 weights against current `avgdl`, but `EmbedNote.perform/1` short-circuits on `content_hash == embed_hash`:

```elixir
{:ok, %Note{content_hash: hash, embed_hash: hash}}
when hash != nil and is_nil(old_path_hmac_b64) -> :ok
```

Every successfully-embedded note hits that clause, which is precisely the population a re-normalizer targets. Nothing clears `embed_hash` first. It's marked "Scaffold" and pre-launch-manual, so it may never have been exercised. Unrelated to this PR — worth its own issue rather than widening this one.

### Re-verified after the fix

- `826 tests, 0 failures` across `test/engram/workers/`, `test/engram/notes/`, `notes_test.exs` (14 in the new file)
- format / `credo --strict` / `compile --warnings-as-errors` / dialyzer / sobelow all clean
- Line-anchored ignores re-anchored again (dialyzer 486 → 519, three notes.ex sobelow skips). Sobelow regeneration diff was 3 additions, 0 removals — no new findings, no new suppressions.

PR #476, opened 2026-06-06
feat(accounts): immediate cascade lifecycle + export pipeline foundation (Gate 1 / #281)

Rasbandit:
## Summary

**Gate 1** of the Account Lifecycle workstream (engram-app/Engram#281). Internal-only refactor + new schema scaffolding. **No new user-visible surface** in this PR — the export feature + inline billing UI ship in Gates 2 and 3.

### What changes for users (saas)

**Immediate cascade on account delete.** `Accounts.delete_self/2` (self-host) and the Clerk `user.deleted` webhook (saas) now run a full cascade purge in-line (Qdrant → S3 → Postgres → Clerk) instead of soft-deleting and waiting 30 days for the inactivity sweep. Same end state, much sooner. Closes the GDPR right-to-erasure gap from launch readiness.

### Backend additions
- `Engram.Accounts.Lifecycle` — `soft_delete/2` and `hard_delete/2` extracted from `InactivityCleanup`. Idempotent. Ordered failure semantics per the spec §7.2 (sessions disconnect first, Paddle cancel, Qdrant/S3 best-effort, `Repo.delete!` as the commit point, Clerk last).
- `Engram.Accounts.Export` — context with `request/1`, `list/2`, `get/2`, `mint_download_url/2`. Rate limits (`account_exports_lifetime` / `_rate_per_24h`) and size cap (`account_export_max_bytes`) wired through the existing limit-keys resolver.
- `Engram.Accounts.Export.Streamer` — multipart-upload pipeline scaffold (Zstream → S3 multipart) for the worker. **MVP only — ships ciphertext, no decryption**, no .obsidian skip, no 10 GB split. Pulled forward to validate the multipart adapter shape; user-facing emission lands in Gate 3 (Plan Tasks 13/14).
- `Engram.Workers.AccountExport` — Oban worker on the new `:export` queue. Happy-path complete.
- `Engram.Paddle.Client.cancel_subscription/3` — new behaviour callback + HTTP impl, used by `Lifecycle.hard_delete`.
- Storage adapter behaviour gains `selfhost?/0`, `sign_url/2`, and S3 multipart callbacks (`start_multipart`, `upload_part`, `complete_multipart_upload`, `abort_multipart_upload`). InMemory + S3 + Database adapters all implemented.
- `Engram.Crypto.HMAC.hash_user_id/1` — used by new `[:engram, :account, :*]` telemetry.

### Schema
[...]

PR #1485, opened 2026-08-27
feat(oban): split worker execution off web nodes; remove crdt_create_batch

Rasbandit:
## Summary

- **`ENGRAM_NODE_ROLE=web` makes a node enqueue but never execute Oban jobs** (`queues: false`). `embed: 5` (peak ≈ 560 MB) and `indexing: 2` can no longer take schedulers or pool connections from a node holding user sockets. Deploy as two ECS services off one image: worker sets nothing, web sets `ENGRAM_NODE_ROLE=web`.
- **Opt-out, not opt-in**, deliberately. Unset keeps the full queue list, so self-host and single-node deploys need no configuration and behave exactly as today. An opt-in flag would mean one unset variable on the worker leaves nothing draining the queues — silently, since enqueues keep succeeding. A `web` node with no `DNS_CLUSTER_QUERY` warns at boot for the same reason, and because `CheckpointNote`'s live-room guard is a `:global` lookup that needs the cluster to avoid checkpointing an open note.
- **`crdt_create_batch` removed** (−823 lines). The plugin retired it for per-file failure isolation; the web app was the last caller and always sent exactly one entry, so a three-phase `Task.async_stream` pipeline and a per-entry pool bound existed to serve single-item calls. It was also the last site starting a SharedDoc room per create — the single `crdt_create` went roomless in #1409 and the batch leg never followed. The web's create-with-content path moves onto `crdt_create`'s existing optional `b64` genesis frame, the same roomless seed.

### One behaviour change, deliberate

A batch entry replied `status: "ok"` whether or not its seed actually landed, so a duplicate whose genesis was declined looked like success and wrote a silently **empty** note. `crdt_create` reports the real outcome, so `crdtCreateNoteWithContent` now rejects on anything but `"stored"`. The reachable case is duplicating a `.canvas`: the roomless seed refuses non-markdown on purpose, since projecting a markdown body over a canvas would erase the board. That now toasts instead of producing a blank file.

## Test plan

- [x] Full backend suite (`mix test`)
- [x] `crdt_channel_test` + `crdt_channel_origin_test` — 95/95 in isolation
- [x] Frontend `src/api` + `src/crdt` — 366/366
- [x] `mix credo --strict`, `mix format --check-formatted`, `mix compile --warnings-as-errors`
- [x] `bunx tsc --noEmit`, biome clean
- [x] `config/runtime.exs` evaluated under each role: unset → `nil` (config.exs list stands), `worker` → `nil`, `web` → `false`
- [ ] **Reviewer:** the node split is not exercised by CI — `runtime.exs` isn't loaded in tests and the boot warning is log-only, matching the untested `dns_cluster_query` / rate-limiter gates beside it. First real proof is a staging boot with the variable set.

### Known flakes, not introduced here

Two `EngramWeb.CrdtChannelTest` tests fail intermittently under full-suite load on the dev box:

1. `crdt_create with b64 … reports occupied, not absent` — **reproduces identically on clean `origin/main` at the same seed** (control run, 4833 tests). Its own comment documents the cause: it parks a global `:after_row_read` hook that every checkpoint in the suite reaches, including background room-drain and timer ones.
2. `crdt_msg rate limiting — sync handshake frames do NOT consume the edit budget` — seven sequential pushes each on a 3s `assert_reply`. Not yet observed on `main`; accepted as load-sensitive rather than proven, and called out here so a reviewer can push back.

Neither has a mechanism connecting it to this diff: the describe block resets the limiter in `setup` and `on_exit`, and the file is `async: false`.


Rasbandit:
Adversarial review pass applied in `dc89e35b` — six findings, all real. Summary of what changed, for the morning read:

- **Orphan row on a declined seed (the MEDIUM one).** Fixed at both ends: a non-markdown path is refused *before* the create, so duplicating a `.canvas` never makes a row; a server-side surprise deletes the orphan before throwing, gated on `absent` alone (deleting on `occupied` would destroy the concurrent write that filled it).
- **`occupied` mistranslated to a name collision.** The id-mismatch check already owns adopt, so that branch means the id matched and the row is ours — a CAS loss. It was toasting "a note with that name already exists" for a free name.
- **Boot guard on the wrong node.** It keyed on `queues == false` (the web node), but the node that can corrupt data is the *worker* — `CheckpointNote`'s live-room guard is a `:global` lookup, so an unclustered worker checkpoints notes that are open. Now keyed on `ENGRAM_NODE_ROLE` being declared at all; still silent for an unset single-node deploy.
- **`ENGRAM_NODE_ROLE` now a closed set.** The opt-out direction made typos silent (`wbe` = run every queue on a node meant to run none). Case/whitespace forgiven, anything else raises.
- **Comment splice** in `application.ex` — the cookie rationale was documenting the wrong function.
- **Coverage.** Ported the cross-vault re-mint-with-content regression test to `crdt_create`, and strengthened the origin test, whose assertion had become identical to the bare-create test above it.

### Two findings NOT taken as code

**No deprecation window on the removed frame.** A browser tab loaded before the deploy still sends `crdt_create_batch` and now hits the catch-all at `crdt_channel.ex:638`, which replies `bad_frame` — so Duplicate fails with a toast in that tab until reload. Bounded and self-healing, and keeping the handler for a release would mean re-adding the code this PR deletes. Flagging rather than fixing; say the word if you want the window.

**Origin test asserts `occupied`, not `stored`.** Worth knowing this is the *correct* answer rather than a tolerated one: a relocate moves a row that already holds a body, so the seed declines. A `stored` there would mean the genesis frame overwrote the note being renamed.

### Flake status

`crdt_create with b64 … reports occupied, not absent` reproduces on clean `origin/main` at seed 0 (4833 tests) — pre-existing, ordering-sensitive. Paired runs so far:

| seed | branch | main |
|---|---|---|
| 0 | fails | **fails** |
| 1 | fails | clean |

The second flake (`crdt_msg rate limiting … edit budget`) has not been observed on main and is accepted as load-sensitive rather than proven — called out so a reviewer can push back.

Rasbandit:
## Overnight CI triage

Two jobs red on `dc89e35b`. Diagnosis below; **neither is caused by this diff**, and one is a real pre-existing test bug I'd rather you look at than have me patch unsupervised.

### 1. `unit-tests` — `crdt_create with b64 … a write ABORTED by a concurrent one reports occupied, not absent`

Not a load flake. The actual failure:

```
** (FunctionClauseError) no function clause matching in EngramWeb.CrdtChannel.handle_info/2
    # 1
    :release_park
```

The test's bespoke interleave hook parks **inside the channel process**, inside `do_checkpoint`'s `run_with_tenant` transaction. Two consequences:

1. The parked channel is holding a DB connection — and under the sandbox that's the **owner's** connection, shared with the test process via `Sandbox.allow/3` in `join!`. So the test's concurrent `upsert_note/3` queues *behind the park* and cannot commit while it holds.
2. The park therefore always expires at 5s, the hook returns, and the test's later `send(channel_pid, :release_park)` reaches `Phoenix.Channel.Server.handle_info/2`, matches none of `CrdtChannel`'s three clauses, and kills the channel.

Whether the test passes is a coin flip on whether that crash lands before or after `assert_reply`. It reproduces on clean `origin/main` at seed 0, so it predates this PR.

I tried raising the park to 30s — that **deadlocks**, because it exceeds DBConnection's 15s checkout timeout while holding the connection the write needs. So 5s is an upper bound, not a guess.

The real fix is the one the sibling tests already use: `feed_interleave_test.exs` and `checkpoint_interleave_test.exs` take REAL connections via `CheckpointInterleave.checkout_real!()`, precisely because *"the sandbox is one connection in one never-committed transaction, so nothing can commit into the gap and a sandboxed version of this test proves nothing at all."* This test is in the sandboxed file and structurally cannot do what it claims. Moving it to the real-connection harness is the fix — deliberately **not** attempted here, since it's a subtle concurrency test unrelated to this PR. Happy to do it as a follow-up.

### 2. `e2e-crdt` — 4 failed, 16 passed

Three fan-out specs (`test_concurrent_cold_edits_survive_over_fanout`, `test_cold_send_over_fanout_opens_no_room`, `test_fanout_sequential_edits_converge_preserving_prior_state`) plus `test_web_to_obsidian_live`, the last being the documented CDP `Connection refused` browser-half flake.

This diff touches nothing on that path: `CrdtDeliver.fanout_idle/3` and `seed_detached/4` are unmodified, and the plugin never sent `crdt_create_batch` (it retired the frame itself). `e2e-browser` — the arm that actually exercises the changed web path — is green.

Recent `main` runs fail `e2e-clerk`, not `e2e-crdt`, so I'm not claiming pre-existing without evidence. Re-running the failed jobs on the same SHA for a second data point rather than asserting flake.

**Not merged**, per instruction.

Rasbandit:
## Morning summary — one red job, and it is not this diff

After a rerun on the same SHA (`dc89e35b`), **every check is green except `unit-tests`**:

```
SUCCESS  e2e-browser, e2e-clerk, e2e-crdt, e2e harness lint, fingerprint,
         frontend lint+types+unit, headless-protocol, lint, migration gates,
         prebuild-ci-image, prebuild-mix, static checks, storage-database,
         version-check
FAILURE  unit-tests
```

`e2e-crdt` and `e2e-clerk` both failed then passed on an identical SHA — nondeterministic, and `e2e-browser` (the arm covering the changed web path) was green on both passes.

### The one real blocker

`crdt_create with b64 … a write ABORTED by a concurrent one reports occupied, not absent` failed **both** CI runs with the same `FunctionClauseError` on `:release_park`. So on CI it is deterministic, not a coin flip — the runner is slow enough that the park always expires. It reproduces on clean `origin/main` at seed 0.

**I deliberately did not fix it, and want your call.** Three things I established trying:

1. **Raising the park deadlocks.** The park sits inside `do_checkpoint`'s transaction holding the sandbox's *owner* connection — shared with the test process via `Sandbox.allow/3`. At 30s the test's own concurrent write times out at DBConnection's 15s checkout waiting for that connection. 5s is a ceiling, not a guess.
2. **The premise never holds under the sandbox.** I added a `{:park_expired}` handshake so the test reports expiry instead of crashing — it flunked, meaning the concurrent write *never commits during the park*. The test cannot be exercising the abort path it is named for. That it passes locally at all appears to depend on whether the stray-release channel crash lands before or after `assert_reply`.
3. **The correct fix is the harness its siblings already use.** `feed_interleave_test.exs` and `checkpoint_interleave_test.exs` take real connections via `CheckpointInterleave.checkout_real!()` precisely because *"the sandbox is one connection in one never-committed transaction, so nothing can commit into the gap and a sandboxed version of this test proves nothing at all."* Moving this test there means driving a Phoenix channel outside the sandbox, which that harness has never hosted.

I stopped there rather than redesign a concurrency test that guards a data-loss class (#476) unsupervised overnight. Suppressing or excluding it was not on the table. Say the word and I will do the harness move as a follow-up PR — it should not ride in this one either way, since it predates it.

**Not merged.**

Rasbandit:
## CI is green ✅

`afba8a1a` fixes the last red job. `ci: SUCCESS`, merge state `CLEAN`. **Not merged.**

`unit-tests` was deterministically red (2/2 runs) before this and is now green.

### What the fix actually was

Not a timeout bump — the sandboxed test could not test what it claimed. Its hook parks the channel inside `do_checkpoint`'s transaction, and under `ChannelCase` that transaction runs on the sandbox's ONE connection, shared with the test process via `Sandbox.allow/3`. So the competing write queued *behind* the park rather than committing during it, the park always expired, and the release message then reached `Phoenix.Channel.Server.handle_info/2`, matched no clause, and killed the channel. Passing or failing came down to whether that crash landed before or after `assert_reply`.

Moved to `test/engram_web/channels/crdt_seed_interleave_test.exs` on the harness its siblings use, plus the one piece they did not need: `Sandbox.mode(Repo, :auto)`, so the CHANNEL draws its own pooled connection instead of sharing the test's. The competing write now commits on a different connection while the park is held — the actual interleave. `await_parked/2` raises if the park never happens, so unlike its predecessor it cannot pass vacuously.

Reproduced on clean `origin/main` beforehand, so it predated this PR.

### One caveat on this run

The e2e jobs show **SKIPPED**, not green — the fingerprint short-circuited them because `afba8a1a` is test-only. They passed on the previous SHA (`dc89e35b`) after a rerun. Worth a fresh e2e pass before merge if you want them green on this exact commit.

### Also unresolved, noted not fixed

`ObanQueueConfigTest` failed on one local full-suite run, passed on the next, and passes in isolation. Nothing mutates the Oban env and it is `async: true`, so my `async: false` addition cannot be racing it. It has not appeared in CI. Flagging rather than claiming it is fine.

Issue #1502, opened 2026-08-28
Worker hot path re-resolves per-job context: thread it once, cache it across jobs


Rasbandit:
Umbrella for the DB-call pattern behind #1494 and #1501. Filing separately because the individual issues each describe a symptom and the fix is one shape applied in two places.

## The pattern

The worker resolves the same immutable-for-this-job context over and over. Measured in prod 2026-08-28 at ~13 jobs/sec: **~220 Ecto queries/sec, about 17 per job**, of which ~41 q/s is pure re-resolution of `users`, `vaults` and `subscriptions` — the same rows, for one tenant, thousands of times in a few minutes.

## Two distinct problems, two different fixes

**1. Within a single job — thread it, don't cache it.**

`users` runs at **2.1 per job**. A cache would hide that; the right fix is that one job should resolve identity once. Two known sites on the embed path:

- `Engram.Indexing.prepare_index/2` — `user = Engram.Accounts.get_user!(note.user_id)`
- `Engram.Workers.EmbedNote.embed_budget_gate/1` — `user = Accounts.get_user!(user_id)`

Same user, same job, two round trips. Worth walking the whole path for others before deciding the shape (a resolved-context struct passed down, versus each function taking `%User{}` instead of a `user_id`).

**2. Across jobs — cache it.**

A bulk import is thousands of jobs for ONE user and ONE vault. `Accounts.get_user!/1` is a bare `Repo.get!`. There is no user or vault cache.

The infrastructure for this is already built and proven — `Engram.Cache.NodeLocalEts` gives a named public ETS table, TTL, `Engram.Cluster.CacheSync` PubSub eviction, and Postgres `LISTEN/NOTIFY`. Eight caches use it today:

```
billing/override_cache      keyword_index/stats_cache    usage/daily_cap/cache
billing/entitlement_cache   keyword_index/stem_cache     usage_meters/activity_cache
onboarding/terms_cache      onboarding/gate_cache
```

None covers `users` or `vaults`. The gap is not tooling; it is that the pattern was applied to billing and onboarding and never to identity.

## Same shape, non-DB: #1501

`Qdrant.ensure_collection/2` runs per note and costs two round trips (a `PUT` that 409s, then `collection_info` to verify shape) against a collection that cannot change per note. Identical reasoning — per-job work that is really per-process — and a TTL cache keyed on `{url, collection, dims}` is the same fix. Keying on URL also keeps the async Bypass suites isolated.

[...]

Issue #1501, opened 2026-08-28
Every embed makes 4 Qdrant round trips where 1 is the work (ensure_collection + collection_info per note)


Rasbandit:
Measured in prod 2026-08-28 across a 1.4k-file sync.

## Numbers

Cumulative `engram_prom_ex_qdrant_request_total`, all `status=ok`:

```
upsert             853    ← the actual write
ensure_collection  853    ← 1:1 with upsert
collection_info    853    ← 1:1 with upsert
delete             873
```

The 1:1 ratio is the finding. Every single `EmbedNote` execution calls `ensure_collection` and `collection_info` before doing its upsert.

Neither can produce a different answer per note. The collection exists after the first call of the process's life, and its metadata does not vary by which note is being embedded. That is **1,706 network round trips to Qdrant Cloud** across this sync that carry no information.

## Why it matters more than it looks

`embed` is the slowest queue by an order of magnitude — 738 ms average versus 47 ms for `indexing` — and it is the one that gates a note becoming searchable. These calls are serial preamble inside that 738 ms, against a hosted Qdrant over the public internet, so they are latency rather than CPU.

At concurrency 5 the queue's ceiling is `5 / 0.738s` ≈ 6.8 jobs/sec. Removing two of four round trips should move that ceiling directly, with no change to concurrency, pool size or task count — i.e. it is throughput that costs nothing to buy.

It also scales the wrong way: the waste is per-note, so a 100k-note import pays it 200k times.

## Suggested direction

`ensure_collection` is the classic once-per-process concern — a `persistent_term` flag, a boot-time call, or memoization keyed on collection name. `collection_info` needs a look at why the embed path reads it at all; if it is fetching vector dimensions or distance metric, those are static config and belong next to the collection setup rather than in the hot path.

Worth confirming whether the same preamble sits in the search read path, where the latency is user-visible.

## Related

- #1494 — the same shape on the DB side: each job re-resolves user/vault/subscription rows it already had.
- Both were found while auditing the fleet after the web/worker split. Neither is a regression from it; concentrating job execution on one worker just made them legible.

Issue #1495, opened 2026-08-28
RLS tenant_enter/tenant_exit is ~half of the web nodes' DB traffic on the request path


Rasbandit:
Measured in prod during a 1.4k-file sync, 2026-08-28 ~07:38 UTC.

## Numbers

Web nodes (the request path — HTTP, channels, CRDT rooms; no Oban queues since the split):

| source | queries/sec |
|---|---|
| `tenant_txn` | 109.4 |
| `notes` | 63.1 |
| **`tenant_exit`** | **56.3** |
| **`tenant_enter`** | **56.3** |
| `source_unavailable` | 37.6 |

`tenant_enter` and `tenant_exit` are exactly equal, as expected for a paired RLS context set/reset. Together they are **112 queries/sec** — comparable to `tenant_txn` itself, and larger than `notes`, the actual work.

So roughly one in three statements the web nodes send is RLS bookkeeping rather than data access, and it sits on the latency-sensitive path.

## Why raise it now

The point of the web/worker split was to make the request path as cheap as possible. Having done that, this is now the largest remaining non-essential cost on that path. It is round-trip bound rather than CPU bound, so it shows up as latency rather than as a CPU line on a dashboard — which is why it has not been obvious.

## Directions worth evaluating

Not proposing a fix — the tradeoffs here are real and RLS is a security boundary, so nothing should change without care:

- **Batch the set with the work.** If `SET LOCAL` can ride in the same round trip as the first statement of the transaction, the enter cost disappears without weakening isolation.
- **Skip the reset where the connection is discarded anyway.** `tenant_exit` matters when a connection returns to the pool dirty. If some paths tear the connection down regardless, the reset is redundant there.
- **Confirm no path sets tenant context more than once per transaction.** 56/s against 109/s `tenant_txn` is about one pair per transaction, which is correct — worth keeping it that way.

Any change here needs the multi-tenant isolation E2E tests as the gate, and probably a deliberate decision recorded in `docs/context/`, since weakening tenant context to save a round trip is the kind of optimisation that looks fine until it is a cross-tenant read.

## Not a regression

This predates the worker split — the split just made it the dominant remaining item on the request path. No errors, and Ecto queue time on web was 0.21 ms throughout.

Issue #1493, opened 2026-08-28
Bulk sync still opens a room per pushed note via the handshake path (#1409 premise not met)


Rasbandit:
Observed in prod during a real 1.4k-file Obsidian sync, 2026-08-28 07:33–07:39 UTC, on `engram-saas-prod:150`.

## What happened

```
crdt room LRU evicting 16 room(s) — resident=314 cap=64 backlog=234
                                    resident=303
                                    resident=287
                                    resident=271
                                    resident=255   ← 07:38:57
```

Resident peaked at **314 against a cap of 64** — 4.9x over — and drained at exactly 16 rooms per 30s tick. It did recover, so this is not an incident. The concern is the shape.

## Why the shape matters

Eviction is a fixed batch per tick and does not respond to how far over cap we are. Arrival during a bulk sync is bursty; drain is linear at ~0.53 rooms/sec. From 314 that is ~8 minutes back to cap.

That ratio does not hold as vault size grows. A 14k-file sync overshoots proportionally further while the drain rate stays constant, so time-above-cap grows linearly with vault size. There is no feedback term — nothing makes eviction more aggressive when the backlog is larger.

Worth deciding explicitly: is the cap a *target* the LRU should converge on quickly, or a soft advisory? Right now it behaves as the latter while being named as the former.

## The part we cannot see

Each resident room holds a Yjs doc inside the **y_ex NIF**, which is off-heap. `:erlang.memory()` does not count it, so every BEAM memory metric we have is blind to the dominant cost of a room.

During this sync, web reported:

```
beam_memory_allocated      168 MB
beam_memory_binary_total    17 MB
```

Neither figure includes the 314 rooms. The container cap is 820 MB. We currently have no metric that would warn us before room memory pushed a web task into an OOM kill — we would see the kill, not the approach.

Suggested: export resident room count and, if the NIF can report it, an approximate bytes-per-room, so the LRU cap can be tuned against something measured rather than assumed. Related prior art in `docs/context/crdt-room-lifetime-and-drain.md`.

## Notes

- All evictions logged with `role="web"`, which is correct — rooms belong where connections terminate.
- The plugin-side fix that stops bulk sync enrolling a room per idle note shipped as engram-app/Engram-obsidian#466 (`gate create-ack self-heal enroll on isLiveBound`). Rooms still reached 5x cap here, so either the syncing client predates that build or there is a second un-gated enrolment path.
- No errors during the window. 3,187 Oban jobs, zero exceptions.

[...]

PR #1504, opened 2026-08-28
perf(indexing): resolve the user once per job instead of up to four times

Rasbandit:
First step of #1502. **Elimination, not caching** — nothing is cached here; three of four calls stop existing.

## The problem

Indexing one note re-fetched the same `%User{}` from Postgres up to four times, each caller resolving it for itself:

```
lib/engram/indexing.ex:38     index_note/2
lib/engram/indexing.ex:81     prepare_index/2
lib/engram/workers/embed_note.ex:104   embed_budget_gate/1
lib/engram/workers/embed_note.ex:246   run_embed/2
```

Same user, same job, for a row that cannot change mid-job. Measured in prod 2026-08-28: **2.1 `users` queries per job** at ~13 jobs/sec, part of ~17 queries per job overall.

## Why not a cache

A cache converts a slow repeated call into a fast repeated call. The call is still made, still allocates, still has to be correct under invalidation. 17 queries per job with a user cache is still 17 calls — they just land on ETS.

`EmbedNote.perform/1` now resolves once and threads the struct down. The other three lookups are deleted.

## API

`index_note/3` and `prepare_index/3` take the user as an **optional** third argument. All six existing test modules and any future caller keep working with two args; only the hot path passes what it already has.

## The guard

Counts real Ecto queries rather than inspecting call sites — a grep-shaped test passes while a helper three frames down re-fetches.

The counter keys on `meta.source == "users"`, which is the same field the prod measurement uses:

```promql
sum by (source) (rate(engram_prom_ex_ecto_repo_query_total_time_milliseconds_count))
```

So the test and the dashboard count the same thing. Asserts **0** queries when the user is threaded and **exactly 1** when it is not — the second direction matters, since a test that stops observing the query would otherwise pass silently.

## One trap, recorded in the test

The struct you thread must be one the callee would have fetched. A factory struct lacks the encrypted DEK fields and fails with `:no_dek`. `EmbedNote.perform/1` threads the result of `get_user!/1`, so prod is unaffected — but it is the obvious way to get this wrong.

## Verification

`mix format --check-formatted`, `mix compile --warnings-as-errors`, `mix credo --strict` clean. All indexing suites plus every worker suite and search integration: **270 tests, 0 failures**.

## Not included

- `ensure_collection` per note (#1501) — same shape, separate change to the Qdrant client
- User/vault TTL cache (#1502 step 3) — needs an eviction story for DEK rotation and plan changes
- #1493 (CRDT handshake) and #1495 (RLS boundary) — deliberately out of scope; one is the highest-risk protocol path, the other is the tenant isolation boundary

### DIFF
--- a/.sobelow-skips
+++ b/.sobelow-skips
@@ -6,10 +6,10 @@ Config.CSWH: Cross-Site Websocket Hijacking,lib/engram_web/endpoint.ex:45,6EA19F
 Config.HTTPS: HTTPS Not Enabled,config/prod.exs:0,2B5C077
 Misc.BinToTerm: Unsafe `binary_to_term`,lib/engram/crypto.ex:581,7EA51D6
 Misc.BinToTerm: Unsafe `binary_to_term`,lib/engram/crypto/user_dek_rotation.ex:1411,772DFE4
-Misc.BinToTerm: Unsafe `binary_to_term`,lib/engram/notes.ex:4665,1171D7E
-Misc.BinToTerm: Unsafe `binary_to_term`,lib/engram/notes.ex:5000,8A8E13
+Misc.BinToTerm: Unsafe `binary_to_term`,lib/engram/notes.ex:4685,32355C9
+Misc.BinToTerm: Unsafe `binary_to_term`,lib/engram/notes.ex:5020,13240BD
 SQL.Query: SQL injection,lib/engram/attachments.ex:1049,2991CD5
-SQL.Query: SQL injection,lib/engram/notes.ex:6650,199FE8B
+SQL.Query: SQL injection,lib/engram/notes.ex:6670,242F41F
 Traversal.FileModule: Directory Traversal in `File.read!`,lib/engram/legal/seeder.ex:84,12C6307
 Traversal.FileModule: Directory Traversal in `File.read!`,lib/engram/release/preflight.ex:114,30D4B80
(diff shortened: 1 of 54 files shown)


Rasbandit:
Added the #1496 P0 to this PR (commits 823e356c, df0e5352).

**Fix** — all 31 Oban workers now define a finite `timeout/1` (5 min crdt_checkpoint/indexing/default, 10 embed, 15 maintenance, 60 crypto_backfill/export/cleanup). Previously every one inherited Oban's `:infinity`, so a blocked job held its slot forever and the queue stopped with `executing` pinned at the limit — the 2026-08-28 stall.

**Detection** — `ObanDiscardHandler` now logs an `Oban.TimeoutError` on *any* attempt, not just at `:discarded` (which on embed is five backoffs later, hours after the queue stopped), and emits a new `..._oban_timeout_total{worker,queue}` counter. Occupancy metrics are blind to this failure by construction.

**Guard** — `Engram.ObanWorkerTimeoutTest` enumerates workers from their .beam attribute chunks, so a new worker cannot silently reintroduce `:infinity`.

Also fixed three assertions that were red under concurrent load rather than on a defect (details in df0e5352).

Verified: 4843 tests / 0 failures, format + credo --strict + dialyzer + sobelow clean.

Follow-up owed in engram-infra: a Grafana alert on `rate(engram_prom_ex_oban_timeout_total[15m])`, once the metric reaches prod via a release.

PR #1517, opened 2026-08-29
test(e2e): frontmatter round-trip between Obsidian and the web SPA

Rasbandit:
Reproduction suite for engram-app/Engram-obsidian#483. Landed before the fixes,
as the bug report asked.

## Why nothing caught this

The whole CRDT e2e suite drives body text. The one place frontmatter is asserted
(`test_100_frontmatter_echo_single_lineage`) is a single-device lineage check.
Nothing exercised frontmatter across two interfaces, which is exactly where the
plugin's missing half lived.

## The four cases

Both directions and both bindings, because the binding is the variable that
decides the outcome:

| case | unfixed plugin |
|---|---|
| closed, obsidian -> web | PASSED (disk-save path, the control) |
| open, obsidian -> web | FAILED |
| open, web -> obsidian | FAILED |
| bad YAML keeps good keys | PASSED (weak oracle, see below) |

That result corrected the diagnosis. Outbound from an open note was believed to
be merely slow, waiting on the file save. It is never delivered at all:
`handleModify` skips the disk-driven CRDT route while bound (`sync.ts:3594`) and
the live binding drops frontmatter keystrokes (`classifyEditSpan` ->
`"frontmatter"`), so there is no route between them.

Against the fixes on Engram-obsidian#482 all four pass, and the run drops from
332s to 92s because convergence happens instead of timing out.

## Weak oracle, kept deliberately

The bad-YAML test passes on the unfixed plugin: single-device, nothing triggers
a flush, so the deletion happens in the doc and never reaches disk. It is kept
for the shape, not the proof. The real coverage for that defect is
`tests/crdt-note-seed-frontmatter.test.ts` in the plugin, where it is red before
the fix.

## Harness

Driving the SPA needs the properties widget, not the CodeMirror editor — that
editor is body-only, so frontmatter cannot be typed into it. `WebSpaPeer` gains
`add_property` / `set_property` / `remove_property` / `property_value_locator`.

Locally these need `xvfb-run` + `E2E_WEB_HEADED=1`: headless Chromium on the dev
box produces no rAF frames, so every Playwright click times out at sign-in
(`docs/context/headless-chromium-no-raf-playwright.md`). CI runs them headless
without issue.

## Also included

`test_room_cap_pressure.py`, from chasing an earlier theory that convergence
collapses when concurrently-diverged cold notes exceed
`CRDT_MAX_RESIDENT_ROOMS`. It does not — 54 cold notes against a cap of 3 all
converge — so these reproduce nothing. Kept because fleet convergence under cap
pressure had no coverage and every existing test converges one or two notes at a
time. Happy to drop them if that is not wanted.

Refs engram-app/Engram-obsidian#483

### DIFF
--- a/e2e/helpers/web_spa.py
+++ b/e2e/helpers/web_spa.py
@@ -103,6 +103,46 @@ class WebSpaPeer:
     async def read(self) -> str:
         return await self._editor().inner_text()
 
+    # --- Properties widget (frontmatter) -----------------------------------
+    #
+    # The SPA edits frontmatter as TYPED PROPERTIES against the doc's
+    # `frontmatter` / `frontmatter_order` / `frontmatter_types` shared types,
+    # not as YAML text in the CodeMirror editor (which is body-only). So a
+    # frontmatter round-trip test cannot go through `append`/`read` — it has to
+    # drive this widget, the way a user does.
+
+    def properties_locator(self):
+        """The properties section, for auto-retrying `expect(...)` asserts."""
+        return self._page.get_by_test_id("note-properties")
+
+    def property_value_locator(self, key: str):
+        """The VALUE field of property `key`. `aria-label` is `"<key> value"`
+        (property-fields.tsx), which is stable across the type variants."""
+        return self._page.get_by_test_id(f"property-row-{key}").get_by_label(f"{key} value")
+
+    async def add_property(self, key: str, value: str) -> None:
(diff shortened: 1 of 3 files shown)


Rasbandit:
Superseded by #1520 — the e2e commit is merged in there, alongside the properties widget work and the `tags: true` backend fix, so all the frontmatter work in this repo lands as one PR.

Issue #858, opened 2026-07-02
Sharing/publish: design public note links — or remove the dead /share/*path route


Rasbandit:
## Problem

From the 2026-07-02 backend audit. `router.ex` whitelists `get "/share/*path"` as an SPA route, but nothing exists behind it: no share route in `frontend/src/router.tsx`, no `/api/share*` endpoints, no share schema, no public-read pipeline. The route is vestigial and ships ambiguous intent.

## Why it matters

Public read-only note links are the biggest viral-growth loop in PKB products (Obsidian Publish, Notion public pages) and a classic paid-tier lever. Per-user encryption at rest means a share requires a deliberate decrypt-and-serve path — it will never happen by accident, so it needs design.

## Proposal

Decide one of:
- **A. Design + ship:** share-tokens table, public pipeline bypassing Auth/VaultPlug (tokenized URL, optional expiry), careful crypto/redaction review, SPA share page. Effort M–L.
- **B. Defer:** remove the `/share/*path` whitelist from router.ex until the feature is scheduled (S), so the router doesn't advertise a surface that doesn't exist.

Rasbandit:
**Resolved by taking the "or" branch — closing.** This issue offered two exits: design public note links, or remove the dead route. PR #878 (`chore: remove dead /share/*path SPA route`) took the second one and merged.

Confirmed on `main`: no `/share` route remains anywhere in `frontend/src`.

The dead-code problem this issue was filed for is gone. Public note sharing as a *feature* is a product decision that no longer has anything stale attached to it — if it gets picked up it deserves a fresh issue with real requirements, not this one's "or remove the route" framing.

PR #1519, opened 2026-08-29
refactor(routing): namespace vault URLs under /v

Rasbandit:
## What

Vault-scoped SPA URLs move from `/:slug` and `/:slug/:id` to **`/v/:slug`** and
**`/v/:slug/:id`**. Nothing dynamic sits at the URL root anymore.

Both reserved-slug lists are **deleted**: `Vault.@reserved_slugs` (18 entries)
and its hand-maintained TypeScript mirror `frontend/src/api/reserved-slugs.ts`.

## Why

Vault slugs are `slugify(user-supplied vault name)`. While they lived at the
root, every top-level segment was ambiguous in both directions:

- a vault named "Link" or "Settings" was shadowed by the static SPA route of
  the same name, and
- a typo'd `/api/notez` matched `/:slug/:id` and returned an HTML 200 that
  masked a broken API call (#858).

The mitigation was the reserved list, folded into `unique_slug/3` so a
colliding name silently became `link-2`. It worked, but nothing derived it
from the actual routers, so every new top-level route silently widened the
hazard.

### It had already drifted

The Cloudflare **zone** firewall (`engram-infra`, `kind = "zone"`, so it covers
`app.engram.page`) blocks these paths outright:

```
/.env  /.git/  /wp-admin  /wp-login  /wp-content  /wp-includes
/phpmyadmin  /phpinfo  /server-status  /cgi-bin  /vendor
```

Nine of those are reachable through `slugify` (`"Vendor"` → `vendor`,
`"WP Admin"` → `wp-admin`). **None were in the reserved list.** A user who
named a vault "Vendor" got a Cloudflare block page — not a 404, not a renamed
slug — one hop before Phoenix, where neither list nor deny-list could see it.
The prefix fixes that for free.

## Why not prefix the app routes instead

Every externally-baked URL is an **app** route; zero are vault routes:

| Location | URL |
|---|---|
| `plugin/src/sync-progress-modal.ts:94` | `${webUrl}/settings/billing` |
| `lib/engram/onboarding.ex:452` | `app.engram.page/onboard/agreement` (in sent email) |
| Clerk dashboard | sign-in / sign-up URL config |
| `.well-known/oauth-authorization-server` | authorize URL, cached by registered MCP clients |
| `engram-marketing` docs | `app.engram.page/settings/api-keys` (12 pages) |

Moving app routes would break shipped plugin builds and already-delivered
email. Moving the vault route broke 17 internal call sites and nothing else.
It also wouldn't have worked: `/api`, `/socket`, `/.well-known`, `/assets`,
`/metrics` and `/favicon.ico` are Plug.Static mounts, endpoint macros and an
RFC-mandated path — not app routes you can relocate — so a root wildcard would
still have needed both lists.

## Kept deliberately

- **The router deny-list.** Its original job (stopping fall-through to the root
  wildcard) is gone, but its second job outlives it: a clean non-HTML 404 for
  API-shaped paths instead of Phoenix's default HTML error page.
- **`/note/:id`**, which still redirects to the canonical vault URL.

## Falls out of the move

The vault slug is now always URL segment 2, which `sentry.ts` redacts
unconditionally. Previously it was segment 1 and stayed out of Sentry only by
missing an allowlist.

## No back-compat redirect

Pre-customers, so there are no live bookmarks to preserve. Old `/:slug/:id`
URLs 404.

## The guard that replaced the list

`frontend/src/router.test.tsx` walks the built router config and fails if any
route resolves to a root-level `/:param`. That is the check the hand-maintained
list could never do. Precedent: `HostRewrite.__api_top_segments__` is compared
against `Router.__routes__` the same way.

## Verification

- `mix test` — full suite
- `mix credo --strict` — no issues
- `mix dialyzer`
- `mix format --check-formatted`
- `bun run test` — 1656 tests
- `tsc --noEmit`, `biome check`

## Notes for the reviewer

- The first sweep grepped three call syntaxes and missed 12 builders in
  `wiki-link.ts` and friends; the correct sweep is ``grep -rn '`/\${' src/``.
  Second commit covers it.
- `wiki-link.test.ts` used `"v"` as its fixture slug, which would have made
  expectations read `/v/v/...`; renamed to `work`.
- Deep-linking `/v/:slug/wiki/*` on a **hard page load** still 404s at Phoenix,
  because the backend only routes 1- and 2-segment vault paths. That is
  pre-existing behaviour, unchanged here — wikilinks resolve client-side.

Context doc: `docs/context/vault-url-prefix-and-collision-surface.md`


Rasbandit:
## CI state

`ci` rollup: **pass**. All gating jobs green.

`e2e-crdt` is red and **report-only**. Both runs failed in the fanout-pacer cold-path family, and the set rotates between runs:

| run | failures |
|---|---|
| 1 | `test_cold_send_over_fanout_opens_no_room`, `test_deaf_live_bound_note_converges_via_socket_replay` |
| 2 | `test_cold_send_over_fanout_opens_no_room`, `test_concurrent_cold_edits_survive_over_fanout` |

Pre-existing, not this branch:

- `test_cold_send_over_fanout_opens_no_room` **fails on main** with a byte-identical message in run [33246147019](https://github.com/engram-app/Engram/actions/runs/33246147019), and **passes on main** in run [33251105801](https://github.com/engram-app/Engram/actions/runs/33251105801). Same test, both outcomes, without this branch.
- None of the three reference vault slugs or SPA routes.
- Closest existing issue: #844 (same failure class, different test). Playbook: `docs/context/fanout-pacer-hot-cold-and-testing.md`.

Counter-evidence that the URL move is sound: `test_obsidian_to_web_live` and `test_web_to_obsidian_live` — the tests that drive a real browser through the SPA — **pass here**, and `test_obsidian_edit_renders_live_in_spa` was among the failures on main's red run. Those drive `/note/{id}` → `/v/:slug/:id` through the live redirect, so the new URLs are exercised end-to-end.

Issue #1550, opened 2026-09-01
filemeta_v0 keeps a claim for a note whose row never existed — vault index projection wedged permanently (blocks #1401)


Rasbandit:
Found while checking #1401's prerequisite ("`filemeta_v0` has been populated for real vaults"). It has — and one vault's index has been **permanently unable to converge for 16+ hours**, with two of the user's notes never reaching the server.

## The evidence

`engram_prom_ex_crdt_index_projection_total{phase="unresolved"}` was flat zero for 8 days, then went non-zero at **2026-09-01T00:12Z** and has stayed non-zero on every run since. It is not a rate — it is the **same two note_ids on every single run**:

```
vault index projection did not converge: applied=0 unresolved=2 passes=1
vault index projection skipped an entry naming an unknown note
  user_id  019ec4e2-a865-73b9-8018-d37cbcf9596a
  vault_id 12dc6735-52f2-4ce4-9117-91f0ce2389a7
  note_id  01a05a4f-1cf8-79fe-a0e9-dba790c6f00d
  note_id  01a05a56-1f8e-7cea-8dda-bc565e8aeec6
```

Last run 2026-09-01T06:30Z, still `applied=0 unresolved=2`. Nothing clears it.

## What happened

Those note_ids are uuid7. Decoded, they were minted at **00:12:09Z** and **00:19:48Z** — and the *first* `crdt_msg → :error :not_found` for `01a05a4f` is at **00:12:09Z**, the same second it was minted. The client pushed ops for a freshly-minted id before any `crdt_create` could have round-tripped.

Six distinct ids took `:not_found` in that one 8-minute window:

```
00:11:16  01a05a4e-4c68…      00:12:09  01a05a4f-1cf8…  ← wedged
00:11:37  01a05a4e-9ebb…      00:18:43  01a05a55-2049…
00:11:47  01a05a4e-c682…      00:19:48  01a05a56-1f8e…  ← wedged
```

**Four self-healed. Two did not** — and the difference is whether the index claim had already landed.

## The defect

The `:not_found` recovery path (#955's id-map reconcile) heals the **client's** id map. It does not release the **index's** claim.

So the client writes its `filemeta_v0` entry independently of `crdt_create` being acked. When the create then fails to land, the id is abandoned client-side but the claim survives in the index forever — naming a note that does not exist and cannot be created, because nothing is left holding that id.

[...]

PR #1551, opened 2026-09-01
test(e2e): an orphaned index claim must recover the note, not strand it

Rasbandit:
The e2e for engram-app/Engram-obsidian#497. **Branch named identically to the plugin branch on purpose** — priority-3 auto-pairing then runs this against the plugin PR's code rather than plugin `main`, which is the difference between testing the fix and testing its absence.

## Why it exists

#497 is unit-tested at the level of "the create is **enqueued**". Nothing asserted the thing that actually matters, and that no unit test can reach: the stranded note gets to the server, under the same id, and the vault's index converges again.

This is the same gap `test_index_crdt_client_writes.py` was written to close — its docstring notes a write-only client shipped through three green CI runs because every test had our client on both ends.

## The state under test

`getOrMint` publishes a `filemeta_v0` claim the moment it mints, without waiting for `crdt_create` to be acked. When the create fails to land:

- the index claims `path → note_id`
- no `notes` row exists for that id
- the file is still on disk

Every `crdt_msg` is dropped `note_not_found` and `ProjectVaultIndex` retries the unresolvable entry forever — `applied=0 unresolved=N` for that whole vault. Measured in prod 2026-09-01: two notes stranded 16+ hours (#1550).

## Two details that make it a real test rather than a green one

**Clearing `crdtHead` is not optional.** Deleting the row alone leaves the client still believing `hasServerNote`, because the head survives — that is the state after a *remote delete*, not after a lost create, and the repair deliberately does not touch it. Without also clearing the client's head the test constructs a different state and fails for the wrong reason. It asserts the clear took effect rather than assuming it.

**The row goes away by raw SQL, not `delete_note_by_id/4`.** The well-behaved path tombstones the id and releases the claim — precisely the state we are trying *not* to build.

Also asserts the recovered note carries the **same id**, not merely that a note reappeared: a fresh mint would satisfy "it came back" while stranding the local Y.Doc holding the user's content and leaving a second claim behind.

The trigger is opening the note in Obsidian's editor. The repair fires on the `note_not_found` reply, and `canSendLive` holds ordinary ops behind `hasServerNote`, so what has to escape is a handshake — which means the note must be enrolled. In the prod case the same STEP1 came from `reEnrollUnsent` on reconnect.

## Honest status

I have not run this locally — it needs the CI stack plus Obsidian. It is lint- and format-clean (`ruff check`, `ruff format --check`) and reuses the `load_doc/2` index oracle and the CDP open-file idiom from existing passing tests, but its first real execution will be in CI here. If it is red, treat the test as suspect before the fix.

Refs #1550, engram-app/Engram-obsidian#497

### DIFF
--- a/e2e/tests/crdt/test_orphaned_index_claim.py
+++ b/e2e/tests/crdt/test_orphaned_index_claim.py
@@ -1,0 +1,205 @@
+"""An index claim the server cannot honour must repair itself, not strand the note.
+
+`getOrMint` publishes a `filemeta_v0` claim the moment it mints, without waiting
+for `crdt_create` to be acked. When the create fails to land, three sources of
+identity disagree and stay that way:
+
+  * the index claims `path -> note_id`
+  * no `notes` row exists for that id
+  * the file is still on disk
+
+The note then lives on exactly ONE device. Every `crdt_msg` for it is dropped
+`note_not_found`, and `ProjectVaultIndex` retries the unresolvable entry forever
+— `applied=0 unresolved=N` for that whole vault, permanently, which also pins
+the only signal we have for index/row disagreement.
+
+Measured in prod on 2026-09-01 (engram-app/engram#1550): six ids took
+`note_not_found` in one 8-minute window on one vault. Four healed. The two whose
+claim had already landed stayed stranded for 16+ hours with no user-visible
+signal, because `ensureNoteIdMapped` early-returns on `pathForId !== null` and
+for an orphan the mapping IS the claim — the id vouches for itself.
+
+The plugin fix (Engram-obsidian#497) re-drives the create from the
+`note_not_found` reply. Its unit tests assert the create is ENQUEUED. This
+asserts the thing that actually matters and that no unit test can: the stranded
+note reaches the server, under the same id, and the vault's index converges
+again.
+
+Refs engram-app/engram#1550, engram-app/Engram-obsidian#497.
+"""
+
+from __future__ import annotations
+
+import os
+import time
+import uuid
+
+import pytest
+
+from helpers.backend_rpc import backend_rpc
+from helpers.latency import DELIVERY_TIMEOUT
+from helpers.vault import write_note
+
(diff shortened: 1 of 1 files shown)


Rasbandit:
**The e2e is verified to discriminate.** Both directions run, not just the green one.

| plugin ref | plugin SHA | result |
|---|---|---|
| `fix/orphan-index-claim` (the fix) | `c8b2fdca` | `28 passed` — this test **PASSED** |
| `main` (no fix) | `05d3223` | `1 failed, 27 passed` — this test **FAILED** |

The failure is on the right assertion, not on setup:

```
FAILED tests/crdt/test_orphaned_index_claim.py::test_an_orphaned_claim_re_creates_the_note_under_the_same_id
  - AssertionError: note row 01a05e29-141c-779f-ac91-d6340399ffc1 did not reappear within 120.0s
```

That is the stranded note, reproduced: claim present, row gone, and without the repair it never comes back.

### Two things worth recording about how this was measured

**The auto-pairing works and is visible.** The passing run logs `🔗 Auto-linked plugin branch: fix/orphan-index-claim` and pins `c8b2fdca`, so the green result is against the fix rather than plugin `main`. 28 collected / 0 skipped also rules out the `E2E_ENABLE_CRDT` guard silently skipping it.

**`gh workflow run ... -f plugin_branch=main` did NOT work**, and it fails in a way that looks like success. The fingerprint job still resolved `matching branch (fix/orphan-index-claim)`, so the plugin SHA was unchanged from the prior run, the whole e2e set was skipped as a no-op, and the run reported **success** with `e2e-crdt: skipped`. Read as-is that is a green "proof" of nothing.

The reliable way to force the unpaired run was a throwaway branch deliberately *not* named to match any plugin branch, so resolution falls through to priority 4 (`plugin main HEAD`). That branch has been deleted. Worth a line in `oauth-e2e-pairing-and-token-binding.md`, since the doc presents the `-f plugin_branch=` override as the clean way to do exactly this.

PR #1588, opened 2026-09-08
feat(api): minimum plugin version floor

Rasbandit:
## Summary

- Refuse clients below `Engram.PluginVersion.minimum/0` on **both** transports: `426 plugin_upgrade_required` from a new `:authed_api` plug, and the same reason from `ChannelGate` on a sync/crdt join. The socket half is the load-bearing one — sync runs over channels, so a REST-only floor would leave a blocked client syncing.
- The floor is a **compile-time constant**, not an env var. You need it exactly when you deploy a backend old clients cannot talk to, so it is a property of that release and cannot drift between nodes or be forgotten in a task definition.
- **Ships as a no-op.** `@minimum` is set to the current plugin version, and no released plugin sends the header yet, so nothing is refused today.

### Everything unreadable is ALLOWED

Absent, empty, unparseable, over-long, non-binary — all pass. Every client shipped to date predates the header, as do the web SPA and every MCP client; failing closed on absence would brick the installed base on day one. Only a version parsed and found strictly below the floor is blocked.

It follows that this is a **compatibility gate, not a security control** — the header is self-reported, and anyone who wants past it sends a higher number. Never put an entitlement or authorization decision behind it.

### Notes for review

- `ChannelGate.check/2` becomes `check/3` with a **mandatory** `plugin_version`, for the same fail-open reason `api_key` is mandatory: an optional argument makes the ungated call the convenient one that still compiles and passes dialyzer.
- Precedence is after `AccountDeleted` (terminal) but ahead of onboarding and subscription — telling a user to finish the wizard is useless advice for a client that cannot speak the protocol.
- Not enforced at `UserSocket.connect/3`, which would be one line instead of a threaded parameter. `connect/3` can only return `:error`, which reaches the client as an anonymous close with no payload; the client would retry forever with nothing to show the user.
- `ws connect` now carries `plugin_version` — one field per socket, not per request. It reaches **CloudWatch and not Loki**: `:info` + `:websocket` is in `@info_to_loki` but absent from the Fluent Bit category regex. Pre-existing mismatch, documented in `category.ex`, deliberately not widened here.
- The `.sobelow-skips` commit is a pure line shift (172 → 173) caused by the new plug line, regenerated wholesale per `docs/context/sobelow-silent-no-op-and-fingerprint-skips.md`.

## Test plan

- [x] `mix test test/engram/plugin_version_test.exs` — ordering incl. numeric-not-lexical (`1.9.0 < 1.28.0`) and pre-release sorting
- [x] `mix test test/engram_web/plugs/require_plugin_version_test.exs` — 426 body shape; absent/unparseable/over-long all pass
- [x] `mix test test/engram_web/plugin_version_gate_channel_test.exs` — both channels, plus two precedence pins: upgrade beats onboarding, and a foreign topic still reports `unauthorized` (the plugin's identity self-heal keys on that string)
- [x] `mix test test/engram_web/` — 1527 tests, 0 failures
- [x] `mix format --check-formatted`, `mix credo --strict`, `mix dialyzer` — all clean

Paired client change: engram-app/Engram-obsidian#TBD (independent — this PR allows an absent header, so neither branch needs the other).

https://claude.ai/code/session_019ZU8U3d3gxBK8wi2qR6jfc


Rasbandit:
## Adversarial review pass — three real defects fixed

Reviewed the diff adversarially after opening. Findings that changed the code:

**1. The floor would have blocked the release that fixes it.** Comparison used `Version.compare/2` — semver ordering, under which `1.31.0-beta.3 < 1.31.0`. But `engram-obsidian-sync/scripts/release-version.mjs` names a preview build after the release it previews (`betaVersion` → `1.31.0-beta.3`, `prVersion` → `1.31.0-pr.512.g876f2c2`) and `pr-build.yml` stamps that into the shipped `manifest.json`. Shipping `@minimum "1.31.0"` alongside plugin 1.31.0 would have refused every beta tester and PR reviewer running a build that **already contains the fix**, then sent them to a plugin pane with nothing newer to install — a dead end, on the feature's first real use. Comparison is now `{major, minor, patch}` with the pre-release tag discarded, and the repo's real formats are in the test table; their absence is what let this through.

**2. The gate sat ahead of the liveness stamp, so blocked-but-active accounts would be deleted.** A refused request/join never wrote `last_active_at`, and `InactivityCleanup` soft-deletes at 90 days — dropping Qdrant points and S3 attachments. A user syncing daily on an old plugin would have been deleted *for being blocked*. `ChannelGate`'s own moduledoc refuses that exact trade for `api_access/2`: *"the opposite error is soft-DELETING a blocked-but-active account, and data loss beats a stale row."* Moved after `BumpActivity` / `stamp_activity` on both transports. Cost: onboarding is now reported before the upgrade verdict, pinned by a test that explains why.

**3. Both wiring links were untested.** Deleting the `plug` line from `:authed_api`, or `plugin_version:` from `UserSocket`'s assigns, left every test green while the respective floor was gone entirely — the silent-skip shape the mandatory third argument on `check/3` exists to prevent, reproduced one layer up. Added a pipeline-level 426 test and the first `UserSocket` connect test in the repo.

### Also

- `plugin_version` is clamped to 32 bytes before reaching Logger metadata. It is attacker-controlled, prod serializes all metadata, and both `request_logger.ex` (`user_agent`) and `logs.ex` (`plugin_version`) already clamp the same class of value.
- Corrected a false comment: 1.28.0 did **not** ship the header. Documented the consequence — reporting starts at some version R, so the gate's reachable range is `[R, floor)` and **every client older than R is permanently exempt**. The oldest clients, the ones most likely to speak a protocol we changed, are the ones this can never reach.
- Tests now assert literals instead of re-deriving expectations from the module under test.
- Fixed the stale `router.ex:NN` citations and the second copy of the pipeline ledger in `docs/context/onboarding-gate-is-http-only.md`.
- The wiring test file is `async: false` — it flips `:billing_enabled` in the global application env, and run async that leaked into six unrelated `MultiTenantTest` cases. Caught by the full suite, not by the file itself.

### Verification

- `mix test` — **5152 tests, 0 failures**
- `mix format --check-formatted`, `mix credo --strict`, `mix dialyzer` — clean
- `.sobelow-skips` regenerated for the router line shift (one entry moved, nothing added or dropped)

`e2e-clerk` red here is pre-existing on `main` and unrelated — an Obsidian 1.12.7 → 1.13.7 runner upgrade moved Settings into its own window. Fixed separately in #1590.

Issue #1635, opened 2026-09-15
p0: CIMD conformance covers exactly one vendor document (MCPJam's) — a refused vendor cannot turn it red


Rasbandit:
## Nothing verifies we can still talk to the vendors we claim to support

We found out ChatGPT cannot connect to our MCP server because a user hit it, quit, and deleted their account. It had been failing silently since at least 2026-09-09, and our docs still list ChatGPT as a verified connector.

The OAuth suite did not catch it. It is not missing coverage — it is pointed the wrong way.

## What the current tests actually assert

- `test/engram/oauth/cimd_test.exs:129` — *"rejects a document asking for a confidential auth method"*, passing.
- `test/engram_web/controllers/oauth_register_controller_test.exs:315` — *"still rejects an auth method we do not implement"*, posting `private_key_jwt`, asserting 400, passing.

Both encode a product decision as correctness. They stay green on the day that decision starts costing us a vendor.

Two structural gaps reinforce it:

**Every CIMD test feeds us a document we wrote ourselves.** All of them go through `FetcherMock`. No test has ever fetched a real vendor metadata document, so a vendor changing their auth method turns nothing red.

**Every fixture is Claude.** The shared `document()` helper (`cimd_test.exs:37`) hardcodes `"client_name" => "Claude Code"` with a loopback callback; `claude.ai` redirect URIs run through the rest of the OAuth suite. ChatGPT appears in the whole backend suite exactly once — `test/engram/onboarding_test.exs:554`, inside `~w(claude chatgpt grok mistral ...)`, a list of UI name slugs. Never in an OAuth flow.

## Ask

A scheduled conformance check (cron, not PR-gating — it depends on third-party endpoints) that for each vendor we advertise support for:

1. Fetches their published CIMD document / OAuth metadata.
2. Asserts we would accept it — `token_endpoint_auth_method`, `redirect_uris`, `client_name`, content-type.
3. Fails loudly, to the existing Discord alert path, naming the vendor and the offending field.

Vendor list to seed from `docs/context/connections-client-identity.md:167` — Claude, ChatGPT, Grok, Mistral.

This is the thing that converts "ChatGPT is a verified connector" in our docs into a claim CI can falsify. The cheap version is step 1 + 2 only; it would have gone red on 09-09 instead of us learning it from a lost signup.

Rasbandit:
## Correction — the conformance suite exists, runs daily, and was green

[...]

Issue #1634, opened 2026-09-15
p0: discovery advertises client_secret_post/basic but the CIMD path refuses them


Rasbandit:
## We advertise two auth methods we then refuse

`lib/engram_web/controllers/well_known_controller.ex:112` publishes:

```elixir
token_endpoint_auth_methods_supported: [
  "none",
  "client_secret_post",
  "client_secret_basic"
]
```

`lib/engram/oauth/cimd.ex:355` then rejects any CIMD document naming anything but `none`:

```elixir
document["token_endpoint_auth_method"] not in [nil, "none"] ->
  {:error, :confidential_not_supported}
```

A client that reads our discovery metadata, picks `client_secret_post`, and presents it via CIMD gets a 400 for doing exactly what we advertised. The failure is `mcp_cimd_rejected` + `GET 400` on `OAuthAuthorizeController#show`, with no indication to the client that the method it chose from our own list is the problem.

DCR clients can still use the secret methods, so the advertisement is not wholly false — it is false **on the CIMD path specifically**, which is the path we steer clients onto by advertising `client_id_metadata_document_supported: true`.

## Pick one

1. Narrow the advertisement per-path, or
2. Support the secret methods on CIMD, or
3. Keep refusing but return an error that names the supported method, so the client can retry as a public client instead of dead-ending.

Whichever way this lands, the comment block at `well_known_controller.ex:107-111` needs updating — it currently explains only why `none` must stay first-class, and says the secret methods are "additive, for server-side connectors that cannot hold a public client." Those connectors cannot actually use them over CIMD.

Related: the `private_key_jwt` gap filed separately.

Rasbandit:
An adversarial review of #1640 landed on this exact issue from the code side, with a concrete failure scenario and the reason it is not a one-line fix.

**The site:** `lib/engram/oauth/cimd.ex:419`

```elixir
document["token_endpoint_auth_method"] not in [nil, "none", "private_key_jwt"] ->
  {:error, :confidential_not_supported}
```

This is the same "preference treated as a requirement" defect that #1639 and #1640 removed from the two arms directly above it. It was left behind.

**Scenario:** a vendor publishes `token_endpoint_auth_method: "client_secret_basic"` with `token_endpoint_auth_methods_supported: ["client_secret_basic", "none"]`. That is a *reasonable* document to publish, because our own authorization-server metadata advertises `client_secret_basic` as supported (`well_known_controller.ex:117-122`) — which is precisely what this issue is about.

[...]

PR #1637, opened 2026-09-15
feat(oauth): accept private_key_jwt on CIMD

### LINKED ISSUE #1633: p0: ChatGPT cannot connect — CIMD refuses private_key_jwt, and we have no client_assertion support
## ChatGPT cannot connect to our MCP server, and has never been able to

`lib/engram/oauth/cimd.ex:355` refuses any CIMD document that names an auth method other than `none`:

```elixir
document["token_endpoint_auth_method"] not in [nil, "none"] ->
  {:error, :confidential_not_supported}
```

ChatGPT identifies itself with a signed client assertion (`private_key_jwt`), not a shared secret. Verified from their side: `https://chatgpt.com/oauth/jwks.json` returns 200 `application/json` with an RSA signing key whose `kid` is **`cimd-20260428030119`**. They are doing CIMD the standards-track way; we reject it before anything else runs.

## Prod evidence

`mcp_cimd_rejected` / `reason=:confidential_not_supported` / `cimd_host=chatgpt.com`:

- 2026-09-09 ~08:41Z ×2 (`engram-saas-prod:157`)
- 2026-09-14 15:28Z ×2 (`engram-saas-prod:162`)

**Zero successful ChatGPT grants in 30 days.** The 09-14 pair belongs to a new signup who deleted their account 5 minutes later.

The user-facing failure is a `GET 400` on `OAuthAuthorizeController#show`. Nothing tells them why.

## We have none of the machinery

- `@valid_auth_methods` (`lib/engram/oauth/client.ex:27`) is `~w(none client_secret_post client_secret_basic)` — `private_key_jwt` is not in the vocabulary.
- `client_assertion` appears **zero** times in `lib/` and `test/`. We could not verify their JWT even if we accepted the method.
- The only occurrence of `private_key_jwt` in the entire repo is `test/engram_web/controllers/oauth_register_controller_test.exs:315`, a test named *"still rejects an auth method we do not implement"* asserting a 400.

## Scope

Accept `private_key_jwt` on the CIMD path: read the JWKS URI from the metadata document, fetch + cache it (SSRF guard applies, same as the document fetch), verify the `client_assertion` JWT at the token endpoint, bind it to the client. PKCE stays mandatory.

The original refusal was a defensible call — see the comment at `cimd.ex:349-354`, no secret was ever minted so honouring or downgrading both fail opaquely. That reasoning holds for `client_secret_*`. It does not hold for `private_key_jwt`, which needs no minted secret at all.

Our own docs already list ChatGPT as a supported, verified connector (`docs/context/connections-client-identity.md:167`, `docs/context/mcp-oauth.md:92`). That claim is currently false.

Rasbandit:
Closes #1633. Refs #1634, #1635.

## Why

ChatGPT has never been able to connect to our MCP server. It declares `token_endpoint_auth_method: private_key_jwt` and does not negotiate down even though we advertise `none` first, so `Cimd.validate_document/2` refused its document before anything else ran.

Prod evidence: `mcp_cimd_rejected` / `reason=:confidential_not_supported` / `cimd_host=chatgpt.com` — **5 refusals in 90 days, zero successful grants ever**. One of them is a user who signed up on 2026-09-14, hit it, and deleted their account 5 minutes later.

The original refusal was sound for `client_secret_*`: a client that never registered has no minted secret, so honouring or downgrading both fail opaquely. That reasoning does not reach `private_key_jwt`, which needs no secret at all — the client signs an assertion with a key published at its own `jwks_uri`, bound to the vendor's host by the same argument that binds the document itself.

## What

- `Cimd` accepts `private_key_jwt` when the document publishes a usable `jwks_uri`, and refuses it as `:jwks_uri_required` when it does not. Secret-based methods still refuse.
- **New `Cimd.Jwks`** verifies the assertion. The algorithm allowlist is the load-bearing part: taking `alg` from the attacker-supplied header admits `none`, and admits HS256 forged with the RSA *public* key. Reuses the CIMD fetcher seam, so the SSRF guard, body cap, redirect refusal and content-type check are the ones already tested.
- `authenticate_client/3` dispatches on the registered method, which still binds both directions — an assertion client must present an assertion and must not present a secret.
- Discovery advertises `private_key_jwt`; clients read that list to decide what to send.
- Migration adds `jwks_uri` + `token_endpoint_auth_signing_alg`, both nullable. `phase/expand`, no backfill. The token path is a pure DB read by design, so the key location has to live on the row.

## Tests

Three things, and the second one is the point.

**The existing test asserted the bug.** `cimd_test.exs` had *"rejects a document asking for a confidential auth method"*, passing. A test that guards a decision passes just as green on the day the decision becomes wrong — which is why ChatGPT being unable to connect never turned anything red. Flipped, and the same species found in `well_known_controller_test` (*"advertises only public PKCE — no confidential auth methods"*).

**ChatGPT's real published document is now a fixture.** Every other fixture in that file is Claude-shaped; that monoculture is why a vendor declaring a different auth method was invisible.

**A vendor check against live endpoints.** `vendor_conformance_test.exs` fetches what vendors actually publish, through the real `HttpFetcher`, and asserts we accept it. Verified:

```
2 tests, 0 failures   (VENDOR_CONFORMANCE=1)
  ChatGPT (connector)  https://chatgpt.com/oauth/client.json
  ChatGPT (Codex)      https://chatgpt.com/oauth/codex/client.json
```

Gates: `156 tests, 0 failures` across the OAuth surface, `mix format` clean, `credo --strict` no issues, `dialyzer` passed.

## Deliberately not in this PR

**The vendor check is opt-in and unscheduled** (`VENDOR_CONFORMANCE=1`, excluded by default). `cron.yml` has no Elixir or Postgres job today, and `test_helper.exs:2` starts the Repo, so scheduling it means importing verify.yml's ephemeral-postgres setup for a dependency the check does not have. Scoped onto #1635.

**DCR still refuses `private_key_jwt`** — a stranger POSTing to `/oauth/register` has no host-bound document to publish keys in. That leaves discovery advertising a method one path rejects; that is #1634, and the split is encoded in the well-known test rather than papered over.

## Reviewer notes

Replay is bounded by a short `exp` rather than a `jti` store: the assertion only authenticates a token request, and the code it guards is already single-use and PKCE-bound. Reasoning is in the `Cimd.Jwks` moduledoc — worth a second opinion.

`oauth_token_controller.ex` derives `client_id` from the assertion's *unverified* `iss` when the body omits it (RFC 7521 §4.2 makes `client_id` optional beside an assertion). It only selects which client's published keys the signature is then checked against, so a wrong guess fails that check — but it is the sharpest edge in the diff.

I verified ChatGPT's *document* is accepted, which unblocks authorize. The token exchange is covered by unit tests with a synthetic keypair, not by an assertion ChatGPT actually signed. First real connect is still the first real proof.

https://claude.ai/code/session_01D2oRrcjF7tZY66ZEKdXtf2


Rasbandit:
## Review round — one critical defect found and fixed

Three independent reviews ran against the first two commits (security, silent-failure, and an integration-alerting audit). Pushed `cac1fab2` in response. The PR body above describes the pre-review shape; this is what changed.

### Critical — no JWKS cache (both reviews converged on this independently)

Every token request refetched the key set, behind a per-host bucket. That stopped being a backstop and became the **capacity ceiling**: all users of a vendor shared 10 token exchanges per minute, and the request that crossed it got a 401 that RFC 6749 §5.2 makes **terminal** — the connector stops and the user sees a permanently dead integration.

Three aggravators:
- **Refresh grants burn the same budget.** They are unattended and cluster in time, so they saturate it with no user action at all.
- **The fetch happened before signature verification**, so one IP posting junk assertions could drain the budget and starve real traffic. `/oauth/token` is unauthenticated.
- It broke the promise the migration comment makes — `jwks_uri` was copied onto the row *precisely* so the token path would not go looking.

Fixed with `Cimd.JwksCache` on the existing `Cache.NodeLocalEts` scaffolding, TTL-bounded, plus a separately-bucketed refresh so an unknown `kid` (a vendor rotating) self-heals instead of waiting out the TTL.

This one is worth dwelling on: the defect would have shipped as *"ChatGPT connects, then randomly breaks"* — the exact failure class this PR set out to end.

### Critical — our throttle reported as the vendor's fault, terminally

A rate-limit denial was laundered into `:jwks_unavailable` and surfaced as `invalid_client`. The `{:error, :rate_limited}` branch written to prevent that was **dead code** — unreachable, because `rate_limit/1` returned the wrong atom.

Transient reasons now map to a retryable **503 `temporarily_unavailable`**, the same split `cimd_error/1` already makes on the authorize path. The operator-facing half matters as much: the old log sent you to check whether OpenAI was down, when we had throttled ourselves.

### The rest: one reason per failure, and every branch logs

The reviews found the same species of bug this PR exists to eliminate — distinct failures collapsed into one atom, and rejection branches that log nothing.

- `:invalid_claims` → `wrong_issuer` / `wrong_audience` / `assertion_expired` / `assertion_not_yet_valid` / `assertion_lifetime_too_long`. That last bound is **ours**, not the RFC's, so a vendor minting a legitimate one-hour assertion must not read as a forgery. Added `nbf` and clock-skew leeway — a vendor running a few minutes fast was being refused for policy.
- `:unsupported_alg` → `alg_not_allowed` (the forgery) / `alg_pin_mismatch` (a vendor mid-rotation). An attack and a routine rotation should never share a log line.
- `:unknown_kid` → also `no_usable_key` / `ambiguous_key`. Two usable keys and no `kid` is standard mid-rotation, not an unknown key.
- The blanket `rescue` narrowed to `Signer.create/2`, returning `:unusable_key`. A vendor publishing an unparseable key never self-heals and must not be reported as a forged signature. `Joken.verify/2` never raises, so wrapping it only hid our own future bugs.
- **Six silent branches now log with `cimd_host`**, including all three in the controller that reject *before* `authenticate_client/3` is reached. The highest-probability real interop failure here — a vendor sending a whitespace-padded `client_assertion_type` — used to 401 forever and produce no evidence anywhere.
- An assertion presented to a `none`-registered client is refused rather than silently ignored. Also surfaces the post-deploy window where a row has not yet refreshed.

### Two security tightenings

- **`jwks_uri` must be same-origin as the client_id URL.** The document's host binding does not extend to keys named on another origin — that converts a binding into an unbounded delegation, where taking over the delegate host impersonates the client without ever touching the vendor. Same-origin also guarantees the URI passed the SSRF guard that fetched the document, closing a divergence where a valid-looking `jwks_uri` failed forever at token time.
- **`jwks_uri` / signing alg dropped from the DCR cast.** They were two unbounded attacker-controlled strings persisted from an anonymous public endpoint that no DCR client can ever use. CIMD sets them via `put_change`.

### Tests

The wiring had **zero** coverage — `grep client_assertion test/` returned nothing. Given this PR's own thesis, that was the conspicuous gap.

New `oauth_token_assertion_test.exs`: happy path, client_id derived from the assertion (RFC 7521 §4.2), all three rejection branches, and the 503. Plus cache-hit, rotation-refetch, clock-skew, `nbf`, unusable-key and cross-origin cases.

**171 tests, 0 failures.** `mix format`, `credo --strict`, and `dialyzer` all clean.

### Cleared by review, recorded so nobody re-derives it

The algorithm allowlist is airtight (`none` and HS* unreachable; Joken verifies via `verify_strict`, so the header cannot re-select the verifier). The unverified-`iss` fallback is safe — it only selects which row to check against, and `validate_claims/3` re-checks `iss`/`sub` against that row *after* signature verification. No cross-client confusion. The audience is not defeatable via a spoofed `Host`. The SSRF seam genuinely applies to the new caller.

### Still open, deliberately

The vendor conformance check remains opt-in and unscheduled (#1635). And the alerting gap that let this hide for a week is now filed as engram-app/engram-infra#1157 — our rules key on failure *volume*, so a fully-broken low-traffic integration is the quietest thing in the system.

PR #1664, opened 2026-09-16
docs(context): MCP-first signup dead-ends on onboarding_required

Rasbandit:
## What

New context doc: `docs/context/mcp-first-signup-onboarding-deadend.md`, plus two trigger entries in the `AGENTS.md` (`CLAUDE.md`) index under **Auth, OAuth & MCP**.

## Why

Verified 2026-09-16 against prod RDS (engram-audit read-only bastion) and Grafana Loki. A user who reaches Engram **through an MCP client's OAuth flow** — never through app.engram.page — signs up inside the grant, receives valid refresh tokens, and then every `/api/mcp` tool call returns `403 onboarding_required`. Forever.

`/api/mcp` pipes `:authed_api`, which includes `RequireOnboarding` (`router.ex:55`) — correctly. The missing half is that nothing in the OAuth path (`oauth_authorize_controller.ex`, `oauth_token_controller.ex`, `oauth_register_controller.ex`) runs or links onboarding, and the MCP client renders the 403 as an opaque tool failure. `DeviceAuthController` already carries the precedent fix for the plugin's entry point (`device_auth_controller.ex:21`, `skip_vault: true`).

Prod case 2026-09-15: DCR registers "Google Antigravity" → user row created → grant succeeds → first 4 tool calls 403 → 12 more DCR registrations and 41 `no_auth` 401s over ~6h → user gave up. Final state: 0 vaults, 0 notes, 0 agreements, 0 subscriptions, live refresh tokens. A second account (2026-09-11) shows the same shape.

Doc is a record of the defect class + the investigation gotchas; **no code change here**, the fix is still open.

## Investigation gotchas recorded

- Prod Loki ships **warning-and-above only** for `{env="prod", role="web"}` — successful 2xx traffic is invisible; never conclude "the user did nothing" from log absence.
- `request_path` is `[REDACTED]` and `route` was null — correlate by `user_id` + DB timestamps.
- `engram_audit_ro` is RLS-bound: a correlated subquery over `users` returns **0 for every row, silently**. Query one tenant at a time.

https://claude.ai/code/session_01UGn1VTY4YQjg3xS5Xy98LS


Rasbandit:
Folded into #1670 so the fix, its tests, and this doc land together — the doc describes the defect that PR closes, and reviewing them apart means reading the same investigation twice. Commit 0af924e7 is cherry-picked onto `fix/mcp-onboarding-403-resume-url` unchanged.

Issue #1668, opened 2026-09-16
Merge the wizard vault step with the OAuth vault picker for MCP-first signups


Rasbandit:
## Idea

The MCP-first signup flow currently shows two screens that both create a vault:

- `/onboard/vault` (`onboard-vault-page.tsx`) — picks the source (Obsidian vs fresh) and, on the fresh path, creates the first vault.
- `/oauth/consent` — picks which vaults the connecting client may access, and for a brand-new account that list is empty.

For someone arriving through an MCP connector these are the same decision asked twice. Merge them: drop the Obsidian source pick from this path, let the consent screen create the first vault, and move the Obsidian story to its own "how to connect Obsidian" step.

## Constraints found while implementing #1666

Anyone picking this up should know these before designing it.

**`uses_obsidian` is load-bearing, not a preference.** `Engram.Onboarding.derive_gate/3` computes `vault_required = profile_ok and profile["uses_obsidian"] != true`, and `set_profile/2` only stamps `completed_at` once **both** `tools` and `uses_obsidian` are present. A merged screen that stops asking must still **write** `uses_obsidian: false`, or `profile_complete?` stays false, the gate keeps failing on `profile`, and the dead end this epic closed reopens one layer over.

**Defaulting it to false is right for this path and wrong as a general rule.** Obsidian-first users arrive via `/link`, which carries its own `RequireOnboarding, [skip_vault: true]` relaxation (`device_auth_controller.ex:21`) precisely so the plugin's first sync can create the vault. If a merged screen forced a vault on them, first sync would land beside an empty vault they never asked for — the duplicate-vault class already documented in `folder-rename-mint-resurrection.md` and `crdt-create-cross-vault-id-remint`.

**An added step changes the chain for everyone.** `build_steps/2` feeds the `Step X of N` counter in `OnboardLayout`. A "learn about Obsidian" step makes every signup, including self-host, count one higher. It also needs a completion condition, or `next_step` can never advance past it.

**The consent screen already has vault-creation adjacency.** It renders the vault picker with an "All vaults" option that stays all-vaults as new ones are created, so a create affordance there has to decide whether a vault made during consent lands inside an explicit selection or the implicit all-vaults grant.

## Scope

Not a follow-up to bolt onto #1666 — it changes the wizard's shape for every user, not just the MCP-first path. Wants its own design pass.

Refs #1666

Issue #1204, opened 2026-08-02
p1: a registered-but-unused vendor redirect grants the verified badge (impersonation)


Rasbandit:
`LogoAllowlist.resolve/4` grants `verified: true` if **any** entry in a client's registered `redirect_uris` matches a vendor host. It does not check which redirect the grant actually used. Since `/oauth/register` is public DCR, anyone can put a vendor host in their list, authorize with a different one, and wear the vendor's badge and logo.

## Reproduction (run against staging 0.13.0)

```elixir
LogoAllowlist.resolve(nil,
  ["http://localhost:9999/steal", "https://claude.ai/api/mcp/auth_callback"],
  "Totally Legit App")
```

```
verified=true  name="Claude"  logo="/assets/clients/claude.svg"
```

The attacker authorizes with `redirect_uri=http://localhost:9999/steal`, so the code is delivered to their machine, while the victim's connections list shows **Claude, verified, with Anthropic's logo**.

## Why the existing argument does not cover this

The module's own rationale is:

> A forger can claim `redirect_uri=https://claude.ai/...`, but the auth code is then delivered to Anthropic, not to them.

That holds only when the grant *uses* that URI. With a multi-entry list the forger claims the vendor host for the badge and uses a loopback for delivery. The two are decoupled.

This is the same class as the `software_id` tightening (#1156): identity that reads as *proven* while actually being *claimed*. That one was closed by deleting guessed entries; this path stayed open because the host looked like proof.

## Cause

`lookup_by_host/1` is `Enum.find_value/3` over the whole registered list — first vendor match wins, no relationship to the grant:

```elixir
defp lookup_by_host(uris) when is_list(uris) do
  Enum.find_value(uris, @empty, fn uri -> ... end)
end
```

## Two fixes

**A. Interim, small: require EVERY redirect to be a vendor-owned HTTPS host.** If a client can receive codes anywhere other than the vendor's host, that host proves nothing about delivery. One function plus tests, no migration.

- attacker `[localhost, claude.ai]` → contains a non-vendor URI → **unverified**
- Cursor `[cursor://…, www.cursor.com, localhost:8787]` → **unverified** (already is, and correctly so: its desktop grant lands on loopback)
- single-redirect vendors (Claude, ChatGPT, Grok, Mistral, Antigravity per the observed table in `connections-client-identity.md`) → unchanged

[...]

PR #1670, opened 2026-09-16
fix: let an MCP-first signup finish onboarding and resume consent

### LINKED ISSUE #1666: p1: MCP-first signup dead-ends — OAuth succeeds, every tool call 403s onboarding_required forever
## Summary

A user who reaches Engram through an MCP client's OAuth flow, and never through `app.engram.page`, ends up with an account that can do nothing. OAuth succeeds, tokens are issued, and then every MCP tool call returns `403 onboarding_required` forever. Nothing in the OAuth path runs or links to the onboarding wizard, so the user has no path forward and no way to know what is wrong.

This is a signup funnel leak with a legal component: these accounts exist with **no terms acceptance recorded at all**.

## Evidence (prod, verified against RDS + Loki)

`dgonzalez.integraenergia@gmail.com`, internal `01a0a550-4d17-795b-88ea-92d5e4d8e62a`, Clerk `user_3JMne9wsT1Pxy6hHMpo3FCQAbyr`.

| UTC 2026-09-15 | Event |
|---|---|
| 13:44:42 | DCR registers client `052a941b-…` name `Google Antigravity` |
| 13:44:58 | User row created (signup happened inside the OAuth grant screen) |
| 13:45:18 | Grant succeeds: 2 `oauth_refresh_tokens`, scope `mcp`, exp 2026-12-14 |
| 13:45:18 | First 4 MCP tool calls → `POST 403`, `onboarding_required`, `missing: ["profile","subscription","terms","vault"]` |
| 13:49:55 – 19:25:32 | 12 further DCR registrations, 41 `auth rejected reason=no_auth` 401s, zero new tokens |

Final DB state: 0 vaults, 0 notes, 0 chunks, 0 attachments, 0 `crdt_update_log`, 0 `api_keys`, 0 `client_logs`, 0 `user_agreements`, 0 `onboarding_actions`, 0 `subscriptions`, `plan_id` NULL, `free_tier_accepted_at` NULL, `onboarding_profile` `{}`. Account never deleted, never suspended.

Not a one-off. `aortizsm@gmail.com` (2026-09-11) shows the same zero-everything shape. Contrast a healthy account, `sabio@web.de`: 2 agreements, 1 vault, 100 notes.

## Root cause

`/api/mcp` pipes through the shared `:authed_api` pipeline, which carries `EngramWeb.Plugs.RequireOnboarding` (`lib/engram_web/router.ex:55`). The gate is correct; the problem is that no OAuth controller (`authorize`, `token`, `register`) references onboarding at all, so an MCP-first user can never satisfy it without independently discovering the web app.

`device_auth_controller.ex:21` is the only route-level relaxation that exists today (`RequireOnboarding, [skip_vault: true]`). The MCP grant path has no equivalent.

## Proposed fix

Split the four gate conditions by what they actually are.

**`terms` is legally load-bearing.** Capture it on the OAuth consent screen for users who have not accepted, as clickwrap: terms and privacy visible or clearly linked, acceptance recorded on the same affirmative click that grants the scope. Record version hash, timestamp, IP, and user agent into `user_agreements`. `terms_versions` already stores a content hash, which is the right evidence model.

**`subscription`, `profile`, and `vault` are product state, not law.** Auto-provision Free tier and a default vault at grant time. Gating MCP on them yields nothing and costs the user.

**Backstop:** make the 403 body actionable. `next_step` is already returned; include an absolute URL the MCP client can surface so a user who still lands here has somewhere to go.

## Out of scope

No change to the gate itself, and no relaxation on the WebSocket path. `Engram.Onboarding.gate/2` stays authoritative for both transports.

## Notes

- Full write-up: `docs/context/mcp-first-signup-onboarding-deadend.md` (#1664).
- Related: #364 (wizard-side half).
- The observed `missing` list is version-dependent. `Onboarding.derive_gate/3` computes `vault_required = profile_ok and profile["uses_obsidian"] != true`, so the list shape varies; treat the recorded value as observed, not stable.

Rasbandit:
## Summary

- A user who reaches Engram through an MCP client's OAuth flow completed the grant, got valid tokens, and then hit `403 onboarding_required` on every tool call forever. `/api/mcp` pipes through the shared `:authed_api` pipeline which carries `RequireOnboarding` (`router.ex:55`), and no OAuth controller referenced onboarding, so there was no path forward and nothing in the response said so. Verified in prod: two accounts sit at zero vaults, zero notes, zero agreements.
- The consent page now parks the in-flight authorization, sends the user through the existing wizard, and returns them to finish approving. Terms capture stays in the wizard rather than being forked onto a second screen, so the binding-grade path (version + content hash + IP + user agent) is untouched. The FTUX tool question is pre-answered from the connecting client instead of asked, since arriving through an MCP connector has already answered it.
- Refusals on `/api/mcp` are now JSON-RPC errors carrying a readable sentence. `McpController` always shaped its own failures that way, but the seven halting plugs in `:authed_api` never reach it, so their flat REST bodies went out as-is and MCP clients showed a bare "HTTP 403" with the body dropped. Status codes are unchanged.

Interrupting the authorization leg for signup is expected behavior under OAuth 2.1, provided the request parameters stay associated, the user stays in one origin, the context is preserved, and cancelling returns a standards-compliant error. All four hold here: the full consent URL including `state` and the PKCE challenge survives the detour, the wizard names the app that is waiting, and Cancel reports `access_denied` to the original redirect.

### Notable decisions

- **`McpErrorEnvelope` installs on the MCP scope, not in `Plugs.Halt`.** `Halt.json/3` is shared by ~20 call sites across six pipelines and its moduledoc states that extras stay at the call site. Same placement and `register_before_send` mechanism as the existing `McpAuthChallenge`.
- **401 is deliberately not reshaped.** It is the OAuth discovery entry point (RFC 9728 §5.1) and its bare body plus `WWW-Authenticate` is what a spec-following client acts on.
- **`build_steps/2` drops `:tools` once answered.** It feeds the "Step X of N" header, and a pre-answered step that never renders must not be counted; without this the MCP-first path read "Step 2 of 4" then jumped to "Step 4 of 4".
- **The client slug resolves from the redirect the grant is using**, and a `redirect_uri` the client never registered is ignored rather than trusted. Scanning a registered list for any vendor host is what #1204 was.

### Tests added

The un-onboarded state this fix is about was **unreachable in either e2e suite**: `frontend/e2e/global-setup.ts`, `e2e/helpers/oauth.py` and `e2e/helpers/clerk_auth.py` all pre-complete onboarding before the first assertion. That is structural, and it is why the harness could never catch this. Two tests close that hole:

- `e2e/tests/api_only/test_89_mcp_onboarding_envelope.py` — proves `McpErrorEnvelope` is actually installed on a booted server's MCP scope, in the right order relative to the plugs it wraps, and that REST refusals keep their flat shape. Unit tests can prove the plug's logic but never its wiring.
- `frontend/e2e/oauth-consent-onboarding.spec.ts` — the full round trip in a real browser: consent → bounce → terms → plan → vault → back to consent with `state` intact → approve → `code=`. `global-setup.ts` now provisions a second user and deliberately skips onboarding for it. This is also the only place the SPA meets the real backend on this path; every unit test mocks `../api/oauth`, so nothing else proves `/api/oauth/clients/:id?redirect_uri=` returns `slug` in the shape the consent page reads.

`docs/context/mcp-first-signup-onboarding-deadend.md` is included here too (was #1664, closed) so the defect write-up lands with the fix.

### Three pre-existing tests were modified

Each preserves what the test was written to pin:

- `onboarding_test.exs:155` and `:216` — invariance fixtures carried an incidental `tools: ["claude"]` while their names pin `uses_obsidian`. Narrowed to the field they name, with new cases added for the drop behaviour.
- `onboarding_gate_integration_test.exs:112` — asserts the MCP gate halts. The status assertion is unchanged; the reason moved one level down into the JSON-RPC envelope.

## Test plan

- [x] Backend full suite: 5360 tests, 0 failures
- [x] `mix credo` clean on every touched module; `sobelow` clean
- [x] `mix dialyzer` passes
- [x] Frontend: 192 files, 1828 tests; `tsc --noEmit` clean
- [x] `biome check --error-on-warnings` clean (run with the pinned 2.5.12 via `bunx`, see below)
- [ ] `test_89` and `oauth-consent-onboarding.spec.ts` have **not** been run locally — both need the CI stack plus a Clerk secret. CI is their first real execution.
- [ ] Manual smoke still worth doing before merge: connect an actual MCP client against a fresh account.

Note for anyone running frontend lint locally: `package.json` pins `@biomejs/biome` 2.5.12, but a worktree's hardlinked `node_modules` can carry 2.5.11, which aborts with a config error on every file including untouched ones. `bunx @biomejs/biome@2.5.12 check ...` reproduces CI without mutating the shared install.

Reviewer notes: the consent page stays outside `OnboardingGate` on purpose. That gate redirects to `/onboard/<step>` and would discard the authorization request; bouncing from inside the page is what lets the request survive.

Follow-up filed as #1668 (merging the wizard vault step with the OAuth vault picker), deliberately not in this PR because it changes the wizard's shape for every user.

Closes #1666

https://claude.ai/code/session_01UGn1VTY4YQjg3xS5Xy98LS


Rasbandit:
Alerting half is open as a companion infra PR: engram-app/engram-infra#1184.

It adds one `loki_category_alerts` entry keyed on the `onboarding refused` line that commit 2bfbcf8e in this PR emits.

**Merge this PR first.** The rule watches for a log line that no deployed build emits until this ships — it won't error, it just never fires.

Threshold there is 0, picked from measurement: prod holds only twelve `status:403` lines total across seven days, and this incident was four of them. Any volume-keyed threshold above three would have missed the thing it exists to catch.

Rasbandit:
## Reviewed, and it found two blockers

Ran a four-dimension adversarial review over this branch (correctness, silent failures, test quality, comment accuracy). ~20 findings. All code defects are fixed in 0a4d9afb.

**Both blockers were fresh dead ends introduced by this PR**, of the same class it exists to remove:

1. **`uses_obsidian` users bounced forever.** `derive_gate/3` admits them while `next_step/5` pins them at `:vault` until the plugin first-syncs. The consent page gated on `next_step !== "done"`, so wizard and consent threw those users back and forth with nothing left to collect and nothing left to refuse. `status/1` now publishes `gate_ok` from the same `gate_missing/1` that `gate/2` enforces with.
2. **Cross-account authorization hijack.** The sessionStorage stash survived sign-out, so A parks a request, signs out from the wizard header, B signs up in that tab, finishes the wizard, and lands on A's consent with A's `state` and `redirect_uri`. Approving minted a grant on B's account and shipped the code to A's redirect.

Also fixed: the step counter froze for every ordinary signup (`build_steps/2` keyed on live profile state, so answering the question deleted the step from your own chain); 5xx read as "The request was refused."; Cancel was dead for `cursor://`-style clients and destroyed the parked request anyway; the consent copy promised a return trip the stash cannot guarantee; the tool pre-answer failure was silent; `clerkSignIn` could return without attempting a sign-in.

Four comments claimed more than the code supports. The dangerous one: `logo_allowlist.ex` asserted `devin` is NOT in `valid_tools/0`. It is (`onboarding.ex:41`). Acting on it would have 422'd every Devin pre-answer into a then-silent catch.

### Gates

Backend `mix test`: 5375 tests, 1 failure, `Engram.Search.MMRTest`. **Not from this branch** — it touches nothing under `lib/engram/search/`, passes alone in 4.3s, and asserts a 5s `Task.yield` wall-clock budget that loses under the full parallel suite. Filed separately.

Frontend: tsc, biome and 1830/1830 unit tests clean.

### Still open

Test coverage, filed separately rather than bundled: this PR's headline e2e proves an authorization **code** was minted, which always worked. It does not prove the resulting token can call a tool, which is the thing that was broken.

https://claude.ai/code/session_01UGn1VTY4YQjg3xS5Xy98LS

PR #1719, opened 2026-09-18
fix(rls): let API-key auth read its own credential row

Rasbandit:
Every API-key request returned **401 `invalid_key` for keys that exist**, locking out plugin sync, MCP clients and scripts. Found on staging while verifying the merged RLS work.

`validate_api_key/1` resolves a `key_hash` to its owner, so the `user_id` is that lookup's **output** and no tenant can be set beforehand. `tenant_isolation_api_keys` filtered the row to nothing.

Latent everywhere until staging's app pool dropped from the superuser to `engram_app`. **Prod connects as its BYPASSRLS migrator, so prod was never affected** and would have hit this at cutover.

## The fix

```sql
CREATE POLICY api_keys_discovery ON api_keys FOR SELECT
  USING (coalesce((SELECT current_setting('app.current_tenant', true)), '') = '');
```

Permissive policies OR within a command and AND across commands, so a `FOR SELECT` policy widens **reads only**. INSERT keeps `tenant_isolation_api_keys`'s `WITH CHECK`; UPDATE and DELETE keep its `USING`.

Nothing else moves: `api_keys` stays in `@tenant_tables` so the `prepare_query/3` tripwire and four derived lints survive, `structure.sql` needs no edit, and the call site keeps its existing `Repo.cross_tenant/1` wrapper.

## Why "no tenant set" and not `USING (true)`

Both fix auth. `true` also lets tenant-scoped code read every user's keys. Measured as `engram_app`, rows visible:

| policy | no tenant (auth path) | inside `with_tenant` |
|---|---|---|
| no RLS at all | all | all |
| `USING (true)` | all | all |
| `USING (tenant unset)` | all | **own only** |

With a tenant set, a foreign INSERT still raises 42501 and cross-tenant UPDATE/DELETE still report 0 rows. All four verbs verified on a scratch database.

## History

This PR previously dropped `api_keys` from the tenant policy set entirely. Adversarial review killed that approach: it discarded the INSERT `WITH CHECK`, the UPDATE/DELETE `USING`, the tripwire and four lints to fix a single filtered SELECT, and "compensated" by revoking UPDATE, a verb with zero callers. It also broke six CI jobs by revoking before the table existed. Rebuilt from main as one migration.

## Tests

Two cases. The first pins that auth works; the second pins the predicate's *shape* and is the one that fails under `USING (true)`, so a later widening can't silently surrender read isolation. Both carry a control proving the role drop engaged.

Context doc: `docs/context/rls-cutover-breaks-api-key-auth.md` (updated here to describe what shipped).


Rasbandit:
## Code review round — two critical findings, both real

A review pass caught a self-inflicted CI breaker that my local suite **structurally could not detect**, plus a second bug that would have silently voided the security control this PR rests on. Both are fixed in `5a496ccd`. Recording the reasoning since the first commit asserted the opposite.

### 1. The revoke ran before the table existed

`prepare_database/0` runs **before** `migrate/0` — `mix.exs:222-224` documents that ordering deliberately, so the baseline dump's GRANTs resolve. On an empty database `api_keys` does not exist yet, so `REVOKE UPDATE` raised:

```
** (Postgrex.Error) ERROR 42P01 (undefined_table) relation "api_keys" does not exist
    (engram 0.28.0) lib/engram/release.ex:296: Engram.Release.do_prepare_database/1
```

Six jobs died on it — every one that builds a fresh database: `unit-tests`, `migration gates`, `storage-database`, `e2e-crdt`, `e2e-browser`, `headless-protocol`. My local run passed because that database already had the table, so no amount of local testing would have found this.

### 2. Even without the crash, the control did not exist

`structure.sql:1695` re-GRANTs `UPDATE` during the migrate that follows `prepare_database`. So on any rebuilt database the sequence was revoke-then-grant, and `engram_app` kept UPDATE. The compensating control existed only on long-lived prod, which is exactly where the original bug never bit.

### The fix

Both are the same ordering mistake. The REVOKE moves into the migration, which runs after the table exists and after the baseline GRANT. `ALTER DEFAULT PRIVILEGES` only grants on *future* tables, so nothing re-grants it afterward. Guarded on the role existing, since self-host connects as its own owner and may have no `engram_app`.

Verified on a genuinely empty database, the same path CI takes:

```
CREATE_RC=0  PREPARE_RC=0  MIGRATE_RC=0
update=false (want false) | select=true insert=true delete=true
api_keys rls=false force=false policies=0
```

### Also corrected

- `skip-tenant-check-audit.md`: bucket-C count 6 → 5, and the `accounts.ex` api_keys entry removed (that site is bucket A now — the table is no longer tenant-scoped).
- Eleven `structure.sql` line anchors in `database-schema-rls.md` that my hand-edit shifted by two, plus the RLS block range.

### Reviewed clean

`structure.sql` fidelity (my least-confident edit) was checked against pg_dump's conventions in neighbouring tables and is byte-faithful. Removing `api_keys` from `@tenant_tables` loses no guard: all four query sites already filter `user_id` independently. `REVOKE UPDATE` is safe — every `api_keys` reference in `lib/` is a read, and `ApiKey.changeset/2` has exactly one caller, the insert.

One forward-looking note: this cements `last_used` as a dead column. Anyone implementing stamping later gets a runtime `42501`, not a compile error, and must lift the revoke first.

Rasbandit:
## splinter waiver

CI's schema linter flagged the design directly:

```
[WARN] multiple_permissive_policies — Table `public.api_keys` has multiple
permissive policies for role `engram_app` for action `SELECT`.
Policies include {api_keys_discovery, tenant_isolation_api_keys}
```

That is the shape by construction: a second permissive SELECT policy is the fix. The advisory is about per-query policy evaluation cost, and this table is read once per request by unique index.

Collapsing the two policies into one would satisfy the linter but **widen DELETE** — DELETE would then consult the permissive predicate instead of the strict one. So `lint_schema.sh` gets a waiver instead, matching lint **and** object rather than disabling `multiple_permissive_policies` schema-wide.

Verified in both directions, reproducing the linter locally:

| case | result |
|---|---|
| real schema | pass |
| same lint, different table (`public.notes`) | **still fails** |
| different lint, same table (`auth_rls_initplan`) | **still fails** |
| the waived pair | pass |
| empty input | pass (no `set -e` abort) |

The last row matters: `grep -v` exits 1 when it filters every line, which under `set -euo pipefail` would otherwise abort the gate. It's `|| true` with a comment saying why.

PR #1723, opened 2026-09-18
fix(rls): make the vault purge fail loudly

Rasbandit:
Follow-up to #1722. Adversarial review of that merged commit found it left the **same silent-failure class one line below the part it fixed**.

## 1. The write was never checked

`CleanupVault.enqueue_now/2` ends in `Oban.insert/1`, which returns `{:error, changeset}` rather than raising. `Enum.each/2` discarded that and returned `:ok` unconditionally, and `Admin.UserController.delete/2` ignores the return and answers `{"ok": true}`.

So a rejected insert reproduced the exact symptom #1722's own commit message described: *the endpoint reports success while nothing is reaped, nothing raises, nothing logs.* #1722 fixed the read and left the write unchecked.

## 2. The "outside the transaction" rationale was backwards

#1722 moved the enqueue out, arguing Oban rows tied to the transaction "would be discarded on a rollback". But nothing in the block calls `Repo.rollback/1`, so the only rollback path is a raise — where discarding the jobs is the **correct atomic outcome**, not a hazard to design around.

What it actually bought was a partial-failure window: vault 3 of 5 raises, vaults 1–2 have jobs, 3–5 never will, nothing retries, and the admin gets a 500 with no record of which half succeeded. The enqueue moves back inside.

## Also

- **Logging.** An irreversible admin action whose entire bug history is "it silently did nothing" emitted nothing at all. Now logs the count under `:lifecycle`.
- **Docstring.** Gains the precondition (raises if called inside another tenant's scope) and the qualifier #1722 lacked: this was only ever live where the app connects as a role the policy applies to — staging and self-host — never prod on its BYPASSRLS migrator.
- **Test.** Seeded one vault, so the `Enum.each` loop carrying the risk was never exercised, and an over-broad enumeration that dropped the `user_id` predicate would still have counted 1 and passed green. Now seeds two vaults plus a bystander user and asserts `{2, 0}`, pinning both failure directions.
- **Trap 2 recorded** in the test moduledoc: `with_tenant/2` resets the role on exit but not the tenant, so anything appended after the purge call inside that closure runs with a tenant in force and would false-green.

## What review cleared

The two things I thought most likely to be wrong were not: the nested-tenant raise does **not** fire (the admin path holds no tenant at that point — traced through the pipeline, plugs, and `soft_delete_user/1`), and the ratchet 36→35 is correct rather than a lucky accident (the counter strips heredocs, so the one prose mention is properly excluded).

`53 tests, 0 failures`; format + credo clean.


Rasbandit:
## Second adversarial pass (on `ca2001db`)

Ran a review on this PR itself, not just on what it fixes — a prescription can't catch the prescription being implemented wrong. It found two broken things, both mine.

### 1. The bystander assertion pinned nothing

The enqueue passes the **outer** `user_id`, not `v.user_id`:

```elixir
Enum.each(vaults, fn v -> CleanupVault.enqueue_now(v.id, user_id) end)
```

So every job carries the purged user's id regardless of what the `vaults` read returns, and `count_for.(other.id)` is structurally unable to be non-zero. Dropping the `user_id` predicate — the exact regression the assertion claimed to catch — yields `{3, 0}`, not `{2, 1}`. The test still failed, but through the first element, with a message sending the reader after a cross-tenant leak the assertion couldn't detect.

Now keyed on the bystander's **vault id**, which does move. Counts are also worker-scoped so an unrelated job on the same user can't inflate them.

### 2. The success log was emitted before COMMIT

Moving the enqueue inside the transaction — this PR's headline — also moved the log inside it. It asserted "enqueued N" while those rows were uncommitted; a drop or timeout before COMMIT leaves a log claiming jobs that don't exist. That's the report-success-while-doing-nothing shape this function exists to remove, one layer up. Now logged after `with_tenant!` returns, which is free since it already returns the count.

### 3. The prod exculpation is withdrawn

I wrote that prod connects as a BYPASSRLS migrator and was never affected. **Three in-repo sources say otherwise:** `Engram.Onboarding` records `engram_admin` as `rolbypassrls=false`; `docs/context/migrations-force-rls-data-dml.md` says the prod migrator has neither SUPERUSER nor BYPASSRLS; `TenancyGuard`'s moduledoc says prod connects as a superuser.

If the first two are right, admin-deleted users' vaults were never reaped **in prod either** — a data-retention question, not a footnote. I tried to settle it from `TenancyGuard`'s boot log (`:boot` ships at `:info` per `Logger.Category`), but nothing matched in Loki over 30 days, so either prod's logs aren't in the stream I queried or `:boot` is dropped by Fluent Bit routing — `category.ex:46-49` warns the Elixir list and the routing regex are not the same set.

The docstring now states the question as **unverified** and names the contradiction, rather than repeating a reassurance I never checked. Worth chasing separately.

### Cleared on review

The raise does roll back (Ecto catches and re-raises); moving Oban inserts inside the role-dropped transaction is safe (`engram_app` holds the grants, `oban_jobs` carries no policy, `pg_notify` is transactional so no pre-commit race); `:lifecycle` is valid and ships; raw `user_id` in log metadata is the convention here, not a leak; the return-type change has no dependent caller or spec; and the moduledoc's trap-2 claim is accurate on both halves.

`53 tests, 0 failures`; format + credo clean.

Issue #1649, opened 2026-09-16
RLS is not enforced for the app's DB role — tenant isolation currently rests on app-layer filters only


Rasbandit:
## Summary

Row Level Security provides **no actual isolation** on the application's database connection in prod. The app connects as `engram_admin`, and that role is not filtered by the `tenant_isolation_*` policies — it can read every tenant's rows regardless of `app.current_tenant`.

Found while investigating whether #1632's failure mode affected the app path. It does not — but for a worse reason than expected.

## Evidence (prod, read-only, as `engram_admin`)

```
current_user                     = engram_admin
rolsuper                         = false
rolbypassrls                     = false
notes.relrowsecurity             = true
notes.relforcerowsecurity        = true
notes owner                      = engram_admin
current_setting('app.current_tenant') = <user-uuid>     (confirmed set)

select count(*) from notes                          -> 3602   (ALL rows)
select count(*) from notes where user_id = <uuid>   -> 1
```

The tenant was verifiably set and the policy still did not apply.

## Controlled comparison

`engram_audit_ro` — same database, same tables, **not** a member of `pg_read_all_data`:

```
select count(*) from notes   (no tenant set)   -> 0      (correctly filtered)
```

So the policies work. They just don't apply to the app's role.

## Likely cause

`engram_admin`'s role memberships include:

```
pg_read_all_data, pg_write_all_data, pg_maintain, pg_monitor,
pg_read_all_settings, pg_read_all_stats, pg_signal_backend,
pg_stat_scan_tables, rds_superuser, rds_password, rds_replication, ...
```

`pg_read_all_data` / `pg_write_all_data` are the obvious suspects, and they are the difference between the two roles tested.

**Caveat:** the PostgreSQL docs state `pg_read_all_data` does *not* set `BYPASSRLS`, so the exact mechanism needs confirming rather than assuming. Reproduce on staging before changing prod.

## Impact

Tenant isolation today comes from two things, neither of which is Postgres:

1. explicit `where user_id == ^user_id` predicates in queries
2. the app-level `Repo.prepare_query/3` guard

Both are real and currently correct, but RLS is documented throughout the codebase as the enforcing layer (`docs/context/database-schema-rls.md`: "App connects as engram_app (subject to RLS)"). The wiring does not match: `ecs_secrets.tf` builds `DATABASE_URL` from `aws_db_instance.main.username`, which is the RDS master.

Two consequences:

[...]

Issue #1357, opened 2026-08-11
Audit: request-path tenant-table reads outside with_tenant (67 sites)


Rasbandit:
Follow-up to #1354, which confirmed against prod that any tenant-table query outside `Repo.with_tenant/2` returns zero rows (or raises on a WITH CHECK).

#1354 fixes the three sites on the onboarding request path. This issue is the systematic sweep, because that path was found by accident — it happened to be adjacent to the #1349 backfill work — not by looking.

## Scope

`skip_tenant_check: true` appears ~280 times across `lib/`. Filtering to occurrences whose surrounding query references a tenant-backed schema (`Note`, `Attachment`, `Vault`, `ApiKey`, `UserAgreement`, `CrdtUpdateLog`, `NoteLink`, `Chunk`, onboarding `Action`) gives **67 sites across 25 files**:

```
 15  engram/links.ex                     3  engram/workers/embed_note.ex
 11  engram/crypto/user_dek_rotation.ex  2  engram/accounts/export/streamer.ex
  5  engram/crypto/aad_rebind.ex         2  engram/accounts.ex
  3  engram/indexing.ex                  2  engram/onboarding.ex
  3  engram/links/rewriter.ex            2  engram/workers/cleanup_vault.ex
  3  engram/vaults.ex                    2  engram/workers/extract_note_links.ex
  (+ 13 files with 1 each)
```

**Most are almost certainly fine.** Oban workers typically wrap their whole run in `with_tenant` already, and `skip_tenant_check` inside a tenant context is merely redundant, not broken. The dangerous subset is the sites reached from a controller or channel *without* an enclosing `with_tenant` — which needs call-graph work to determine, not grep.

## Why the existing lints do not cover this

- `tenant_enumeration_lint_test.exs` (#1355) deliberately targets only the *enumerate-every-tenant* shape (`group_by`/`select` on `user_id`). A single-user read with a correct `where user_id == ^id` predicate is not flagged — and `Vaults.count_for/1` proved that shape is broken too, because RLS filters before the app predicate is ever relevant.
- `raw_sql_tenant_table_lint_test.exs` covers raw SQL only.

A lint for this class would need to know whether a caller is inside `with_tenant`, which grep cannot answer. Options worth weighing: a runtime assertion in `prepare_query/3` that raises (rather than logs) on a tenant-table query with no tenant in non-prod, or a `telemetry` alert on `tenant_check_skipped` from a web process.

## Suggested approach

[...]

Issue #1354, opened 2026-08-10
Runtime onboarding_actions reads/writes are unscoped the same way the backfill was (#1349)


Rasbandit:
Surfaced by the review of the #1349 fix. The backfill is fixed; the **runtime** paths against the same FORCE-RLS table are not.

| function | shape |
|---|---|
| `Engram.Onboarding.list_actions/1` (onboarding.ex:441) | `Repo.all(query, skip_tenant_check: true)`, no `with_tenant` |
| `Engram.Onboarding.record_action/2` (~line 425) | `Repo.insert(..., skip_tenant_check: true)`, no `with_tenant` |

`list_actions/1` is called from `OnboardingController.status_payload/1` and there is no plug wrapping the request in `Repo.with_tenant/2` (grep for `with_tenant` under `lib/engram_web/` returns only sync_controller, embed_status_controller and a crdt_channel comment).

## Why this matters

Under #1349’s premise — prod `DATABASE_URL` is `engram_admin`, the RDS master, which per `docs/context/migrations-force-rls-data-dml.md` has neither SUPERUSER nor BYPASSRLS, and `onboarding_actions` carries FORCE ROW LEVEL SECURITY with `USING (user_id::text = current_setting(app.current_tenant, true))` — these should return zero rows / raise on prod today.

So one of two things is true, and I could not determine which without a prod query:

1. **They are broken today.** `list_actions/1` returns `[]` for every user, so the wizard shows no completed actions. Plausibly unnoticed: `RequireOnboarding` gates on TOS + subscription (non-tenant tables), not on these rows, so the gate still works and only the wizard display is wrong. `record_action/2` would additionally raise on the policy’s `WITH CHECK`.
2. **The premise is incomplete** — some path grants the app role an RLS bypass that the doc does not mention — in which case #1349’s fix was unnecessary (though not harmful: scoping the discovery is correct either way) and the doc needs correcting.

**Resolving this is the actual task**, and it needs one read-only prod query, not more code:

```sql
SELECT rolname, rolsuper, rolbypassrls FROM pg_roles WHERE rolname = current_user;
SELECT relname, relrowsecurity, relforcerowsecurity FROM pg_class WHERE relname = onboarding_actions;
```

[...]

Issue #790, opened 2026-06-29
Verify prod app DB role does not have rolbypassrls (RLS fails open by default otherwise)


Rasbandit:
## Summary

In the **dev** DB, the app's default connection role is a superuser with `rolbypassrls`:

```
 rolname    | rolsuper | rolbypassrls
 engram     | t        | t      ← app's default connection role
 engram_app | f        | f      ← role dropped to inside with_tenant
```

RLS only enforces after `Repo.with_tenant/2` drops to `engram_app` (`lib/engram/repo.ex:74-78`). Outside `with_tenant`, the connection runs as `engram`, which **bypasses every RLS policy**. This means RLS is the *primary* isolation guarantee only inside `with_tenant`; everywhere else it is inert and isolation rests entirely on hand-written `where user_id == ...`.

This inverts the safe default. Best practice is for the app's normal identity to be a non-superuser whose RLS is always on, so a forgotten `with_tenant` / missing `where user_id` fails **closed** (empty/denied) rather than open.

## The question to answer

Does the **prod** (RDS) app pool role carry `rolbypassrls`? Run against prod:

```sql
SELECT rolname, rolsuper, rolbypassrls FROM pg_roles WHERE rolname = current_user;
```

Note: the RDS master user is `rds_superuser`, which **does** have `rolbypassrls`. If the prod app connects as that (or any bypass role), prod has the same fail-open posture as dev.

## Possible outcomes

- **Prod app role has `rolbypassrls`** → RLS is load-bearing only inside `with_tenant`; consider connecting the app pool as a dedicated non-bypass role so RLS is always enforced (defense-in-depth, fail-closed). The `with_tenant` role-drop becomes redundant-but-harmless.
- **Prod app role is non-bypass** → posture is correct; document it and add a boot assertion / test so it can't regress.

Either way: add a startup check or test asserting the runtime DB role's `rolbypassrls = false` for the request path.

Related: #788, #789 (RLS source-of-truth drift surfaced during the same audit).

Rasbandit:
**Resolved by live prod audit (2026-06-29).** Connected read-only to prod RDS and checked `pg_roles`:

- `engram_admin` (the app's DATABASE_URL role): `rolbypassrls = false`, `rolsuper = false`
- `engram_app` (the with_tenant role): `rolbypassrls = false`
- Only AWS-managed `rdsadmin` bypasses — the app never uses it.

[...]

PR #1726, opened 2026-09-20
docs(rls): record the prod conflict and kill a dead hypothesis

Rasbandit:
Closes out the RLS audit. Doc-only.

## 1. The prod question, stated as the conflict it is

The doc previously said prod enforcement was "unresolved" and gestured at conflicting evidence. It now lays out what is actually on record, because the conflict is the useful part:

- **#1649 read prod directly**: tenant verifiably set, `notes` carrying both `relrowsecurity` and `relforcerowsecurity`, and `select count(*) from notes` returned **all 3,602 rows**. That says NOT enforced.
- **#1354 / #1357 record the opposite**, also against prod: a tenant-table read outside `with_tenant` returning **zero** rows. If the role simply ignored RLS, that would have returned everything.
- `OrphanSweep` running rather than refusing also points at not-enforced.

Both prod observations cannot hold unless something not yet identified conditions them. Settling it needs one read against prod, which I have not run.

## 2. A dead hypothesis, marked dead

#1649 proposed `pg_read_all_data` / `pg_write_all_data` membership as the mechanism, while flagging that the PostgreSQL docs say `pg_read_all_data` does not set `BYPASSRLS`.

Tested on staging against `notes` (FORCE RLS, 5,411 rows, no tenant set):

| role | `pg_read_all_data` | rows visible |
|---|---|---|
| `rls_probe_plain` | no | **0** |
| `rls_probe_readall` | **yes** | **0** |

It does not defeat FORCE RLS. Ruled out. The remaining candidate from `engram_admin`'s membership list is `rds_superuser`, which cannot be reproduced on FastRaid because FastRaid is not RDS. Probe roles were revoked and dropped; nothing left on staging.

## Context

This is the last item in the RLS code track. Bucket D is discharged 32/32, the ratchet is green, and the fix is validated on staging where the app genuinely connects as `engram_app` — a re-extraction bound all 12 genuinely-unambiguous dangling `note_links` edges, 12 → 0. What remains is infrastructure: the maintenance role, `MAINTENANCE_DATABASE_URL`, un-refusing `OrphanSweep`, and the prod credential flip.

Related: #1649, #790 (both commented with these findings), #1357 (closed as superseded).


Rasbandit:
## The behavioural probe ran on prod. It saw nothing.

`0.30.0` shipped `TenancyGuard`'s behavioural probe and it ran on every prod task at 2026-09-24 ~06:51 UTC (task defs `engram-saas-prod:169`, `engram-worker-prod:20`).

**Result: the probe set `app.current_tenant` to `00000000-0000-0000-0000-000000000000` and saw zero rows across all eleven tenant tables.**

### How that is derived, since the log does not say it directly

The boot line reports `:enforced`, and there is **no divergence warning** in Loki for that window. Working backwards through `enforcement/0`:

- `observed_enforcement/0` can only return `:bypassed` or `:unknown` — it deliberately cannot assert `:enforced` (see the moduledoc for why).
- `combine/2` returns `:enforced` only if one of the two signals is `:enforced`. It was not the probe, so it was `claimed_enforcement/0`.
- `report_divergence/2` stays silent when the two agree **or** when either abstains. `observed = :bypassed` against `claimed = :enforced` is a genuine disagreement and would have logged `DISAGREE` at warning severity. It did not.

The only combination that produces "reports `:enforced`, logs no divergence" is `observed = :unknown`, `claimed = :enforced`. And `:unknown` from the probe means no tenant table returned a row.

### What that does to this issue

This is the **opposite** of #1649, which watched the same role return all 3,602 `notes` rows with a tenant verifiably set. It agrees with #1354/#1357.

Two things are now settled that were not before:

1. **`engram_admin` has neither `rolsuper` nor `rolbypassrls`** — confirmed from inside the app's own pool, killing the "BYPASSRLS granted out of band" hypothesis (recorded when `0.29.0` shipped).
2. **Prod is not currently leaking rows across tenants.** Measured, not inferred from role attributes.

### What is still unexplained

#1649's observation. It was a direct read against prod with the tenant confirmed set, and it is not the kind of thing that is simply wrong. Either something changed between 2026-09-16 and now, or that read was conditioned on something nobody has isolated — a different connection path, a tenant that did not actually take, or a per-table `NO FORCE` window.

**That last one remains the most checkable candidate.** Ten migrations in this repo issue `ALTER TABLE <t> NO FORCE ROW LEVEL SECURITY` and re-`FORCE` at the end; an interrupted one leaves a single table exposed. The probe now covers every tenant table specifically so that state cannot hide, but it says nothing about what was true on 09-16.

### Caveat worth keeping

`:unknown` is formally three-way ambiguous: filtered, empty, or invisible to the transaction. Prod has neither empty tenant tables nor a sandbox, so filtering is the live explanation — strong evidence, not proof. The probe is deliberately built to only make the positive claim (`:bypassed`) with confidence.

### Consequence

None, by design. `enforced?/0` stayed true, so `OrphanSweep` continued to refuse — which #1739 exists to guarantee. If anything this result makes that refusal *correct* rather than over-cautious, and moves `MAINTENANCE_DATABASE_URL` from "nice to have" to the real unblock for Phase 2.

